PR #5, opened 2021-10-11
Search Toolkit component and unit tests.

mhdostal:
This is the PR for the Search Toolkit component and unit tests.  The Search component design is [here](https://devtopia.esri.com/runtime/common-toolkit/blob/master/designs/Search/Search.md).  Test design is [here](https://devtopia.esri.com/runtime/common-toolkit/blob/master/designs/Search/Search_Test_Design.md).

Some areas of interest:

- the SwiftUI List view does not have a built-in selection mechanism; instead I'm using a checkmark to denote the selected item:  
image
- I had to create a `SearchError` struct because it needs to implement `Equatable`; `Error` itself cannot be extended to do that.  I'm not sure I did that the best way, but it does compile and return errors.
- The component uses a MVVM approach; there's a `SearchViewModel` class which derives from `ObservableObject` and is marked as `@MainActor`, as it's `@Published` properties are used to update the UI, which is defined in `SearchView.swift`.
- The `acceptSuggestion`, `updateSuggestions`, `commitSearch` and `repeatSearch` model methods are defined as `async` in the design, but are NOT async in the model.  This is because the UI only triggers off changes to the `@Published` properties `results`, `suggestions`, and `selectedResult` and waiting on the methods was unnecessary.  This caused some challenges with the unit tests, having to wait on those properties, but everything now works great, both in the Examples app and unit tests.
- Internally, the methods mentioned above use a `Task`, as the underlying calls to the geocode/locator are still `async`.
- In the `Utility` folder are some modifiers for views.  I've prefaced them with `Esri` to indicate that they can be reusable "Esri" modifiers, but if there's a better name, or you don't like the prefix, let me know.
As an FYI, the Swift API worked flawlessly for this component.  Search a big, somewhat complicated bit of code, so if you find anything at all, let me know.  Thanks!

ps, the Toolkit repo doc for Search will be in a separate PR.  There was an effort underway to standardize each platform's Toolkit README and the Search doc will follow that.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Search/SearchView.swift
+++ b/Sources/ArcGISToolkit/Components/Search/SearchView.swift
@@ -297,7 +297,7 @@ extension ResultRow {
     }
 }
 
-/// A modifier which displays a background and shadow for a view.  Used to represent a selected view.
+/// A modifier which displays a background and shadow for a view. Used to represent a selected view.
 struct SelectedModifier: ViewModifier {
     /// `true` if the view should display as selected, `false` otherwise.
     var isSelected: Bool


Sources/ArcGISToolkit/Utility/SearchField.swift
```diff
@@ -0,0 +1,88 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import SwiftUI
+
+/// A custom view implementing a SearchField. It contains a search button, text field, delete text button,
+/// and a button to allow users to hide/show the search results list.
+public struct SearchField: View {
```
philium, line 18: Just out of curiosity, why was this made public? It seems more like an implementation detail and not something that should have been exposed outside the toolkit.

mhdostal: It was originally thought that the SearchField could be something that users could use separately from the rest of the Search component. I'm not sure it was ever tested at the time, though, nor if it's ever been used that way.

philium: Okay. Thanks!

PR #210, opened 2022-12-08
Add bottom content padding to Floating Panel

### LINKED ISSUE #209: UN Trace Tool: "Trace" button against border
In the Utility Network Trace Tool, the "Trace" button is pushed against the bottom edge of the floating panel hosting the tool. This doesn't look so great - there should be a little bit of spacing.

Screenshot 2022-12-08 at 13 05 12

dfeinzimer:
Closes #209

### DIFF
--- a/Sources/ArcGISToolkit/Components/FloatingPanel/FloatingPanel.swift
+++ b/Sources/ArcGISToolkit/Components/FloatingPanel/FloatingPanel.swift
@@ -84,7 +84,7 @@ struct FloatingPanel<Content>: View where Content: View {
                     }
                     content
                         .frame(height: height)
-                        .padding(.bottom, isCompact ? 25 : .zero)
+                        .padding(.bottom, isCompact ? 25 : 10)
                     if !isCompact {
                         Divider()
                         makeHandleView()


Sources/ArcGISToolkit/Components/FloatingPanel/FloatingPanel.swift
```diff
@@ -84,7 +84,7 @@ struct FloatingPanel<Content>: View where Content: View {
                     }
                     content
                         .frame(height: height)
-                        .padding(.bottom, isCompact ? 25 : .zero)
+                        .padding(.bottom, isCompact ? 25 : 10)
```
dfeinzimer, line 87: If `isCompact` is false, `10` provides an amount of content bottom padding that is similar to the amount of padding around the handle.

Screenshot 2022-12-08 at 14 21 49

If `isCompact` is true, we continue to use `25`. This values helps any content avoid layout conflicts with the iOS Home Indicator. On devices without the Home Indicator, `25` is more than needed but SwiftUI doesn't provide a method to detect the top of the Home Indicator so we need to use a value that works for both conditions.

Screenshot 2022-12-08 at 13 58 00
Screenshot 2022-12-08 at 13 58 53

mhdostal (approved):

PR #215, opened 2022-12-09
Resolve `@unknown default` warnings

### LINKED ISSUE #214: Resolve new compile warnings
image

Xcode 14.1
iPad Pro 10.5" iOS 16.1

dfeinzimer:
Closes #214 as reported by @zkline101 and @mhdostal

### DIFF
--- a/Sources/ArcGISToolkit/Components/Authentication/NetworkChallengeContinuation.swift
+++ b/Sources/ArcGISToolkit/Components/Authentication/NetworkChallengeContinuation.swift
@@ -71,6 +71,8 @@ extension NetworkChallengeContinuation.Kind {
             self = .login
         case .clientCertificate:
             self = .certificate
+        @unknown default:
+            fatalError("Unknown NetworkAuthenticationChallenge.Kind")
         }
     }
 }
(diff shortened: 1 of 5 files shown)


dfeinzimer (commented):
Screenshot 2022-12-09 at 11 36 18

There is still one warning remaining that I'm unsure of how to best handle. I've reached out to @njarecha for advice.

Sources/ArcGISToolkit/Components/Authentication/NetworkChallengeContinuation.swift
```diff
@@ -71,6 +71,8 @@ extension NetworkChallengeContinuation.Kind {
             self = .login
         case .clientCertificate:
             self = .certificate
+        @unknown default:
+            fatalError("Unknown NetworkAuthenticationChallenge.Kind")
```
dfeinzimer, line 75: I figure the message may be helpful if this is ever hit but I'd be fine with removing or changing it.

Sources/ArcGISToolkit/Components/Popups/Attachments/AttachmentsPopupElementModel.swift
```diff
@@ -58,6 +58,8 @@ import QuickLook
             defaultSystemName = "film"
         case .document, .other:
             defaultSystemName = "doc"
+        @unknown default:
+            defaultSystemName = nil
```
mhdostal, line 62: Have `defaultSystemName == nil` means that the user set their own thumbnail, so a non-nil value is required.  A "question mark" should denote that something is amiss...

```suggestion
            defaultSystemName = "questionmark"
```

mhdostal (changes requested):

mhdostal (approved):

PR #216, opened 2022-12-09
Fix jumpy Floor Filter behavior

### LINKED ISSUE #212: FloorFilter flashes when panning between sites.
Tapping on the Site button in the Floor Filter, zooming in, and then panning between sites causes the Sites and Facilities view to switch several times before settling down.  It only seems to happen when panning between sites, not between facilities in site.

Using the out-of-the-box FF example.

Xcode 14.1
iPad Pro 10.5" iOS 16.1

https://user-images.githubusercontent.com/3998072/206755808-a37e94e2-bfaf-4226-85cb-529089e9ff88.mov

dfeinzimer:
Closes #212

### DIFF
--- a/Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
+++ b/Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
@@ -43,9 +43,8 @@ struct SiteAndFacilitySelector: View {
         /// A site name filter phrase entered by the user.
         @State private var query: String = ""
         
-        /// Indicates that a user pressed the back button in the navigation view, indicating that
-        /// the site should technically be considered "de-selected" even though the viewpoint hasn't
-        /// changed.
+        /// Indicates that the user pressed the back button in the navigation view, indicating the
+        /// site should appear "de-selected" even though the viewpoint hasn't changed.
         @State private var userBackedOutOfSelectedSite = false
         
         /// Allows the user to toggle the visibility of the site and facility selector.
--- a/Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
+++ b/Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
@@ -43,8 +43,10 @@ struct SiteAndFacilitySelector: View {
         /// A site name filter phrase entered by the user.
         @State private var query: String = ""
         
-        /// Indicates whether a site entry should be considered "selected" in the list.
-        @State var shouldAutoSelect = false
+        /// Indicates that a user pressed the back button in the navigation view, indicating that
+        /// the site should technically be considered "de-selected" even though the viewpoint hasn't
+        /// changed.
+        @State private var userBackedOutOfSelectedSite = false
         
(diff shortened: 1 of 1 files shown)


Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
```diff
@@ -130,14 +131,19 @@ struct SiteAndFacilitySelector: View {
                         facilities: site.facilities,
                         isHidden: isHidden
                     )
-                    .onDisappear {
-                        shouldAutoSelect = false
-                    }
```
dfeinzimer, line 135: If a site was deselected as a result of a map pan, the `NavigationLink` would close causing the computed binding above to return `nil`, producing the jumpy behavior.

Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
```diff
@@ -117,7 +118,7 @@ struct SiteAndFacilitySelector: View {
                     tag: site,
                     selection: Binding(
                         get: {
-                            return shouldAutoSelect ? viewModel.selectedSite : nil
+                            userBackedOutOfSelectedSite ? nil : viewModel.selectedSite
```
dfeinzimer, line 121: Now, the binding will only return `nil` if the user tapped the back arrow, meaning we won't visually loose the selection as a result of a pan gesture.

Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
```diff
@@ -130,14 +131,19 @@ struct SiteAndFacilitySelector: View {
                         facilities: site.facilities,
                         isHidden: isHidden
                     )
-                    .onDisappear {
-                        shouldAutoSelect = false
-                    }
+                    .navigationBarBackButtonHidden(true)
+                    .navigationBarItems(
+                        leading: Button {
```
dfeinzimer, line 136: Because SwiftIUI doesn't provide a way to get a callback when the back arrow in a navigation view is pressed, we hide the default back button and provide our own which gives us this functionality.

mhdostal (commented):
If I keep the map panning within a site, the facility view will select the correct facility just fine.  If I pan between sites, the facility view will get dismissed, displaying the site view, and then jump back to the facility view after selecting the new site (and facility). So the multiple "jumps" have been resolved, but there's still the one.

Is that just part of the way the navigation works when programmatically selecting a new site?  Is there any way around that?

dfeinzimer:
> Is that just part of the way the navigation works when programmatically selecting a new site?

Yes, this is by design. The selection is bound to currently selected site & facility. In your recording, as you pan west across Alabama street you leave the bounds of the east site, causing the momentary "backing out" to the sites list until you once again enter the boundary of the west site.

> Is there any way around that?

Yup! Users can use `.alwaysNotClearing` to maintain the selection during panning until a new facility is found. E.g.:
Screenshot 2022-12-09 at 15 34 36
Simulator Screen Recording - iPad Pro (12 9-inch) (6th generation) - 2022-12-09 at 15 35 33

PR #217, opened 2022-12-12
Fix issue with thumbnails and ios 16.1: put selection and error repor…

### LINKED ISSUE #213: Incorrect spacing/alignment of BasemapGallery thumbnails
Using the sample list of basemaps, selecting the OSM (Blueprint) baseman causes the selection outline to be not clamped to the thumbnail and for the thumbnail to move down.  Also, selecting the World_Imagery basemap causes that thumbnail to move down a bit to accommodate the error badge.  Both of these are new-ish issues.

Using the out-of-the-box BasemapGallery example.

Xcode 14.1
iPad Pro 10.5" iOS 16.1

image

mhdostal:
Closes #213 - This was a regression between iOS v16.0 and v16.1.  The code behaves differently on iOS 16.1 (ugh).  The solution was to put the selection in an overlay, instead of as an element in the Stack. I used the same overlay as error reporting for convenience.

### DIFF
--- a/Sources/ArcGISToolkit/Components/BasemapGallery/BasemapGalleryCell.swift
+++ b/Sources/ArcGISToolkit/Components/BasemapGallery/BasemapGalleryCell.swift
@@ -75,7 +75,6 @@ struct BasemapGalleryCell: View {
                                 .frame(width: 16, height: 16)
                             Image(systemName: "exclamationmark.circle.fill")
                                 .foregroundColor(.red)
-                                .background(Color.clear)
                                 .frame(width: 32, height: 32)
                         }
                         Spacer()
--- a/Sources/ArcGISToolkit/Components/BasemapGallery/BasemapGalleryCell.swift
+++ b/Sources/ArcGISToolkit/Components/BasemapGallery/BasemapGalleryCell.swift
@@ -38,18 +38,11 @@ struct BasemapGalleryCell: View {
                                 .resizable()
                                 .aspectRatio(contentMode: .fit)
                                 .clipShape(roundedRect)
-                            if isSelected {
-                                roundedRect
-                                    .stroke(lineWidth: 2)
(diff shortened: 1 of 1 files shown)


Sources/ArcGISToolkit/Components/BasemapGallery/BasemapGalleryCell.swift
```diff
@@ -67,26 +60,34 @@ struct BasemapGalleryCell: View {
...
-                        .background(Color.clear)
-                        .frame(width: 32, height: 32)
+    /// Creates an overlay which is either a selection outline or an error icon.
+    /// - Returns: A thumbnail overlay view.
+    private func makeOverlay() -> some View {
+        Group {
+            if item.hasError {
+                HStack {
+                    Spacer()
+                    VStack{
```
dfeinzimer, line 70:
```suggestion
                    VStack {
```

Sources/ArcGISToolkit/Components/BasemapGallery/BasemapGalleryCell.swift
```diff
@@ -67,26 +60,34 @@ struct BasemapGalleryCell: View {
...
+                        ZStack {
+                            // For a white background behind the exclamation mark.
+                            Circle()
+                                .foregroundColor(.white)
+                                .frame(width: 16, height: 16)
+                            Image(systemName: "exclamationmark.circle.fill")
+                                .foregroundColor(.red)
+                                .background(Color.clear)
```
dfeinzimer, line 78: I diffed with/without this line, not sure that I see any difference. May I be missing something?

mhdostal: I removed it; the default is probably `.clear`.

dfeinzimer (approved):

PR #218, opened 2022-12-12
Final popup example map

mhdostal:
Final map is [here](https://arcgisruntime.maps.arcgis.com/home/item.html?id=9f3a674e998f461580006e626611f9ad).


dfeinzimer (approved):
Good to go!

PR #220, opened 2022-12-13
Update .gitignore

mhdostal:
Updates .gitignore so we don't get the `Package.resolved` file + other SPM files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,12 @@
 .DS_Store
 /.build
-/Packages
 xcuserdata/
 /*.xcodeproj
 DerivedData/
 .swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
+
+## Swift Package Manager
+# Add this line if you want to avoid checking in source code from Swift Package Manager dependencies.
+Packages/
+Package.pins
+Package.resolved


dfeinzimer (commented):
> If you use Swift package dependencies in your project, make sure to include the Package.resolved file in your Git repository and commit any changes to it. Don’t include the file in your .gitignore file. [via Apple Documentation](https://developer.apple.com/documentation/xcode/making-dependencies-available-to-xcode-cloud#:~:text=If%20you%20use%20Swift%20package%20dependencies%20in%20your%20project%2C%20make%20sure%20to%20include%20the%20Package.resolved%20file%20in%20your%20Git%20repository%20and%20commit%20any%20changes%20to%20it.%20Don%E2%80%99t%20include%20the%20file%20in%20your%20.gitignore%20file.)

This is probably in the context of app level projects but we technically depend on the Swift SDK, right? Thoughts?

mhdostal:
Yes, we depend on the SDK  Sounds like we should commit it, instead of changing the .gitignore file.  @yo1995 ?

yo1995:
Please refer to https://github.com/Esri/arcgis-runtime-samples-ios/pull/1062 for more details, where I first added these exclusions to gitignore for samples. The thinking was:

- when we merge samples `v.next` to `main`, no changes will happen to `main` until the next SDK release (except breaking hotfixes)
- that means if the SDK or toolkit bumps minor version, the samples on `main` still builds without any changes
- if we include the package lock files, then the version and SHA are frozen. User needs to explicitly run **Packages** > **Reser Package Caches** in order to build with the latest dependencies, which is not what we want
- besides, the doc shared by David looks like to be CI/CD specific, if not Xcode Cloud only. I don't think that applies to samples

Whether not including package resolved for toolkit is up to your decision. Since toolkit is a Swift package rather than an app, I can see the arguments from both sides. 👀

mhdostal:
> Whether not including package resolved for toolkit is up to your decision. Since toolkit is a Swift package rather than an app, I can see the arguments from both sides.

Thanks @yo1995.

The original Toolkit included `Package.resolved` in the `.gitignore`, so I'm fine putting it in the file here.

yo1995 (approved):

dfeinzimer (approved):

Issue #200, opened 2022-12-02
Remove GeoModel's conformance to ObservableObject in Examples app


mhdostal:
In the SDK, `ObservableObject` conformance was removed from `GeoModel`. 

Model objects need to be added to each of the examples that currently store the map/scene directly as a state object.

PR #223, opened 2022-12-15
200 1 0 build fixes

mhdostal:
Closes https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/200

This fixes the build, handling the removal of ObservableObject conformance by Map and Scene.  It adds a data model, one for Map-based objects and one for Scene.  Is the pattern used correct?  Is there a better way?


mhdostal:
@philium - I'm interested in your opinion on the pattern here in regards to the data models and map/scene support.  This is related to https://devtopia.esri.com/runtime/swift/issues/3174.  Thanks!

philium:
A few thoughts:

- The geo model property of each should be immutable
- The utility of such a class will be fairly limited since it only stores a map. If anything else is needed (graphics overlay, portal, etc.), then a custom class will be needed
- Because the implementation is so simple and straightforward, I would argue that it doesn't pull its own weight and therefore shouldn't be included as part of the toolkit. Instead, it should be in the Examples app itself.
- Due to the standalone nature of samples, even if the classes were part of the toolkit's public API, I don't think that the samples app could use them. Does the same principle (keeping them as self-contained as possible) apply to toolkit examples?

mhdostal:
@philium Thanks for the feedback.  A similar change will have to happen in the Swift Examples app.  Did you have a thought on how that would work?

> * The utility of such a class will be fairly limited since it only stores a map. If anything else is needed (graphics overlay, portal, etc.), then a custom class will be needed

I could create a model inside each example, but that would just duplicate a bunch of code.  I'm open to suggestions.

Sources/ArcGISToolkit/Utility/MapDataModel.swift
```diff
@@ -0,0 +1,27 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import SwiftUI
+import ArcGIS
+
+/// A data model class containing a Map.
+public class MapDataModel: ObservableObject {
+    /// The `Map` used for display in a `MapView`.
+    public var map: Map
```
dfeinzimer, line 20: > By default an [ObservableObject](https://developer.apple.com/documentation/combine/observableobject) synthesizes an [objectWillChange](https://developer.apple.com/documentation/combine/observableobject/objectwillchange-2oa5v) publisher that emits the changed value before any of its @Published properties changes. - [Apple](https://developer.apple.com/documentation/combine/observableobject#:~:text=By%20default%20an%20ObservableObject%20synthesizes%20an%20objectWillChange%20publisher%20that%20emits%20the%20changed%20value%20before%20any%20of%20its%20%40Published%20properties%20changes.)

Should this be marked `@Published` (along with `scene` in `SceneDataModel`)?

philium:
> I could create a model inside each example, but that would just duplicate a bunch of code. I'm open to suggestions.

It would need to be in many places, but without doc it is what, seven lines of code? That's what I mean by it doesn't pull its own weight. It is so trivial to implement in the samples app (or in each example for that matter) that it isn't worth adding the public API of the toolkit. Anything new public API should be tested and needs to be maintained. I just don't think this meets the bar for inclusion.

dfeinzimer (dismissed):

mhdostal:
@dfeinzimer I moved the data models to the Examples app, per @philium's recommendation; I did have to add a data model to the OverviewMap component which needed it.

Sources/ArcGISToolkit/Components/OverviewMap.swift
```diff
@@ -185,3 +185,15 @@ private extension Symbol {
...
+public class MapDataModel: ObservableObject {
+    /// The `Map` used for display in a `MapView`.
+    @Published public var map: Map
+    
+    /// Creates a `MapDataModel`.
+    /// - Parameter map: The `Map` used for display.
+    public init(map: Map) {
+        self.map = map
+    }
+}
```
dfeinzimer, line 199: I think we'll need the equivalent of this for `Scene`.
Screenshot 2022-12-16 at 12 10 16

Also, should this be marked `public`? Otherwise I think the ones added in Examples/Utility are redundant. Not sure why a duplicate symbol error isn't produced.

Examples/Utility/MapDataModel.swift
```diff
@@ -0,0 +1,33 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import SwiftUI
+import ArcGIS
+
+/// A very basic data model class containing a Map. Since a `Map` is not an observable object,
+/// clients can use `MapDataModel` as an example of how you would store a map in a data model
+/// class. The class inherits from `ObservableObject` and the `Map` is defined as an @Published
```
dfeinzimer, line 19:
```suggestion
/// class. The class inherits from `ObservableObject` and the `Map` is defined as a @Published
```

Examples/Utility/SceneDataModel.swift
```diff
@@ -0,0 +1,33 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import SwiftUI
+import ArcGIS
+
+/// A very basic data model class containing a Scene. Since a `Scene` is not an observable object,
+/// clients can use `SceneDataModel` as an example of how you would store a scene in a data model
+/// class. The class inherits from `ObservableObject` and the `Scene` is defined as an @Published
```
dfeinzimer, line 19:
```suggestion
/// class. The class inherits from `ObservableObject` and the `Scene` is defined as a @Published
```

Sources/ArcGISToolkit/Components/OverviewMap.swift
mhdostal: I made the Example models internal and the OverviewMap model private. The `Cannot find 'SceneDataModel' in scope` error shouldn't happen, as the `SceneDataModel` is now in the Examples app.

dfeinzimer: Hmm, still having issues here.
Screenshot 2022-12-16 at 13 17 05

I think the app level `MapDataModel` and `SceneDataModel` aren't being compiled, because they're not in Compile Sources and now that `MapDataModel` in `OverviewMap` is private, non of the example structs can resolve it.

Screenshot 2022-12-16 at 13 21 06

dfeinzimer: Ref #224

dfeinzimer (approved):
Working good!

rolson:
@mhdostal - instead of the 2 classes, can't we just have one generic:

```swift
class GeoViewModel<T: GeoModel>: ObservableObject {
    let geoModel: T
    
    init(geoModel: T) {
        self.geoModel = geoModel
    }
}
```

philium:
And then have type aliases for the two geo model subclasses:

```swift
typealias MapViewModel = GeoViewModel<Map>
typealias SceneViewModel = GeoViewModel<Scene>
```

PR #226, opened 2022-12-19
Remove uses of `CustomStringConvertible.description` from `UtilityNetworkTrace.swift`

philium:
From the doc for `CustomStringConvertible.description`:

> image

Values displayed to the user should always be formatted. This PR replaces uses of `CustomStringConvertible.description` with a formatted value, which includes the correct grouping separator (when appropriate).

### DIFF
--- a/Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTrace.swift
+++ b/Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTrace.swift
@@ -138,10 +138,13 @@ public struct UtilityNetworkTrace: View {
             }
             makeDetailSectionHeader(title: assetGroupName)
             List {
-                ForEach(assetTypeGroups.keys.compactMap({$0}).sorted(), id: \.self) { assetTypeGroupName in
-                    Section(assetTypeGroupName) {
+                ForEach(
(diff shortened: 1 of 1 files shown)


Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTrace.swift
```diff
@@ -635,9 +642,9 @@ public struct UtilityNetworkTrace: View {
     // MARK: Computed Properties
     
     /// Indicates the number of the trace currently being viewed out the total number of traces.
-    private var currentTraceLabel: String {
+    private var currentTraceLabel: LocalizedStringKey {
         guard let index = viewModel.selectedTraceIndex else { return "Error" }
-        return "Trace \(index+1) of \(viewModel.completedTraces.count.description)"
+        return "Trace \(index+1) of \(viewModel.completedTraces.count)"
```
philium, line 647: It may not look like it, but these values are now formatted (rather than the unformatted `description` which string interpolation would normally produce). Because the return type is `LocalizedStringKey`, [this](https://developer.apple.com/documentation/swiftui/localizedstringkey/stringinterpolation/appendinterpolation(_:)-378v9) interpolation method is now being used.

dfeinzimer (approved):
Changes look great, thank you Phil!

mhdostal (approved):

Issue #225, opened 2022-12-19
Runtime warning using `OverviewMap.map(_:)`


philium:
The following runtime warning is generated when using `OverviewMap.map(_:)`:

![image](https://user-images.githubusercontent.com/2257493/208494105-e7346c96-d6c6-452a-95c8-4b3fd38fca02.png)

In addition to the warning, the modifier approach isn't great for a couple reasons:

- SwiftUI view content is usually taken via an initializer rather than a modifier. In general, modifiers are used to configure views, not provide data for them.
- It encourages the map to be created where the modifier is used, in the body of a view, which is an anti-pattern. With that naive approach, the map would get created every time the view's body is re-computed.

A better approach might be to have a separate "initializer" (static method it looks like) that takes a map.

mhdostal:
Thanks @philium !

PR #232, opened 2022-12-22
Fix OverviewMap warning

mhdostal:
Closes https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/225

This removes the `.map` modifier on `OverviewMap` and replaces it with a `map` argument on both `OverviewMap.forSceneView` and `OverviewMap.forMapView`. The argument defaults to `nil`, in which case the existing default map is used.

It also removes the unnecessary distinction between `customOverviewMapForMapView` and `customOverviewMapForSceneView` and replaces them with just one custom map: `customOverviewMap`.

### DIFF
--- a/Examples/Examples/OverviewMapExampleView.swift
+++ b/Examples/Examples/OverviewMapExampleView.swift
@@ -63,12 +63,12 @@ struct OverviewMapForMapView: View {
             .overlay(
                 OverviewMap.forMapView(
                     with: viewpoint,
-                    visibleArea: visibleArea
+                    visibleArea: visibleArea// ,
+                    // map: .customOverviewMap // Uncomment to use a custom map.
                 )
                 // These modifiers show how you can modify the default
                 // values used for the symbol, map, and scaleFactor.
 //                    .symbol(.customFillSymbol)
-//                    .map(.customOverviewMapForMapView)
 //                    .scaleFactor(15.0)
                     .frame(width: 200, height: 132)
                     .padding(),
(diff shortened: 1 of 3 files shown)


Examples/Examples/OverviewMapExampleView.swift
```diff
@@ -131,9 +133,6 @@ private extension Symbol {
 }
 
 private extension Map {
-    /// A custom map for the `OverviewMap` used in a MapView.
-    static let customOverviewMapForMapView = Map(basemapStyle: .arcGISDarkGray)
-
-    /// A custom map for the `OverviewMap` used in a SceneView.
-    static let customOverviewMapForSceneView = Map(basemapStyle: .arcGISDarkGray)
+    /// A custom map for the `OverviewMap`.
+    static let customOverviewMap = Map(basemapStyle: .arcGISDarkGray)
```
philium, line 137: This really isn't great because if it is ever used, it will live on in perpetuity. That may be fine for an examples app, but probably worth adding a comment that in a production app, the map should be stored on a state object and not in a static property.

Sources/ArcGISToolkit/Components/OverviewMap.swift
```diff
@@ -43,37 +41,57 @@ public struct OverviewMap: View {
...
     ) {
         self.visibleArea = visibleArea
         self.viewpoint = viewpoint
         self.symbol = symbol
         
+        _dataModel = StateObject(
+            wrappedValue: MapDataModel(
+                map: map ?? Map(basemapStyle: .arcGISTopographic)
+            )
+        )
```
philium, line 93: I don't know that this will do what you'd expect. Say the user has two maps, one light and one dark, and they toggle between the two. A new state object is created each time, which isn't how state objects are intended to be used. Instead, I would suggest storing the user provided map on the view and creating the default map in the view model. That way the view model itself, and therefore the default map, will only be created if necessary (remember that `StateObject.init(wrappedValue:)` takes an autoclosure, so the wrapped value is lazily evaluated the first time the state object is accessed). It would look something like this:

```swift
// In 'OverviewMap'
private class Model: ObservableObject {
    let defaultMap = Map(basemapStyle: .arcGISTopographic)
}

@StateObject private var model = Model()

private let userProvidedMap: Map?

private var effectiveMap: Map {
    userProvidedMap ?? model.defaultMap
}

// In the 'body' of 'OverviewMap'

MapView(
    map: effectiveMap,
    viewpoint: makeOverviewViewpoint(),
    graphicsOverlays: [graphicsOverlay]
)
```

Examples/Examples/OverviewMapExampleView.swift
mhdostal: I moved this to a property in the two places that need it,

philium (approved):

PR #234, opened 2022-12-28
Resolve Utility Network Trace log

### LINKED ISSUE #233: Utility Network Trace - "geometries must not be empty"
See #227 for original report

- "geometries must not be empty" is logged repeatedly to the console during normal utility network trace use.

dfeinzimer:
Closes #233 

Resolves the log message by avoiding calling `GeometryEngine.combineExtents(of:)` with an empty sequence of geometries which was occurring when running a "Connected" trace.


Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTraceViewModelTrace.swift
```diff
@@ -82,19 +82,21 @@ extension UtilityNetworkTraceViewModel {
         
         /// The extent of the trace's geometry result with a small added buffer.
         var resultExtent: Envelope? {
-            if let resultEnvelope = GeometryEngine.combineExtents(of: [
+            let geometries = [
                 utilityGeometryTraceResult?.multipoint,
                 utilityGeometryTraceResult?.polygon,
                 utilityGeometryTraceResult?.polyline
```
philium, line 88: Instead of unwrapping `utilityGeometryTraceResult` three different times, it would be good to guard that it is not nil.

```suggestion
            guard let utilityGeometryTraceResult else { return nil }
            
            let geometries = [
                utilityGeometryTraceResult.multipoint,
                utilityGeometryTraceResult.polygon,
                utilityGeometryTraceResult.polyline
```

Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTraceViewModelTrace.swift
```diff
@@ -82,19 +82,21 @@ extension UtilityNetworkTraceViewModel {
...
-            ].compactMap { $0 }),
-               let expandedEnvelope = GeometryEngine.buffer(
-                around: resultEnvelope,
-                distance: 200
-               ) {
-                return expandedEnvelope.extent
-            } else {
+            ]
+                .compactMap { $0 }
+                .filter { !$0.isEmpty }
```
philium, line 91: Not a big deal in this case because the collection is small, but in general, it is best to iterate the collection as few times as possible. In this case, the filtering could be done in the transform closure when compact mapping.

```suggestion
                .compactMap { geometry in
                    if let geometry, !geometry.isEmpty {
                        return geometry
                    } else {
                        return nil
                    }
                }
```

Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTraceViewModelTrace.swift
```diff
@@ -82,19 +82,21 @@ extension UtilityNetworkTraceViewModel {
...
+            
+            guard !geometries.isEmpty,
```
philium, line 93: Ideally this check shouldn't be necessary; the geometry engine method should just return `nil` (which it does) and not log anything (which it doesn't).

philium (commented):
FYI: Due to the nature of structures as a value type, it is very helpful to keep the declaration as succinct as possible. Consider moving all computed properties and methods into an extension, keeping on what is needed (stored properties) in the declaration itself.

Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTraceViewModelTrace.swift
dfeinzimer: Agreed. I think this would be covered by Swift 3246, no? Any changes needed in this PR here?

dfeinzimer:
> FYI: Due to the nature of structures as a value type, it is very helpful to keep the declaration as succinct as possible. Consider (in a separate PR) moving all computed properties and methods into an extension

👍 #237

Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTraceViewModelTrace.swift
philium: It is likely that this will be addressed in the near future. Might be best to wait and see.

dfeinzimer: @philium With 3264 merged, I no longer seem to be able to trigger the empty geometries error. I can, however, easily trigger the "The input feature does not belong to this utility network." log. Is that expected after 3264 still?

philium: Yes, it is expected, at least for now.

philium:
I'm confused. Now the changes in this PR look like the changes in #238. Am I missing something?

dfeinzimer:
> I'm confused. Now the changes in this PR look like the changes in #238. Am I missing something?

Good question. That had me going for a minute there too because I'd forgotten what happened. #238 merged into this branch and this is merging into v.next.

`guard !geometries.isEmpty` is gone now:

Screenshot 2023-01-09 at 15 19 54



Also note: different change counts:

Screenshot 2023-01-09 at 15 23 41

Screenshot 2023-01-09 at 15 23 55

philium:
I'm sorry, @dfeinzimer. I forgot about this PR.

> #238 merged into this branch and this is merging into v.next.

Oh! That clears things up. Thanks!

Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTraceViewModelTrace.swift
```diff
@@ -114,6 +67,63 @@ extension UtilityNetworkTraceViewModel {
     }
 }
 
+extension UtilityNetworkTraceViewModel.Trace {
+    /// - Parameter name: A name of a utility asset group.
+    /// - Returns: The set of utility elements returned by the trace that belong to the provided
+    /// asset group, grouped by type.
+    func elementsByTypeInGroup(named name: String) -> [String: [UtilityElement]] {
```
philium, line 74: This method should have a description. Also, the name of it should be `elementsByType(inGroupNamed:)`.

Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTraceViewModelTrace.swift
```diff
@@ -114,6 +67,63 @@ extension UtilityNetworkTraceViewModel {
...
+        let assetsInGroup = elementsInAssetGroup(named: name)
+        var result = [String : [UtilityElement]]()
+        assetsInGroup.forEach { e in
+            var assetTypeGroup = result[e.assetType.name, default: []]
+            assetTypeGroup.append(e)
+            result.updateValue(assetTypeGroup, forKey: e.assetType.name)
+        }
+        return result
```
philium, line 82: Consider using `reduce(into:_:)`.

```suggestion
        return elementsInAssetGroup(named: name)
            .reduce(into: [:]) { result, element in
                let key = element.assetType.name
                var assetTypeGroup = result[key, default: []]
                assetTypeGroup.append(element)
                result[key] = assetTypeGroup
            }
```

Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTraceViewModelTrace.swift
```diff
@@ -114,6 +67,63 @@ extension UtilityNetworkTraceViewModel {
...
+    }
+    
+    /// - Parameter name: A name of a utility asset group.
+    /// - Returns: The set of utility elements returned by the trace that belong to the provided
+    /// asset group.
+    func elementsInAssetGroup(named name: String) -> [UtilityElement] {
```
philium, line 88: FYI: This method should have a description. Also, the name should be `elements(inAssetGroupNamed:)`.

Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTraceViewModelTrace.swift
```diff
@@ -114,6 +67,63 @@ extension UtilityNetworkTraceViewModel {
...
+        return elementResults.filter({ $0.assetGroup.name == name })
```
philium, line 89: Consider using trailing closure.

```suggestion
        return elementResults.filter { $0.assetGroup.name == name }
```

Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTraceViewModelTrace.swift
```diff
@@ -114,6 +67,63 @@ extension UtilityNetworkTraceViewModel {
...
+    }
+    
+    /// A set of the asset group names returned by the trace.
+    var assetGroupNames: Set<String> {
+        var assetGroupNames = Set<String>()
+        elementResults.forEach {
+            assetGroupNames.insert($0.assetGroup.name)
+        }
+        return assetGroupNames
```
philium, line 98:
```suggestion
        Set(elementResults.map(\.assetGroup.name))
```

Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTraceViewModelTrace.swift
```diff
@@ -114,6 +67,63 @@ extension UtilityNetworkTraceViewModel {
...
+            return nil
+        }
+        
+        return expandedEnvelope.extent
+    }
+}
+
 extension UtilityNetworkTraceViewModel.Trace: Equatable {
     static func == (
         lhs: UtilityNetworkTraceViewModel.Trace,
```
philium, line 129: FYI: Since `Trace` is a struct, `Self` can be used here, which makes the declaration fit on one line nicely:

```swift
static func == (lhs: Self, rhs: Self) -> Bool {
    return lhs.id == rhs.id
}
```

philium (approved):

dfeinzimer:
Thank you Phil!

PR #236, opened 2022-12-29
Floor Filter - Fix ability to reselect same site

### LINKED ISSUE #229: Cannot reselect site using floor filter
Steps to reproduce:

1. Open the Floor Filter example
2. Tap/click the button to open the site selector
3. Tap/click on a site
4. Tap/click the back button
5. Tap/click the same site

Expected: The site will be selected and the facilities for that site will be presented.
Actual: Nothing happens.

dfeinzimer:
Closes #229 

Leaving `userBackedOutOfSelectedSite` set `true` caused the computed binding above on line 121 to return `nil` when it shouldn't have. It was being left `true` because the `.onChange` modifier below at line 148 wasn't firing if the same site was selected as before.


philium (approved):
I verified that the change fixes the bug.

PR #240, opened 2023-01-06
Adjust how certificates are imported

carl8382:
- Add security scoping to the url which seems to fix issues accessing iCloud Document url's 
- Switch to using a detached task to avoid deserializing the certificate on the main thread.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
+++ b/Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
@@ -71,14 +71,18 @@ import ArcGIS
             preconditionFailure()
         }
         
-        Task {
+        Task.detached {
             do {
-                challenge.resume(with: .continueWithCredential(try .certificate(at: certificateURL, password: password)))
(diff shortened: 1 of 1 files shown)


Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
```diff
@@ -71,14 +71,18 @@ import ArcGIS
             preconditionFailure()
         }
         
-        Task {
+        Task.detached {
             do {
-                challenge.resume(with: .continueWithCredential(try .certificate(at: certificateURL, password: password)))
+                guard certificateURL.startAccessingSecurityScopedResource() else {
+                    self.showCertificateImportError(nil)
```
rolson, line 77: This will end up showing "The certificate file or password was invalid." Is that ok or should it show some error about permission to access the file if this fails?

rolson: ok, we can get this in and then follow up with a PR to make better error messages.

rolson (approved):

Issue #241, opened 2023-01-06
Add specific error message for when accessing a certificate file fails


rolson:
A [PR was merged](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/240) that fixed an issue with importing a certificate. However we need to follow up on that and make sure to [have specific error message](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/240#discussion_r1063583931) for the failure to access a certificate file.

rolson:
cc @carl8382

rolson:
It might be useful to have a toolkit level specific error:

```swift
enum CertificateError: Error {
    case couldNotAccessSecureScopedResource
    case importError(CertificateImportError)
    case unknown
}

extension CertificateError {
    var message: String {
        switch self {
        case .couldNotAccessSecureScopedResource:
            return "Could not access certificate file."
        case .unknown:
            return "Unknown error accessing and importing the certificate."
        case .importError(let certificateImportError):
            switch certificateImportError {
            case .invalidData:
                return "The certificate file was invalid."
            case .invalidPassword:
                return "The password was invalid."
            default:
                return defaultMessage
            }
        }
    }
}
```

and use that instead of `CertificateImportError`

dfeinzimer:
[Completed](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/249#issuecomment-2013081235)

PR #243, opened 2023-01-06
added certificate errors

njarecha:
Closes https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/241

### DIFF
--- a/Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
+++ b/Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
@@ -83,17 +83,17 @@ import ArcGIS
         
         Task.detached {
             do {
-                guard certificateURL.startAccessingSecurityScopedResource() else {
-                    await self.showCertificateError(CertificateError.couldNotAccessCertificateFile)
-                    return
+                if certificateURL.startAccessingSecurityScopedResource() {
+                    defer { certificateURL.stopAccessingSecurityScopedResource() }
+                    let credential = try NetworkCredential.certificate(at: certificateURL, password: password)
+                    await self.challenge.resume(with: .continueWithCredential(credential))
+                } else {
+                    await self.showCertificateError(.couldNotAccessCertificateFile)
                 }
-                defer { certificateURL.stopAccessingSecurityScopedResource() }
-
-                await self.challenge.resume(with: .continueWithCredential(try .certificate(at: certificateURL, password: password)))
             } catch(let certificateImportError as CertificateImportError) {
-                await self.showCertificateError(CertificateError.importError(certificateImportError))
+                await self.showCertificateError(.importError(certificateImportError))
             } catch {
-                await self.showCertificateError(CertificateError.other(error))
+                await self.showCertificateError(.other(error))
             }
         }
     }
@@ -119,11 +119,10 @@ extension CertificateImportError: LocalizedError {
         case .invalidPassword:
             return String(localized: "The password was invalid.", bundle: .module)
         default:
-            let errorString = SecCopyErrorMessageString(rawValue, nil) as? String ?? String(
+            return SecCopyErrorMessageString(rawValue, nil) as String? ?? String(
                 localized: "The certificate file or password was invalid.",
                 bundle: .module
             )
-            return errorString
         }
     }
 }
@@ -253,7 +252,8 @@ private extension View {
                 viewModel.cancel()
             }
         } message: {
-            Text(viewModel.certificateError?.localizedDescription ?? String(
+            Text(
+                viewModel.certificateError?.localizedDescription ?? String(
                     localized: "The certificate file or password was invalid.",
                     bundle: .module
                  )
--- a/Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
+++ b/Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
@@ -24,8 +24,8 @@ import ArcGIS
         case couldNotAccessCertificateFile
         /// The certificate import error.
         case importError(CertificateImportError)
-        // The failure error.
-        case failure(Error)
+        // The other error.
+        case other(Error)
     }
     
     /// The challenge that requires a certificate to proceed.
@@ -93,7 +93,7 @@ import ArcGIS
             } catch(let certificateImportError as CertificateImportError) {
                 await self.showCertificateError(CertificateError.importError(certificateImportError))
             } catch {
-                await self.showCertificateError(CertificateError.failure(error))
+                await self.showCertificateError(CertificateError.other(error))
             }
         }
     }
@@ -111,21 +111,28 @@ import ArcGIS
     }
 }
 
+extension CertificateImportError: LocalizedError {
+    public var errorDescription: String? {
+        switch self {
+        case .invalidData:
+            return NSLocalizedString("The certificate file was invalid.", comment: "Invalid certificate")
+        case .invalidPassword:
+            return NSLocalizedString("The password was invalid.", comment: "Invalid password")
+        default:
+            let errorString = SecCopyErrorMessageString(rawValue, nil) as? String ?? "The certificate file or password was invalid."
+            return NSLocalizedString(errorString, comment: "Other certificate error")
+        }
+    }
+}
+
 extension CertificatePickerViewModel.CertificateError {
     var message: String {
         switch self {
         case .couldNotAccessCertificateFile:
             return "Could not access the certificate file."
         case .importError(let certificateImportError):
-            switch certificateImportError {
-            case .invalidData:
-                return "The certificate file was invalid."
-            case .invalidPassword:
-                return "The password was invalid."
-            default:
-                return "The certificate file or password was invalid."
-            }
-        case .failure(let error):
+            return certificateImportError.localizedDescription
+        case .other(let error):
             return error.localizedDescription
         }
     }
--- a/Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
+++ b/Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
@@ -128,7 +128,7 @@ extension CertificateImportError: LocalizedError {
 }
 
 extension CertificatePickerViewModel.CertificateError: LocalizedError {
-    public var errorDescription: String? {
+    var errorDescription: String? {
         switch self {
         case .couldNotAccessCertificateFile:
             return String(localized: "Could not access the certificate file.", bundle: .module)
--- a/Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
+++ b/Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
@@ -18,6 +18,16 @@ import ArcGIS
 /// An object that provides the business logic for the workflow of prompting the user for a
 /// certificate and a password.
 @MainActor final class CertificatePickerViewModel: ObservableObject {
+    /// The types of certificate error.
+    enum CertificateError: Error {
+        /// Could not access the certificate file.
+        case couldNotAccessCertificateFile
+        /// The certificate import error.
+        case importError(CertificateImportError)
+        // The failure error.
+        case failure(Error)
+    }
+    
     /// The challenge that requires a certificate to proceed.
     let challenge: NetworkChallengeContinuation
     
@@ -34,10 +44,10 @@ import ArcGIS
     @Published var showPassword = false
     
     /// A Boolean value indicating whether to display the error.
-    @Published var showCertificateImportError = false
+    @Published var showCertificateError = false
     
-    /// The certificate import error that occurred.
-    var certificateImportError: CertificateImportError?
+    /// The certificate error that occurred.
+    var certificateError: CertificateError?
     
     /// The host that prompted the challenge.
     var challengingHost: String {
@@ -74,15 +84,16 @@ import ArcGIS
         Task.detached {
             do {
                 guard certificateURL.startAccessingSecurityScopedResource() else {
-                    await self.showCertificateImportError(nil)
+                    await self.showCertificateError(CertificateError.couldNotAccessCertificateFile)
                     return
                 }
-
                 defer { certificateURL.stopAccessingSecurityScopedResource() }
 
                 await self.challenge.resume(with: .continueWithCredential(try .certificate(at: certificateURL, password: password)))
+            } catch(let certificateImportError as CertificateImportError) {
+                await self.showCertificateError(CertificateError.importError(certificateImportError))
             } catch {
-                await self.showCertificateImportError(error)
+                await self.showCertificateError(CertificateError.failure(error))
             }
         }
     }
@@ -92,11 +103,31 @@ import ArcGIS
         challenge.resume(with: .cancel)
     }
 
-    private func showCertificateImportError(_ error: Error?) async {
+    private func showCertificateError(_ error: CertificateError) async {
         // This is required to prevent an "already presenting" error.
         try? await Task.sleep(nanoseconds: 100_000)
-        certificateImportError = error as? CertificateImportError
-        showCertificateImportError = true
+        certificateError = error
+        showCertificateError = true
+    }
+}
+
+extension CertificatePickerViewModel.CertificateError {
+    var message: String {
+        switch self {
+        case .couldNotAccessCertificateFile:
+            return "Could not access the certificate file."
+        case .importError(let certificateImportError):
+            switch certificateImportError {
+            case .invalidData:
+                return "The certificate file was invalid."
+            case .invalidPassword:
+                return "The password was invalid."
+            default:
+                return "The certificate file or password was invalid."
+            }
+        case .failure(let error):
+            return error.localizedDescription
+        }
     }
 }
 
@@ -139,8 +170,8 @@ struct CertificatePickerViewModifier: ViewModifier {
                     }
                 )
             )
-            .alertCertificateImportError(
-                isPresented: $viewModel.showCertificateImportError,
+            .alertCertificateError(
+                isPresented: $viewModel.showCertificateError,
                 viewModel: viewModel
             )
     }
@@ -199,7 +230,7 @@ private extension View {
     /// - Parameters:
     ///   - isPresented: A Boolean value indicating if the view is presented.
     ///   - viewModel: The view model associated with the view.
-    @MainActor @ViewBuilder func alertCertificateImportError(
+    @MainActor @ViewBuilder func alertCertificateError(
         isPresented: Binding<Bool>,
         viewModel: CertificatePickerViewModel
     ) -> some View {
@@ -212,23 +243,7 @@ private extension View {
                 viewModel.cancel()
             }
         } message: {
-            Text(message(for: viewModel.certificateImportError))
-        }
-    }
-    
-    func message(for error: CertificateImportError?) -> String {
-        let defaultMessage = "The certificate file or password was invalid."
-        guard let error = error else {
-            return defaultMessage
-        }
-        
-        switch error {
-        case .invalidData:
-            return "The certificate file was invalid."
-        case .invalidPassword:
-            return "The password was invalid."
-        default:
-            return defaultMessage
+            Text(viewModel.certificateError?.message ?? "The certificate file or password was invalid.")
         }
     }
 }
--- a/Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
+++ b/Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
@@ -115,12 +115,15 @@ extension CertificateImportError: LocalizedError {
     public var errorDescription: String? {
         switch self {
         case .invalidData:
-            return NSLocalizedString("The certificate file was invalid.", comment: "Invalid certificate")
+            return String(localized: "The certificate file was invalid.", bundle: .module)
         case .invalidPassword:
-            return NSLocalizedString("The password was invalid.", comment: "Invalid password")
+            return String(localized: "The password was invalid.", bundle: .module)
         default:
-            let errorString = SecCopyErrorMessageString(rawValue, nil) as? String ?? "The certificate file or password was invalid."
-            return NSLocalizedString(errorString, comment: "Other certificate error")
+            let errorString = SecCopyErrorMessageString(rawValue, nil) as? String ?? String(
+                localized: "The certificate file or password was invalid.",
+                bundle: .module
+            )
+            return errorString
         }
     }
 }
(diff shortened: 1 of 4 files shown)


dfeinzimer (approved):

Tests/ArcGISToolkitTests/CertificatePickerViewModelTests.swift
```diff
@@ -42,7 +42,7 @@ import XCTest
         // Another yield seems to be required to deal with timing when running the test
         // repeatedly.
         await Task.yield()
-        XCTAssertTrue(model.showCertificateImportError)
+        XCTAssertTrue(model.showCertificateError)
```
dfeinzimer, line 45: This assertion seems to be failing.

dfeinzimer (changes requested):

Tests/ArcGISToolkitTests/CertificatePickerViewModelTests.swift
njarecha: This passes for me. Which macOS, Xcode, and iOS versions are you testing?

dfeinzimer: macOS Ventura 13.0.1
Xcode 14.1
iOS 15.5 & 16.1

njarecha: Test passes for me with this config. Does test fail for you with on `v.next` branch as well or just with branch of this PR? 

M1 Max
macOS Monterey 12.6.2
Xcode 14.1
iOS 16.1

dfeinzimer: It passes okay for me on v.next.

rolson: I ran it 1000x times and it failed twice.  So some timing involved.

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
```diff
@@ -18,6 +18,16 @@ import ArcGIS
...
 /// certificate and a password.
 @MainActor final class CertificatePickerViewModel: ObservableObject {
+    /// The types of certificate error.
+    enum CertificateError: Error {
+        /// Could not access the certificate file.
+        case couldNotAccessCertificateFile
+        /// The certificate import error.
+        case importError(CertificateImportError)
+        // The failure error.
+        case failure(Error)
```
rolson, line 28: this might be better called `other`?

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
```diff
@@ -92,11 +103,31 @@ import ArcGIS
...
+            return "Could not access the certificate file."
+        case .importError(let certificateImportError):
+            switch certificateImportError {
+            case .invalidData:
+                return "The certificate file was invalid."
+            case .invalidPassword:
+                return "The password was invalid."
+            default:
+                return "The certificate file or password was invalid."
+            }
```
rolson, line 127: I think actually I would delegate this to the localized description or message of the `CertificateImportError`.

Tests/ArcGISToolkitTests/CertificatePickerViewModelTests.swift
njarecha: I see failure in run in repeat mode. Increasing sleep time on line 41 does not help. How to fix this?

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
njarecha: Firth I thought about `other` but `failure` is being used at other places so thought of using it. If you think `other` is better than I can change it.

rolson: `other` seems to make sense here.

Tests/ArcGISToolkitTests/CertificatePickerViewModelTests.swift
rolson: I'm not sure.  It seems rare and was probably already there, so you can put a bug in and we can try to address it later.

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
njarecha: You mean just use `localizedDescription` property of the error or adopt, `LocalizedError` on `CertificateImportError` and move these strings in it? If we just use the `localizedDescription` then the value will be something like this, `"The operation couldn’t be completed. (ArcGIS.CertificateImportError error -26275.)"`

rolson: We could do the `LocalizedError` thing.  it will have to be done at some point when we localize the toolkit. Or you can just implement a `message` property on CertificateIMportError.  Either way would be preferable to the embedded `switch`.  Might be best to just do the right thing (`LocalizedError`) now while we are editing the code.

Tests/ArcGISToolkitTests/CertificatePickerViewModelTests.swift
njarecha: Ok, https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/245

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
njarecha: ok

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
njarecha: done

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
```diff
@@ -92,11 +103,38 @@ import ArcGIS
...
+        case .invalidPassword:
+            return NSLocalizedString("The password was invalid.", comment: "Invalid password")
+        default:
+            let errorString = SecCopyErrorMessageString(rawValue, nil) as? String ?? "The certificate file or password was invalid."
+            return NSLocalizedString(errorString, comment: "Other certificate error")
+        }
+    }
+}
+
+extension CertificatePickerViewModel.CertificateError {
```
rolson, line 128: this should also extend LocalizedError

Tests/ArcGISToolkitTests/CertificatePickerViewModelTests.swift
```diff
@@ -42,10 +43,27 @@ import XCTest
...
```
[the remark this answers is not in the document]
njarecha: done

rolson (approved):

dfeinzimer (dismissed):

rolson:
I'm ok with this when @philium is.  I don't know how localization works.

rolson (approved):

njarecha:
I've logged an [issue](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/247) move other strings in the Localizable.strings file.

dfeinzimer (approved):

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
```diff
@@ -18,6 +18,16 @@ import ArcGIS
 /// An object that provides the business logic for the workflow of prompting the user for a
 /// certificate and a password.
 @MainActor final class CertificatePickerViewModel: ObservableObject {
+    /// The types of certificate error.
+    enum CertificateError: Error {
```
philium, line 22: Should this be public? Also, it seems a bit odd for it to be nested under the model.

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
```diff
@@ -74,15 +84,16 @@ import ArcGIS
         Task.detached {
```
philium, line 84: Is there any way the enclosing method could be made `async` so as not to create a top-level task?

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
```diff
@@ -74,15 +84,16 @@ import ArcGIS
...
             do {
                 guard certificateURL.startAccessingSecurityScopedResource() else {
-                    await self.showCertificateImportError(nil)
+                    await self.showCertificateError(CertificateError.couldNotAccessCertificateFile)
```
philium, line 87: I realize that this code is just being modified, but I noticed that work is being done in the `else` clause of this guard statement, which often times makes code harder to reason about because it doesn't fit the mental model of guard statements. Consider restructuring to use an if…else statement:

```swift
if certificateURL.startAccessingSecurityScopedResource() {
    defer { certificateURL.stopAccessingSecurityScopedResource() }
    let credential: NetworkCredential = try .certificate(at: certificateURL, password: password)
    await self.challenge.resume(with: .continueWithCredential(credential))
} else {
    await self.showCertificateError(.couldNotAccessCertificateFile)
}
```

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
```diff
@@ -74,15 +84,16 @@ import ArcGIS
...
                     return
                 }
-
                 defer { certificateURL.stopAccessingSecurityScopedResource() }
 
                 await self.challenge.resume(with: .continueWithCredential(try .certificate(at: certificateURL, password: password)))
+            } catch(let certificateImportError as CertificateImportError) {
+                await self.showCertificateError(CertificateError.importError(certificateImportError))
```
philium, line 94:
```suggestion
                await self.showCertificateError(.importError(certificateImportError))
```

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
```diff
@@ -74,15 +84,16 @@ import ArcGIS
...
             } catch {
-                await self.showCertificateImportError(error)
+                await self.showCertificateError(CertificateError.other(error))
```
philium, line 96:
```suggestion
                await self.showCertificateError(.other(error))
```

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
```diff
@@ -92,11 +103,41 @@ import ArcGIS
         challenge.resume(with: .cancel)
     }
 
-    private func showCertificateImportError(_ error: Error?) async {
+    private func showCertificateError(_ error: CertificateError) async {
         // This is required to prevent an "already presenting" error.
         try? await Task.sleep(nanoseconds: 100_000)
```
philium, line 108: I think the proper solution is to annotate this method with `MainActor`.

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
```diff
@@ -92,11 +103,41 @@ import ArcGIS
...
+        case .invalidData:
+            return String(localized: "The certificate file was invalid.", bundle: .module)
+        case .invalidPassword:
+            return String(localized: "The password was invalid.", bundle: .module)
+        default:
+            let errorString = SecCopyErrorMessageString(rawValue, nil) as? String ?? String(
+                localized: "The certificate file or password was invalid.",
+                bundle: .module
+            )
+            return errorString
```
philium, line 126: Since `CFString` is bridgeable to `String`, the cast doesn't need to be conditional.

```suggestion
            return SecCopyErrorMessageString(rawValue, nil) as String? ?? String(
                localized: "The certificate file or password was invalid.",
                bundle: .module
            )
```

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
```diff
@@ -212,23 +253,11 @@ private extension View {
...
-        case .invalidData:
-            return "The certificate file was invalid."
-        case .invalidPassword:
-            return "The password was invalid."
-        default:
-            return defaultMessage
+            Text(viewModel.certificateError?.localizedDescription ?? String(
+                    localized: "The certificate file or password was invalid.",
+                    bundle: .module
+                 )
```
philium, line 259: Even though it isn't, the indentation really looks off here. A little refactoring would address that.

```suggestion
            Text(
                viewModel.certificateError?.localizedDescription ?? String(
                    localized: "The certificate file or password was invalid.",
                    bundle: .module
                )
```

Sources/ArcGISToolkit/Resources/en.lproj/Localizable.strings
```diff
@@ -0,0 +1,7 @@
+"Could not access the certificate file." = "Could not access the certificate file.";
+
+"The certificate file was invalid." = "The certificate file was invalid.";
+
+"The password was invalid." = "The password was invalid.";
+
+"The certificate file or password was invalid." = "The certificate file or password was invalid.";
```
philium, line 7: This file isn't needed. It would only be needed if the keys aren't the same as the values.

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
njarecha: done

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
njarecha: done

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
njarecha: done

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
njarecha: done

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
njarecha: The function is being called from the view modifier. If we make this function `async` then will have to move the `Task` in upper level too. I feel, it good to keep it here.

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
njarecha: Only `Authenticator` and it's modifier are public, everything else is internal. The `CertificateError` is used within `CertificatePickerViewModel` and while displaying using `alertCertificateError`. Should it still be not nested?

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
njarecha: I don't see `already presenting error` with or without `sleep` or `MainActor`. In fact, I see following before this prompt when we ask to input the password.

```
2023-01-23 14:45:44.896226-0800 AuthenticationExample[28329:387920] [DocumentManager] The view service did terminate with error: Error Domain=_UIViewServiceErrorDomain Code=1 "(null)" UserInfo={Terminated=disconnect method}
2023-01-23 14:45:45.005242-0800 AuthenticationExample[28329:387920] [LayoutConstraints] Changing the translatesAutoresizingMaskIntoConstraints property of a UICollectionViewCell that is managed by a UICollectionView is not supported, and will result in incorrect self-sizing. View: <_UIAlertControllerTextFieldViewCollectionCell: 0x13f4083f0; frame = (0 0; 270 24); gestureRecognizers = <NSArray: 0x600001df2520>; layer = <CALayer: 0x6000013beca0>>
2023-01-23 14:45:45.434019-0800 AuthenticationExample[28329:387920] [Presentation] Attempt to present <UIAlertController: 0x149177400> on <_TtGC7SwiftUI19UIHostingControllerGVS_15ModifiedContentVS_7AnyViewVS_12RootModifier__: 0x149027800> (from <UIViewController: 0x13f228090>) which is already presenting <UIAlertController: 0x13d812800>.
```
Simulator Screen Shot - iPhone 14 Pro - 2023-01-23 at 14 45 53

Logged an [issue](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/248) to look into this.

Sources/ArcGISToolkit/Resources/en.lproj/Localizable.strings
njarecha: The exposing string for localization is not working for the toolkit package. See below error. We'll need this file to request the localized strings from localization team. I would suggest keeping this file until we figure out the export issue. I can remove the file if you still feel strongly.  

Screen Shot 2023-01-23 at 2 58 08 PM

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
philium: So users will never see an instance of `CertificateError`? If not, then it is fine where it is.

njarecha: Yes, they will see only error string in the alert.

Sources/ArcGISToolkit/Resources/en.lproj/Localizable.strings
philium: I'm fine with leaving it, but I can foresee it easily becoming out-of-date (new localized string added in code but not in the strings file). 

That localization failure appears to be an Xcode bug. If you look at the build logs, it appears to be building for an unsupported platform (macOS). We should probably submit a feedback report to Apple about that.

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
philium: So the sleep isn't needed? If not, we should remove it.

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
philium: Probably outside the scope of this PR, but perhaps the view modifier could be refactored to use a `task` modifier.

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
```diff
@@ -92,11 +103,40 @@ import ArcGIS
...
+            return SecCopyErrorMessageString(rawValue, nil) as String? ?? String(
+                localized: "The certificate file or password was invalid.",
+                bundle: .module
+            )
+        }
+    }
+}
+
+extension CertificatePickerViewModel.CertificateError: LocalizedError {
+    public var errorDescription: String? {
```
philium, line 131:
```suggestion
    var errorDescription: String? {
```

Sources/ArcGISToolkit/Resources/en.lproj/Localizable.strings
philium: In fact, their file is already incomplete. It is missing the title of the certificate error alert, the button titles, etc.

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
njarecha: `public` has to be there because it is a requirement of `LocalizedError`.

Sources/ArcGISToolkit/Resources/en.lproj/Localizable.strings
njarecha: Yes, I've logged an issue for it.
https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/247

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
njarecha: That was added probably because someone was seeing the error. So I kept as it is and added information in the [issue](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/248) to look into and remove if not needed.

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
njarecha: Logged an [issue](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/249)

Sources/ArcGISToolkit/Resources/en.lproj/Localizable.strings
njarecha: > That localization failure appears to be an Xcode bug. If you look at the build logs, it appears to be building for an unsupported platform (macOS). We should probably submit a feedback report to Apple about that.

We'll have to create a repro project to log an issue. I'll work on it.

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
philium: That's not correct. The declaration needs to match the access level of the type, which is internal, so the declaration itself can be internal.

njarecha: you are correct. It was giving error for `CertificateImportError`'s `errorDescription`.

philium (approved):

rolson (approved):

njarecha:
Thanks @philium, @rolson and @dfeinzimer!

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
```diff
@@ -18,6 +18,16 @@ import ArcGIS
 /// An object that provides the business logic for the workflow of prompting the user for a
 /// certificate and a password.
 @MainActor final class CertificatePickerViewModel: ObservableObject {
+    /// The types of certificate error.
+    enum CertificateError: Error {
+        /// Could not access the certificate file.
+        case couldNotAccessCertificateFile
```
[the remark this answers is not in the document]
njarecha: Ok with `failedToAccessCertificateFile`. However, there is a localized string for the error and we'll have to update it and get new translation too. 

`Could not access the certificate file.`

This PR was merged long time back but issue did not get close.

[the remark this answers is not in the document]
rolson: Its fine. Maybe something to change next release since it's internal API it won't break users.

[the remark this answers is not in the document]
njarecha: Ok, created [issue](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/636).

PR #252, opened 2023-01-27
Latest API changes

yo1995:
Use `Int` instead of `Int32` in the `SearchSource` related types.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Popups/PopupMedia/Charts/ChartData.swift
+++ b/Sources/ArcGISToolkit/Components/Popups/PopupMedia/Charts/ChartData.swift
@@ -28,9 +28,7 @@ struct ChartData: Identifiable {
     ///   - value: The value of the data.
     init(label: String, value: Any) {
         self.label = label
-        if let int32 = value as? Int32 {
-            self.value = Double(int32)
-        } else if let int = value as? Int {
+        if let int = value as? Int {
             self.value = Double(int)
         } else if let float = value as? Float {
             self.value = Double(float)
--- a/Documentation/Search/README.md
+++ b/Documentation/Search/README.md
@@ -40,10 +40,10 @@ public protocol SearchSource {
     var name: String { get set }
 
     /// The maximum results to return when performing a search. Most sources default to `6`.
-    var maximumResults: Int32 { get set }
+    var maximumResults: Int { get set }
 
     /// The maximum suggestions to return. Most sources default to `6`.
-    var maximumSuggestions: Int32 { get set }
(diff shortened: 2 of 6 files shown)


Sources/ArcGISToolkit/Components/Popups/PopupMedia/Charts/ChartData.swift
```diff
@@ -28,9 +28,7 @@ struct ChartData: Identifiable {
     ///   - value: The value of the data.
     init(label: String, value: Any) {
         self.label = label
-        if let int32 = value as? Int32 {
```
yo1995, line 31: I'm not sure if this needs to be removed. Since there won't be `Int32` that gets exposed to the users, I think it might be OK to remove, but let me know if it has other usage!

dfeinzimer: I think this change is okay

dfeinzimer (approved):
Changes look good, tests pass and the search and popup examples appear to continue to function normally. Thank you for the updates!

Sources/ArcGISToolkit/Components/Popups/PopupMedia/Charts/ChartData.swift
mhdostal: @zkline101 The `value` in use here is coming from Core. Are all Core values converted to Int? Or can they still be `Int32`.

The value is essentially a numeric field in a table.

zkline101: > Are all Core values converted to Int? 

If the core value is exposed as an `Any` but the underlying type is an integer then yes that underlying type will be exposed to us as an `Int`. Does that help?

mhdostal: @zkline101 Yes, that helps a lot, thanks.  @yo1995 It's OK to remove that check for `Int32`, as they all should be `Int`s.

mhdostal (approved):

Sources/ArcGISToolkit/Components/Popups/PopupMedia/Charts/ChartData.swift
yo1995: Thanks for the clarification.

rolson:
are there any other int32 or int64 references in the toolkit?

dfeinzimer:
> are there any other int32 or int64 references in the toolkit?

I wasn't able to find any others when I searched

PR #264, opened 2023-02-21
Floor Filter Bug Fixes

### LINKED ISSUE #244: Floor Filter - Two close buttons can appear
The Swift Samples app uses the Floor Filter component with data that appears to cause two close buttons to appear.

Screenshot 2023-01-09 at 11 59 54

### LINKED ISSUE #263: [Floor Filter Component] Floor filter does not work or can't select a facility if Sites layer is not present within the map.
Sites layer is optional in Floor Manager. When Sites layer is missing in the map, Floor Filter just presents a blank Sites filter. 

There is no way to choose Facility and hence Floor Filter cannot be used further.

Simulator Screen Shot - iPhone 14 - 2023-02-17 at 13 46 23

@dfeinzimer - FYI

dfeinzimer:
Closes #244 
Closes #263

### DIFF
--- a/Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
+++ b/Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
@@ -116,12 +116,26 @@ struct SiteAndFacilitySelector: View {
                     text: $query,
                     placement: .navigationBarDrawer(displayMode: .always),
                     prompt: "Filter sites"
+                allSitesButton
+        
+        /// The "All sites" button.
+        ///
+        /// This button presents the facilities list in a special format where the facilities list
+        /// shows every facility in every site within the floor manager.
+        var allSitesButton: some View {
+            NavigationLink("All sites") {
+                FacilitiesList(
+                    usesAllSitesStyling: true,
+                    facilities: viewModel.sites.flatMap(\.facilities),
+                    isHidden: isHidden
                 )
                 .keyboardType(.alphabet)
                 .disableAutocorrection(true)
                 .navigationTitle("Sites")
             }
             .navigationViewStyle(.stack)
+            .buttonStyle(.bordered)
+            .padding([.bottom], horizontalSizeClass == .compact ? 5 : 0)
         }
         
         /// A view containing a list of the site names.
--- a/Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
+++ b/Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
@@ -192,10 +192,11 @@ struct SiteAndFacilitySelector: View {
         var matchingFacilities: [FloorFacility] {
             guard !query.isEmpty else {
                 return facilities
+                    .sorted { $0.name < $1.name }
             }
-            return facilities.filter {
-                $0.name.localizedStandardContains(query)
-            }
+            return facilities
+                .filter { $0.name.localizedStandardContains(query) }
+                .sorted { $0.name < $1.name  }
         }
         
         var body: some View {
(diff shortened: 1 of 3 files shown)


Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
```diff
@@ -29,7 +29,30 @@ struct SiteAndFacilitySelector: View {
...
+        NavigationView {
+            Group {
+                // If there's more than one site
+                if viewModel.sites.count > 1 {
+                    // Show the list of sites for site selection
+                    SitesList(isHidden: isHidden)
+                } else {
+                    // Otherwise there're no sites or only one site, show the list of facilities
+                    FacilitiesList(
+                        usesAllSitesStyling: false,
```
mansibarodia10057, line 41: Design question - Should the same change be made for facilities too i.e. if there is just one facility, skip showing the facility picker and directly show levels picker?

mansibarodia10057:
In floor manager, looks like even facility layer is also optional. Am not sure if there would be a map without facilities in it, if there is then how would floor filter behave in that case? Same goes for levels layer. 

https://developers.arcgis.com/swift/api-reference/documentation/arcgis/floormanager/

Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
dfeinzimer: > Design question - Should the same change be made for facilities too i.e. if there is just one facility, skip showing the facility picker and directly show levels picker? 

Good question! The design guideline (which all of the toolkit platforms adhere to) says:
> If there is only 1 facility, select it but still show the facilities browser.

[Source](https://devtopia.esri.com/runtime/common-toolkit/blob/main/designs/FloorFilter/FloorFilter.md#:~:text=If%20there%20is%20only%201%20facility%2C%20select%20it%20but%20still%20show%20the%20facilities%20browser.)

That being said, if you'd like, I'd be happy to provide this feedback to the toolkit team for consideration.

> If there is just one facility and one level, instead of showing any pickers, just shows the level directly on floor picker.

I believe this too would fall under the design guideline mentioned above.

> This also applies incase if there is not facility and just one level in the map.

I believe the design also covers this scenario.

> It is however not possible to have a Level without a Facility.

[Source](https://devtopia.esri.com/runtime/common-toolkit/blob/main/designs/FloorFilter/FloorFilter.md#:~:text=It%20is%20however%20not%20possible%20to%20have%20a%20Level%20without%20a%20Facility.)

dfeinzimer:
> In floor manager, looks like even facility layer is also optional. Am not sure if there would be a map without facilities in it

Based on this guideline:

> It is however not possible to have a Level without a Facility. [Source](https://devtopia.esri.com/runtime/common-toolkit/blob/main/designs/FloorFilter/FloorFilter.md#:~:text=It%20is%20however%20not%20possible%20to%20have%20a%20Level%20without%20a%20Facility.)

I'm not sure that'd make sense to have floor manager data in a map and not have facilities. I will double check on this with the feature team.

Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
mansibarodia10057: Yes, please get a feedback from the toolkit team. It's going to be less taps/selection for a user.

dfeinzimer: Toolkit common # 208

dfeinzimer:
> I'm not sure that'd make sense to have floor manager data in a map and not have facilities. I will double check on this with the feature team.

Following up on this, `FloorManager`'s documentation states:
> The `FloorManager` will fail to load if:
> ...
> * The mandatory facility or level layer doesn't exist.
> * It is unable to get the facilities or levels from their respective layers.

Therefore, there shouldn't be a map without facilities in it, but if there is, something has gone wrong.

Also note that the floor filter component mainly refers to `FloorManager.getter:facilities` (which isn't nullable) and  `FloorManager.getter:facilityLayer` in only one location where a default value is provided in the event that the layer is `nil`. If the facilities list is empty, the user will simply see an empty list.

mansibarodia10057 (approved):
Tested for both the bugs. Fixes looks good. Thanks!

mhdostal (approved):

PR #272, opened 2023-03-08
Build fix for `Bookmarks`

dfeinzimer:
When building `v.next` with the latest daily binaries, a build error arises in `Bookmarks.swift` that isn't present with building with source.

This area hasn't changed since beta and so I presume the compiler has become more strict in this area.

This can be tested by switching the SDK dependency to the Swift daily repo.


yo1995 (approved):

dfeinzimer:
Thanks @yo1995!

Issue #273, opened 2023-03-09
Need a way to programmatically set the selected site, facility, level


yo1995:
The Indoors location data source needs to be able to set the selected site/floor/level when the user moves through/around a building.

The goal is to keep the floor filter component synchronized with the Indoors positioning system.

From Morakat and the indoors location team at the DevSummit.

Morakat added `setSite`, `setFacility`, and `setFloor` to the model, which works for him. The code is in his fork.

yo1995:
cc: @dfeinzimer @mhdostal

mhdostal:
Related to #271

PR #277, opened 2023-03-22
[Floor Filter] Add `.selection(:FloorFilterSelection)` modifier

### LINKED ISSUE #271: [Floor Filter Component] Exposing sites, facilities and levels read-only info.
Currently there is no way of getting sites, facilities and levels info from floor filter, especially the selected site, selected facility and selected level info. Consuming apps will need this info to keep a track of user's location if the app deals with location related functionality. 

Looking at the current floor filter toolkit code, looks like floor filter view model has all the selection details, but the properties are private. These details can be read-only when exposed.

Please allow access to the details of the sites, facilities and levels in the map especially selected site, selected facility and selected level since other details can also be obtained from floor manager. 

@dfeinzimer - FYI.

dfeinzimer:
Resolves #271, #273

`FloorFilter.selection(_:)` will allow a user of the component to set the active selection and also monitor it for changes.

### DIFF
--- a/Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
+++ b/Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
@@ -180,8 +180,8 @@ public struct FloorFilter: View {
             // Prevent a double-set if the view model triggered the original change.
             guard newValue != viewModel.selection else { return }
             switch newValue {
-            case .site(let site): viewModel.setSite(site)
-            case .facility(let facility): viewModel.setFacility(facility)
+            case .site(let site): viewModel.setSite(site, zoomTo: true)
+            case .facility(let facility): viewModel.setFacility(facility, zoomTo: true)
             case .level(let level): viewModel.setLevel(level)
             case .none: viewModel.clearSelection()
             }
--- a/Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
+++ b/Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
@@ -50,9 +50,6 @@ public struct FloorFilter: View {
     /// The view model used by the `FloorFilter`.
     @StateObject private var viewModel: FloorFilterViewModel
     
-    /// A Boolean value that indicates whether the levels view is currently collapsed.
-    @State private var isLevelsViewCollapsed = false
-    
     /// A Boolean value that indicates whether the site and facility selector is presented.
     @State private var isSitesAndFacilitiesHidden = true
     
--- a/Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
+++ b/Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
@@ -179,7 +179,12 @@ public struct FloorFilter: View {
         .onChange(of: selection?.wrappedValue) { newValue in
             // Prevent a double-set if the view model triggered the original change.
             guard newValue != viewModel.selection else { return }
-            viewModel.selection = newValue
+            switch newValue {
+            case .site(let site): viewModel.setSite(site)
+            case .facility(let facility): viewModel.setFacility(facility)
+            case .level(let level): viewModel.setLevel(level)
+            case .none: viewModel.clearSelection()
+            }
         }
         .onChange(of: viewModel.selection) { newValue in
             // Prevent a double-set if the user triggered the original change.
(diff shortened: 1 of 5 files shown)


Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
```diff
@@ -50,12 +50,12 @@ public struct FloorFilter: View {
     /// The view model used by the `FloorFilter`.
     @StateObject private var viewModel: FloorFilterViewModel
     
-    /// A Boolean value that indicates whether the levels view is currently collapsed.
-    @State private var isLevelsViewCollapsed = false
-
```
dfeinzimer, line 55: This property was unused

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilterViewModel.swift
```diff
@@ -51,15 +41,15 @@ final class FloorFilterViewModel: ObservableObject {
...
+    @Published private(set) var selection: FloorFilterSelection?
     
     // MARK: Constants
     
     /// The selection behavior of the floor filter.
     private let automaticSelectionMode: FloorFilterAutomaticSelectionMode
     
     /// The `FloorManager` containing the site, floor, and level information.
-    let floorManager: FloorManager
+    private let floorManager: FloorManager
```
dfeinzimer, line 52: No reason this needed to be internal.

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilterViewModel.swift
```diff
@@ -18,16 +18,6 @@ import SwiftUI
...
 final class FloorFilterViewModel: ObservableObject {
-    ///  A selected site, floor, or level.
-    enum Selection: Hashable {
-        /// A selected site.
-        case site(FloorSite)
-        /// A selected facility.
-        case facility(FloorFacility)
-        /// A selected level.
-        case level(FloorLevel)
-    }
```
dfeinzimer, line 29: Because this type is no longer used by only the model and is now public, it made sense to un-nest it and give it a new name, `FloorFilterSelection`, similar to `FloorFilterAutomaticSelectionMode`.

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
```diff
@@ -176,5 +176,29 @@ public struct FloorFilter: View {
         .frame(minHeight: 100)
         .environmentObject(viewModel)
         .disabled(viewModel.isLoading)
+        .onChange(of: selection?.wrappedValue) { newValue in
+            // Prevent a double-set if the view model triggered the original change.
+            guard newValue != viewModel.selection else { return }
+            switch newValue {
+            case .site(let site): viewModel.setSite(site, zoomTo: true)
+            case .facility(let facility): viewModel.setFacility(facility, zoomTo: true)
+            case .level(let level): viewModel.setLevel(level)
```
dfeinzimer, line 185: One thing that's unclear to me: should the viewpoint be updated automatically when the user updates the selection? It seems like yes but I think that'd be up for debate. It could also be configurable through another modifier.

dfeinzimer:
@mansibarodia10057, @mpEsri, please review and indicate if this solves your use cases.

mhdostal (dismissed):

dfeinzimer:
> In this case, the floor filter itself can just take the binding as a parameter to the initializer. That would be best for a value like this.

Does that suggestion hold even if the use case for this is outside the scope of the common design? Would a secondary convenience initializer be the best path?

philium:
> Would a secondary convenience initializer be the best path?

That's fine if that's what makes the most sense. `List`, for example, has some initializers that take a selection binding and other that don't.

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
mpEsri: Q1) Viewpoint update: I think we don't need to update the viewpoint automatically until it becomes clear if viewpoint automatic update is common behavior.
Q2) I think it will be rather annoying if the site and facility are auto expanded. I think keeping them closed is an acceptable behavior, but the values should be auto updated if they are already expanded.

dfeinzimer: Great, thank you for that feedback @mpEsri!

mhdostal (approved):

dfeinzimer:
> You might consider taking the convenience properties [here](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/fb8b3e439537fe7bf38e4325e4ac781f9f61ea0a/Sources/ArcGISToolkit/Components/FloorFilter/FloorFilterViewModel.swift#L82-L115) and moving them into a public extension of FloorFilterSelection as a convenience for getting the site, facility, or level

Great idea @philium, done. That seems like a much more appropriate location for them.

mhdostal (approved):

dfeinzimer:
I've gotten confirmation from Mansi that this satisfies her originating issue. Merging.

PR #279, opened 2023-03-23
[Floor Filter] Add `levelSelectorWidth(_:)` modifier

### LINKED ISSUE #258: [Floor Filter] - Allow to adjust default floor filter width and color.
Currently minimum [floor filter width](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/main/Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift#L61) is set to 60. Floor filter still looks big when opened on smaller devices. Also since the app supports branding colors set via MDM, where the user can set the theme any other color other than standard blue. 

Please make changes, where the apps can set minimum width for a floor filter and also change the color of the floor filter to match other components in the app.

dfeinzimer:
Closes #258 

- Adds a new `levelSelectorWidth(_:)` modifier to the floor filter, allowing a user to control the width of the level selector
- Adjusts how an app's accent color is used in styling the component.


dfeinzimer:
@mansibarodia10057, please review and determine if this satisfies your use cases.

Sources/ArcGISToolkit/Components/FloorFilter/LevelSelector.swift
```diff
@@ -30,104 +33,115 @@ struct LevelSelector: View {
...
-                                set: { newIsOn in
-                                    guard newIsOn else { return }
-                                    viewModel.setLevel(level)
-                                }
-                            )
-                        ) {
-                            Text(level.shortName)
-                                .modifier(LevelNameFormat())
-                        }
-                        .toggleStyle(.selectableButton)
```
dfeinzimer, line 105: I found that SwiftUI's default button styles seem to be opinionated about their minimum width. This was going to cause issues if users set the width below what the styles allowed. After much experimentation I found that simply using `Text` with a rounded rectangle background was the best way forward. It also allowed for the removal of this computed binding and some custom toggle styles we'd defined to make it work correctly.

mansibarodia10057 (approved):
Looks good. Also tested it in our app.

PR #280, opened 2023-03-24
Add DocC starter

dfeinzimer:
I think we can make great use of DocC in the toolkit to enhance and keep our existing documentation. I'll write more about that more in a Swift repo issue.

For the time being, I figure we can make this quick addition which'll serve as an starting point for DocC work.

Before:
before
After:
after

This brings us in line with the SDK itself:
sdk


dfeinzimer (commented):
~

Sources/ArcGISToolkit/Documentation.docc/ArcGISToolkit.md
```diff
@@ -0,0 +1,3 @@
+# ``ArcGISToolkit``
+
+The ArcGIS Maps SDK for Swift Toolkit contains components that will simplify your iOS app development. It is built off of the new ArcGIS Maps SDK for Swift.
```
dfeinzimer, line 3: This line borrowed from the repo's readme

philium (approved):

PR #282, opened 2023-03-29
Utility Network Trace - Improve identification

### LINKED ISSUE #269: Utility Network - Only identify layers within the network
When a location is tapped, the utility network tool will run an identify operation on all layers. Instead, it should only use layers that are within the utility network. This could potentially be a contributing factor to the behavior exhibited in #227.

Note: via the UN team, the best way to access the set of all layers in a utility network is network -> definition -> sources -> featureTable -> layer

Reported by @rolson

dfeinzimer:
Closes #269 

- Refactors `UtilityNetworkTraceViewModel.addStartingPoint(at:mapPoint:with:)` into a new func `addStartingPoints(at:mapPoint:with:)` where we only run an identify operation on the layers within the utility network.
- The method is renamed to a plural form because it has the potential to add multiple starting points.

### DIFF
--- a/Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTraceViewModel.swift
+++ b/Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTraceViewModel.swift
@@ -132,7 +132,7 @@ import SwiftUI
         
         var identifyLayerResults = [IdentifyLayerResult]()
         
-        for layer in layers ?? [] {
+        for layer in network?.layers ?? [] {
             if let r = await identify(layer, point) {
                 identifyLayerResults.append(r)
             }
@@ -147,10 +147,6 @@ import SwiftUI
                 await processAndAdd(startingPoint: startingPoint)
             }
         }
-    }
-    
-    var layers: [Layer]? {
-        network?.definition?.networkSources.compactMap { $0.featureTable.layer }
     }
     
     /// Deletes all of the completed traces.
@@ -519,5 +515,11 @@ extension UtilityNetworkTraceViewModel {
             fractionalLengthClosestTo: point,
             tolerance: 10
         )
+    }
+}
+
+extension UtilityNetwork {
+    var layers: [Layer] {
+        definition?.networkSources.compactMap { $0.featureTable.layer } ?? []
     }
 }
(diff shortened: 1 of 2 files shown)


Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTraceViewModel.swift
```diff
@@ -115,27 +115,22 @@ import SwiftUI
...
+    func addStartingPoints(at point: CGPoint, mapPoint: Point, with proxy: MapViewProxy) async {
+        for layer in network?.layers ?? [] {
+            if let result = try? await proxy.identify(on: layer, screenPoint: point, tolerance: 10) {
+                for element in result.geoElements {
+                    await processAndAdd(
+                        startingPoint: UtilityNetworkTraceStartingPoint(geoElement: element, mapPoint: mapPoint)
+                    )
+                }
             }
         }
```
mhdostal, line 135: Currently, calling `identify` on each layer will wait until `identify` is done before moving on to the next layer. You can use a `TaskGroup` to perform all identifies at the same time.  I've used it once before; I'm not sure if the `try? await group.waitForAll()` is necessary (or i you need another one in the inner task group); also, I did not make sure this compiles.

I thought about another task group around the `await processAndAdd` for loop, but there's a lot of stuff happening inside there, including setting the error alert, and I'm not sure what the consequences of performing those all asynchronously would be, but it may be worth a shot.

```suggestion
        await withTaskGroup(of: Void.self) { [weak self] taskGroup in
            guard let self else { return }
            for layer in network?.layers ?? [] {
                taskGroup.addTask {
                    if let result = try? await proxy.identify(on: layer, screenPoint: point, tolerance: 10) {
                        for element in result.geoElements {
                            await processAndAdd(
                                startingPoint: UtilityNetworkTraceStartingPoint(geoElement: element, mapPoint: mapPoint)
                            )
                        }
                    }
                }
            }
            // Make sure that all tasks complete.
            try await group.waitForAll()
        }
```

mhdostal (commented):
Looks good, just the TaskGroup comment/suggestion.

Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTraceViewModel.swift
dfeinzimer: Good suggestion, applied, thank you!

mhdostal (approved):

PR #283, opened 2023-03-30
[Basemap Gallery] Make responsive to environment

### LINKED ISSUE #265: BasemapGallery should conform to the view it's placed in
I tried to use the basemap gallery in a sheet rather than an overlay because the code is quite nice for a sheet and I prefer that look:

```swift
            .sheet(isPresented: $showBasemapGallery) {
                BasemapGallery(items: basemaps(), geoModel: map)
                    .padding(.init(top: 175, leading: 5, bottom: 5, trailing: 15))
            }
```

I added that to Jen's swift-demo-dev-summit project.

This is how it looked:

image

Two things I think should be improved:

1. The treatment (shadow, etc) around the basemap gallery should go away.  We don't want to prescribe a treatment or style. That can be added app level.
2. The gallery should conform to the view that it's placed in.  It should fill the sheet if it's placed in a sheet.

dfeinzimer:
Closes #265 

- [x] Open question: should the static 3 columns for grid layouts be maintained?
  - This PR will make the number of grid columns dynamically computed based on the available width.

### DIFF
--- a/Examples/Examples/BasemapGalleryExampleView.swift
+++ b/Examples/Examples/BasemapGalleryExampleView.swift
@@ -37,9 +37,18 @@ struct BasemapGalleryExampleView: View {
             // Display in a sheet
             //
             .sheet(isPresented: $showBasemapGallery) {
-                BasemapGallery(items: basemaps, geoModel: map)
-                    .style(.grid)
-                    .padding()
+                VStack {
+                    Button {
+                        showBasemapGallery.toggle()
+                    } label: {
+                        Image(systemName: "xmark")
+                    }
+                    .padding([.top, .trailing])
+                    .frame(maxWidth: .infinity, alignment: .trailing)
+                    BasemapGallery(items: basemaps, geoModel: map)
+                        .style(.grid)
+                        .padding()
+                }
             }
             
             // Display in a floating panel
--- a/Examples/Examples/BasemapGalleryExampleView.swift
+++ b/Examples/Examples/BasemapGalleryExampleView.swift
@@ -33,39 +33,15 @@ struct BasemapGalleryExampleView: View {
     
     var body: some View {
         MapView(map: map, viewpoint: initialViewpoint)
-            
-            // Display in a sheet
-            //
             .sheet(isPresented: $showBasemapGallery) {
-                VStack {
+                VStack(alignment: .trailing) {
                     doneButton
-                        .padding([.top, .trailing])
-                        .frame(maxWidth: .infinity, alignment: .trailing)
+                        .padding()
                     BasemapGallery(items: basemaps, geoModel: map)
-                        .style(.grid)
+                        .style(.automatic())
                         .padding()
                 }
             }
-            
-            // Display in a floating panel
-            //
-//            .floatingPanel(isPresented: $showBasemapGallery) {
-//                BasemapGallery(items: basemaps, geoModel: map)
-//                    .style(.grid)
-//            }
-            
-            // Display in an overlay
-            //
-//            .overlay(alignment: .topTrailing) {
-//                if showBasemapGallery {
-//                    BasemapGallery(items: basemaps, geoModel: map)
-//                        .style(.list)
-//                        .frame(width: 150, height: 400)
-//                        .esriBorder()
-//                        .padding()
-//                }
-//            }
-            
             .toolbar {
                 ToolbarItem(placement: .navigationBarTrailing) {
                     Toggle(isOn: $showBasemapGallery) {
--- a/Examples/Examples/BasemapGalleryExampleView.swift
+++ b/Examples/Examples/BasemapGalleryExampleView.swift
@@ -42,7 +42,6 @@ struct BasemapGalleryExampleView: View {
                         .padding()
                 }
             }
-            .navigationTitle("Basemap Gallery")
             .toolbar {
                 ToolbarItem(placement: .navigationBarTrailing) {
                     Toggle(isOn: $showBasemapGallery) {
--- a/Examples/Examples/BasemapGalleryExampleView.swift
+++ b/Examples/Examples/BasemapGalleryExampleView.swift
@@ -35,13 +35,32 @@ struct BasemapGalleryExampleView: View {
     
     var body: some View {
         MapView(map: dataModel.map, viewpoint: initialViewpoint)
-            .overlay(alignment: .topTrailing) {
-                if showBasemapGallery {
-                    BasemapGallery(items: basemaps, geoModel: dataModel.map)
-                        .style(.automatic())
-                        .padding()
-                }
+            
+            // Display in a sheet
+            //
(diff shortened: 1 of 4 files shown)


Examples/Examples/BasemapGalleryExampleView.swift
```diff
@@ -34,15 +32,35 @@ struct BasemapGalleryExampleView: View {
...
-                        .style(.automatic())
-                        .padding()
-                }
+        MapView(map: map, viewpoint: initialViewpoint)
+            
+            // Display in a sheet
+            //
+            .sheet(isPresented: $showBasemapGallery) {
+                BasemapGallery(items: basemaps, geoModel: map)
+                    .style(.grid)
```
dfeinzimer, line 41: `.style(.grid)` is here to prevent the list style from being auto-selected on iPhone's, which would consume the entire screen width.

Sources/ArcGISToolkit/Components/BasemapGallery/BasemapGallery.swift
```diff
@@ -25,11 +25,11 @@ public struct BasemapGallery: View {
...
         /// Defaults to `125` when displayed as a list, `300` when displayed as a grid.
-        case automatic(listWidth: CGFloat = 125, gridWidth: CGFloat = 300)
+        case automatic
         /// The `BasemapGallery` will display as a grid. Defaults to `300`.
-        case grid(width: CGFloat = 300)
+        case grid
         /// The `BasemapGallery` will display as a list. Defaults to `125`.
-        case list(width: CGFloat = 125)
+        case list
     }
```
dfeinzimer, line 33: The associated values are no longer needed because we can just use the width from the geometry reader now.

Sources/ArcGISToolkit/Components/BasemapGallery/BasemapGallery.swift
```diff
@@ -54,19 +54,19 @@ public struct BasemapGallery: View {
...
-        viewModel = BasemapGalleryViewModel(geoModel, portal: portal)
+        _viewModel = StateObject(wrappedValue: BasemapGalleryViewModel(geoModel, portal: portal))
     }
     
     /// The view model used by the view. The `BasemapGalleryViewModel` manages the state
     /// of the `BasemapGallery`. The view observes `BasemapGalleryViewModel` for changes
     /// in state. The view updates the state of the `BasemapGalleryViewModel` in response to
     /// user action.
-    @ObservedObject private var viewModel: BasemapGalleryViewModel
+    @StateObject private var viewModel: BasemapGalleryViewModel
```
dfeinzimer, line 64: Since the model is owned by the gallery view, I think this should be a `StateObject` instead.

Sources/ArcGISToolkit/Components/BasemapGallery/BasemapGallery.swift
```diff
@@ -117,10 +105,11 @@ public struct BasemapGallery: View {
                 } message: { item in
                     Text(item.message)
                 }
-                .frame(height: min(contentHeight, geometry.size.height))
-                .esriBorder()
+                .frame(
+                    width: geometry.size.width,
+                    height: geometry.size.height
```
dfeinzimer, line 107: Use the full height available so content isn't trimmed while scrolling.

Before:


After:

dfeinzimer:
@rolson It would be great if you can confirm this resolves your original request (#265)

rolson:
based on the screen shots above, yes it meets the requirements I was looking for.

Examples/Examples/BasemapGalleryExampleView.swift
```diff
@@ -52,6 +75,17 @@ struct BasemapGalleryExampleView: View {
...
     }
     
+    /// A button that allows a user to close a sheet.
+    ///
+    /// This is especially useful for when the sheet is open an iPhone in landscape.
+    private var xButton: some View {
+        Button {
+            showBasemapGallery.toggle()
+        } label: {
+            Image(systemName: "xmark")
```
mhdostal, line 85: In the other sheets in the toolkit a "Done" button is used (in the Popup) and in the FloorFilter, a `xmark.circle` image is used.  The HIG states: _Position Done and Cancel buttons as people expect. Typically, a Done or Dismiss button belongs in a sheet’s top-right corner (in a left-to-right layout) or top-left corner (in a right-to-left layout)._

```suggestion
            Text("Done")
```

mhdostal (changes requested):

rolson:
> I'm OK with making the grid dynamic.

I think making it dynamic is a good call.

dfeinzimer:
I've made changes to make the grid width dynamically computed.

Using a floating panel:
https://user-images.githubusercontent.com/16397058/230230718-e2a9ce05-f948-48ea-8995-5a9928911ab5.mp4

Using an overlay:
https://user-images.githubusercontent.com/16397058/230230739-e28df193-f7fb-4dd1-a2c7-b41d9390c837.mp4

Using a sheet:
sheetOniPad

Sources/ArcGISToolkit/Components/BasemapGallery/BasemapGallery.swift
```diff
@@ -117,43 +102,51 @@ public struct BasemapGallery: View {
...
+    func makeGridView(_ containerWidth: CGFloat, _ maxItemWidth: CGFloat) -> some View {
         internalMakeGalleryView(
             columns: Array(
                 repeating: GridItem(
                     .flexible(),
                     alignment: .top
                 ),
-                count: 3
+                count: max(
+                    1,
```
mhdostal, line 147: I would make this `2`, so it would always be a grid. Otherwise the `grid` style could look like a list.
```suggestion
                    2,
```

mhdostal (changes requested):
Thoughts on the one suggestion?

Sources/ArcGISToolkit/Components/BasemapGallery/BasemapGallery.swift
dfeinzimer: I'd be fine with 2 if you prefer, my only concern is that things can get a bit cramped if automatic is used and the component is provided little width.

The following code produces the results below:
```swift
.overlay(alignment: .trailing) {
    if showBasemapGallery {
        BasemapGallery(items: basemaps, geoModel: map)
            .style(.automatic(maxGridItemWidth: 200))
            .esriBorder()
            .padding()
            .frame(width: width)
    }
}
```

`max(2,...)` with a total width of 140 provided:
max2

`max(1,...)` with a total with of 140 provided:
max1

mhdostal (approved):
I'm OK with `1` for grid column min.

Sources/ArcGISToolkit/Components/BasemapGallery/BasemapGallery.swift
mhdostal: I'm Ok with 1.

Issue #260, opened 2023-02-16
Integrate new animated viewpoint functionality


dfeinzimer:
Integrate the newly added viewpoint animation functionality into toolkit components where applicable.

- The compass is probably a good starting point for this.

dfeinzimer:
Waiting on outcome of Swift 3882

PR #298, opened 2023-04-11
[Compass] Remove `action` parameter, add `mapViewProxy` parameter, support rotation animation

### LINKED ISSUE #295: Revise Compass animation
See also: Swift 3882

We've come to the determination that the best path forward for now is to pass proxies to toolkit components where needed.

In the case of the compass, we can take this a step further and explore a compass view modifier which will create its own reader/proxy and make compass usage very straightforward.

dfeinzimer:
- Closes #295 
- In Swift 3882, a conclusion was reached that toolkit components should take proxy parameters where needed.
- In the case of the Compass, a map view proxy is used to animate the map view's heading to zero.
- SceneView compass animations will not be supported.
- See also #260

### DIFF
--- a/Documentation/Compass/README.md
+++ b/Documentation/Compass/README.md
@@ -24,8 +24,8 @@ Compass:
     /// direction toward true East, etc.).
     /// - Parameters:
     ///   - heading: The heading of the compass.
-    ///   - action: An action to perform when the compass is tapped.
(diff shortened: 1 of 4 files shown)


mhdostal (dismissed):
Very good.  It makes the Compass more intuitive and easier to use. Passing in the proxy is the way to go.

rolson (approved):

mhdostal (approved):

PR #300, opened 2023-04-11
[Search] Add `geoViewProxy` parameter, support viewpoint animation

dfeinzimer:
Related #260 

- Adds a new `geoViewProxy` parameter to the search module. This is used to perform animated viewpoints now.
- In the example, a map view proxy is added to demonstrate usage of the new parameter.


mhdostal (approved):
Works like it should have from the start!

PR #302, opened 2023-04-11
Fix `AuthenticatorTests` compile failure

zkline101:
The toolkit is released with the closed-source SDK so we cannot use the `@testable` attribute.


mhdostal (approved):

dfeinzimer (approved):
Great this works without testable!

Issue #152, opened 2022-09-27
Review and remove or edit `Utility*` conformances


dfeinzimer:
As mentioned in Swift PR 2548, the toolkit's extended conformance to `Hashable`, `Equatable` and `Identifiable` for `UtilityElement` [here](https://github.com/ArcGIS/arcgis-runtime-toolkit-swift/blob/83b0d0926dabe0322c5abd274e0e264b9a7395f3/Sources/ArcGISToolkit/Extensions/UtilityElement.swift) should be reviewed and edited or removed as "it doesn't match the documented semantics of the `Equatable` protocol".

It would probably be good to check all extended `Utility*` conformances.

Types under consideration:
- [x] UtilityElement #160 
- [x] UtilityAssetType #168 
- [x] UtilityNamedTraceConfiguration #169
- [x] UtilityNetwork #180 
- [x] UtilityTraceFunctionOutput #183 
- [x] UtilityTerminal #182

PR #303, opened 2023-04-11
[UN Trace] - Rename `viewPoint` to `screenPoint`, support viewpoint animation, make proxy a non-binding property

dfeinzimer:
Related #260 

- The `viewPoint` parameter is renamed to `screenPoint` to match the new default names provided when adding the `onSingleTapGesture` to a geo view.
- The proxy is no longer a binding property. There was no need for it be a binding and it simplifies the usage.
- In the sample, the map is now simple a `@State` property rather than a `MapDataModel`
- Tapping "Zoom To" either on a completed trace name or starting point will perform an animated zoom to the proper extent.


mhdostal (commented):
On line #152 (in the `@ViewBuilder private var assetGroupDetail` variable) of UtilityNetworkTrace.swift, the viewpoint is being set in response to a tap on one of assets(?)/features(?), but is not animated.  Was this one supposed to be animated or not?

dfeinzimer:
> Was this one supposed to be animated or not?

Yes, just an oversight - fixed now!

mhdostal (approved):

PR #305, opened 2023-04-12
[Compass] Re-think API

dfeinzimer:
Given a mixture of feedback from Phil R. and Mark on #298 I propose the following Compass API:

Public initializer:
```swift
Compass(rotation: Double?, mapViewProxy: MapViewProxy)
```
Internal initializer:
```swift
Compass(heading: Double, mapViewProxy: MapViewProxy? = nil)
```

- In the public initializer the proxy isn't nullable as the majority of users will want animation. In the internal initializer, the proxy is nullable to keep the compass testable.

Thoughts appreciated from @rolson, @philium 

Additional note:

- Because heading is no longer a binding, when `shouldHide` was accessed from `.onChange(of: heading)` and the change was triggered by snap-to-north, `heading` would still have the pre-snap value causing `shouldHide` to compute incorrectly. For this reason, `shouldHide` became `func shouldHide(forHeading:)` so that `.onChange(of: heading)` could pass its new value.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Compass/Compass.swift
+++ b/Sources/ArcGISToolkit/Components/Compass/Compass.swift
@@ -14,8 +14,7 @@
 import ArcGIS
 import SwiftUI
 
-/// A `Compass` (alias North arrow) shows where north is in a `MapView` or
-/// `SceneView`.
+/// A `Compass` (alias North arrow) shows where north is in a `MapView` or `SceneView`.
 public struct Compass: View {
     /// The opacity of the compass.
     @State private var opacity: Double = .zero
@@ -28,7 +27,7 @@ public struct Compass: View {
     private var heading: Double
     
     /// The proxy to provide access to map view operations.
-    private var mapViewProxy: MapViewProxy
+    private var mapViewProxy: MapViewProxy?
     
     /// The width and height of the compass.
     private var size: CGFloat = 44
@@ -41,7 +40,7 @@ public struct Compass: View {
     ///   - mapViewProxy: The proxy to provide access to map view operations.
     init(
         heading: Double,
-        mapViewProxy: MapViewProxy
+        mapViewProxy: MapViewProxy? = nil
     ) {
         self.heading = heading
         self.mapViewProxy = mapViewProxy
@@ -66,7 +65,7 @@ public struct Compass: View {
                     }
                 }
                 .onTapGesture {
-                    Task { await mapViewProxy.setViewpointRotation(0) }
+                    Task { await mapViewProxy?.setViewpointRotation(0) }
                 }
                 .accessibilityLabel("Compass, heading \(Int(heading.rounded())) degrees \(CompassDirection(heading).rawValue)")
         }
@@ -88,8 +87,7 @@ public extension Compass {
     /// a direction toward true North, 90° indicates a direction toward true
     /// West, etc.).
     /// - Parameters:
-    ///   - viewpointRotation: The viewpoint rotation whose value determines the
-    ///   heading of the compass.
+    ///   - viewpointRotation: The viewpoint rotation whose value determines the heading of the compass.
     ///   - mapViewProxy: The proxy to provide access to map view operations.
     init(
         viewpointRotation: Double?,
--- a/Sources/ArcGISToolkit/Components/Compass/Compass.swift
+++ b/Sources/ArcGISToolkit/Components/Compass/Compass.swift
@@ -82,8 +82,8 @@ extension Compass {
 }
 
 public extension Compass {
-    /// Creates a compass with a rotation (0° indicates a direction toward true North, 90° indicates
-    /// a direction toward true West, etc.).
+    /// Creates a compass with a rotation (0° indicates a map rotation towards true North, 90°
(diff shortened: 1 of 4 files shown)


Sources/ArcGISToolkit/Components/Compass/Compass.swift
```diff
@@ -14,42 +14,35 @@
 import ArcGIS
 import SwiftUI
 
-/// A `Compass` (alias North arrow) shows where north is in a `MapView` or
-/// `SceneView`.
+/// A `Compass` (alias North arrow) shows where north is in a `MapView` or `SceneView`.
```
mhdostal, line 17:
```suggestion
/// A `Compass` (alias North arrow) shows where north is in a `MapView`.
```

mhdostal (changes requested):

philium:
I don't think we need or should have `Compass.init(viewpoint:mapViewProxy:)`. The only thing the viewpoint is used for now is to get the value of the rotation property. In other words, a call to `viewpoint.rotation` is the only different between `Compass.init(viewpoint:mapViewProxy:)` and `Compass.init(viewpointRotation:mapViewProxy:)`. Therefore `Compass.init(viewpoint:mapViewProxy:)` isn't really pulling its own weight and I would suggest that we remove it.

Sources/ArcGISToolkit/Components/Compass/Compass.swift
```diff
@@ -63,65 +56,65 @@ public struct Compass: View {
...
-                scale: oldViewpoint.targetScale,
-                rotation: newViewpointRotation
+        if let viewpointRotation {
+            self.init(
+                heading: viewpointRotation.isZero ? .zero : 360 - viewpointRotation,
+                mapViewProxy: mapViewProxy
             )
+        } else {
+            self.init(heading: .nan, mapViewProxy: mapViewProxy)
         }
```
philium, line 117:
```suggestion
        let heading: Double
        if let viewpointRotation {
            heading = viewpointRotation.isZero ? .zero : 360 - viewpointRotation
        } else {
            heading = .nan
        }
        self.init(heading: heading, mapViewProxy: mapViewProxy)
```

Sources/ArcGISToolkit/Components/Compass/Compass.swift
```diff
@@ -63,65 +55,49 @@ public struct Compass: View {
...
+    func shouldHide(forHeading heading: Double) -> Bool {
+        (heading.isZero || heading.isNaN) && autoHide
     }
-    
-    /// Creates a compass with a binding to an optional viewpoint.
+}
+
+public extension Compass {
+    /// Creates a compass with a rotation (0° indicates a direction toward true North, 90° indicates
+    /// a direction toward true West, etc.).
```
dfeinzimer, line 86: Is it me or is our doc confusing on this? Yes at a rotation of 90° the needle points west but the map is oriented to east.

mhdostal (approved):

Sources/ArcGISToolkit/Components/Compass/Compass.swift
```diff
@@ -14,42 +14,34 @@
...
     private var size: CGFloat = 44
     
-    /// The heading of the compass in degrees.
-    @Binding private var heading: Double
-    
-    /// Creates a compass with a binding to a heading based on compass
-    /// directions (0° indicates a direction toward true North, 90° indicates a
-    /// direction toward true East, etc.).
+    /// Creates a compass with a heading based on compass directions (0° indicates a direction
+    /// toward true North, 90° indicates a direction toward true East, etc.).
```
dfeinzimer, line 36: Same question with this heading doc. Yes at a heading of 90° the needle is pointing eastward but the map would be oriented to the west.

Should this doc be written from the point of view of where the needle is pointing or how the map is oriented?

mhdostal: I makes sense. again, this is talking about direction of rotation, which is opposite for `heading` than `rotation`.

Sources/ArcGISToolkit/Components/Compass/Compass.swift
mhdostal: The doc talks specifically about the direction of rotation, which is correct in this regard (positive values counter-clockwise).

mhdostal (approved):

laurenwinter:
Per [my Slack comment](https://esri-runtime.slack.com/archives/C01C2C56LQH/p1682540748314009), sorry that I didn't see this PR earlier. The main problem I have with this change is the loss of SceneView usage. If you provide a public Compass(heading: Double, mapViewProxy: MapViewProxy? = nil) then I can use it with SceneView but I have no easy way to implement the tap-to-North functionality. The 2 solutions that come to mind are:
1. Bring back the binding rotation 
2. With a mapViewProxy = nil, remove your onTapGesture so that I can implement my own tap handler and function to set the SceneView to North.

I realize that the difference between the MapView and SceneView makes it difficult for you to meet all needs, and that MapView is much more common, so allowing for me to implement SceneView functionality while using your Compass is my preference (#2).

rolson:
Thanks @laurenwinter for bringing this up.  We will add back in the support you need.

PR #320, opened 2023-04-21
Remove `LoginViewModel.signInButtonEnabled`

philium:
It isn't being used. If it ends up being needed in the future, it could just be a computed property (since each of the values it depends on are published) on the view (since the model shouldn't really know about things like buttons).

### DIFF
--- a/Sources/ArcGISToolkit/Components/Authentication/Login.swift
+++ b/Sources/ArcGISToolkit/Components/Authentication/Login.swift
@@ -28,17 +28,10 @@ struct LoginCredential: Hashable {
 @MainActor
 final class LoginViewModel: ObservableObject {
     /// The username.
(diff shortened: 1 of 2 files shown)


mhdostal:
@njarecha Was there something planned for the `signInButtonEnabled` property?

dfeinzimer:
> Was there something planned for the signInButtonEnabled property?

@mhdostal I believe that functionality moved into [CredentialInputView here](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/231e74058b503e1379b1753ccf1825f7fb7a60e8/Sources/ArcGISToolkit/Utility/CredentialInputView.swift#L76-L83) and this was just never removed.

mhdostal (approved):

PR #321, opened 2023-04-24
[Authenticator] Replace `task` with `onAppear`

dfeinzimer:
While recently reviewing nearby code @philium suggested this should use `onAppear` as no asynchronous work is performed within.

This change can be tested on branch "ryan2" by opening the regular toolkit sample app, GeoView -> Authentication -> Multiple token secured resources


rolson:
does that still work? The timing of the task vs onAppear is different. As long as it still works

dfeinzimer:
> does that still work? The timing of the task vs onAppear is different. As long as it still works

I tried the change on your `ryan2` branch with the "Multiple token secured resources" button and it still worked as expected.

mhdostal (approved):

PR #322, opened 2023-04-25
[Authenticator] Update language in the trust host modifier

dfeinzimer:
While recently reviewing nearby code @philium suggested we might consider changing the language of the trust host modifier:

> "Dangerous: Allow Connection" is a really odd name for a button. It seems like the button should just be named "Allow" and the part about it being dangerous should be in the alert's message.

"Continue" is another possibility for the button that I would consider.

Screenshot 2023-04-24 at 17 08 21


mhdostal (approved):

mhdostal:
cc: @njarecha

PR #327, opened 2023-05-05
SceneView support for Compass

### LINKED ISSUE #326: Support Compass use with SceneView
This functionality was inadvisedly removed from the v100.1 release. We need to add it back in to allow clients to use the Compass component with a `SceneView`.

mhdostal:
Closes #326 

This PR adds back SceneView support for Compass. Because there is the ability to use `Viewpoint`s or the the different CameraController types to control  SceneView, a custom _action_ is provided via a view modifier. This allows clients to do whatever is best for their implementation of SceneView instead of relying on a default implementation in the Compass.  ~The `mapViewProxy` argument for the initializer is now optional (and defaults to `nil).~

In order to prevent the confusion with behavior if the proxy is nil, there are now two initializers:

```swift
    init(rotation: Double?, mapViewProxy: MapViewProxy)
    init(rotation: Double?, action: @escaping () -> Void)
```

 The usage in a SceneView would be as follows (using the 2nd initializer, the one that takes an `action`):

```swift
@State private var scene = Scene(basemapStyle: .arcGISImagery)

/// Allows for communication between the Compass and MapView or SceneView.
@State private var viewpoint: Viewpoint?

var body: some View {
    SceneViewReader { proxy in
        SceneView(scene: scene)
            .onViewpointChanged(kind: .centerAndScale) { viewpoint = $0 }
            .overlay(alignment: .topTrailing) {
                Compass(rotation: viewpoint?.rotation)
                    .action {
                        if let viewpoint {
                            Task {
                                await proxy.setViewpoint(viewpoint.withRotation(.zero))
                            }
                        }
                    }
                    .padding()
            }
    }
}
```

@laurenwinter Please give this a look and let me know if it works for you.

### DIFF
--- a/Documentation/Compass/README.md
+++ b/Documentation/Compass/README.md
@@ -23,20 +23,23 @@ Compass:
     /// a direction toward true West, etc.).
     /// - Parameters:
     ///   - rotation: The rotation whose value determines the heading of the compass.
-    ///   - mapViewProxy: The proxy to provide access to map view operations.
-    public init(rotation: Double?, mapViewProxy: MapViewProxy)
+    ///   - mapViewProxy: The proxy to provide access to map view operations. If `mapViewProxy`
+    ///   is non-`nil`, the proxy will be used to set the rotation to `.zero` when the compass is
+    ///   tapped on. Otherwise, the `action` will be called (set using the `action` view modifier).
+    public init(rotation: Double?, mapViewProxy: MapViewProxy? = nil)
 ```
 
 `Compass` has the following modifiers:
 
 - `func compassSize(size: CGFloat)` - The size of the `Compass`, specifying both the width and height of the compass.
 - `func autoHideDisabled(_:)` - Specifies whether the ``Compass`` should automatically hide when the heading is 0.
+- `action(perform action: @escaping () -> Void)` - An action to perform when the compass is tapped. If `mapViewProxy` is non-`nil`, then this will have no effect.
 
 ## Behavior:
 
-Whenever the map is not orientated North (non-zero bearing) the compass appears. When reset to north, it disappears. The `automaticallyHides` view modifier allows you to disable the auto-hide feature so that it is always displayed.
+Whenever the map is not orientated North (non-zero bearing) the compass appears. When reset to north, it disappears. The `autoHideDisabled` view modifier allows you to disable the auto-hide feature so that it is always displayed.
 
-When the compass is tapped, the map orients back to north (zero bearing).
+When the compass is tapped, the map orients back to north (zero bearing). 
 
 ## Usage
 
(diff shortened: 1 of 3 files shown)


Documentation/Compass/README.md
```diff
@@ -23,20 +23,23 @@ Compass:
     /// a direction toward true West, etc.).
     /// - Parameters:
     ///   - rotation: The rotation whose value determines the heading of the compass.
-    ///   - mapViewProxy: The proxy to provide access to map view operations.
-    public init(rotation: Double?, mapViewProxy: MapViewProxy)
+    ///   - mapViewProxy: The proxy to provide access to map view operations. If `mapViewProxy`
+    ///   is non-`nil`, the proxy will be used to set the rotation to `.zero` when the compass is
+    ///   tapped on. Otherwise, the `action` will be called (set using the `action` view modifier).
+    public init(rotation: Double?, mapViewProxy: MapViewProxy? = nil)
```
rolson, line 29: This change makes the init dual purpose which I find confusing.

I would rather have 2 inits that had 2 different behaviors.  One you provide a proxy, the other you provide an action. 

WIth this you don't have undefined or non-working behavior available if I pass `nil` proxy but don't provide an action:

```swift
// No need to set an action - everything is done via provided proxy
public init(rotation: Double?, mapViewProxy: MapViewProxy)

// You must provide an action
public init(rotation: Double?, action: @escaping () -> Void)
```

rolson (changes requested):

Documentation/Compass/README.md
mhdostal: Yea, I like that better. Thanks.

rolson (dismissed):

Documentation/Compass/README.md
```diff
@@ -34,9 +41,9 @@ Compass:
 
 ## Behavior:
 
-Whenever the map is not orientated North (non-zero bearing) the compass appears. When reset to north, it disappears. The `automaticallyHides` view modifier allows you to disable the auto-hide feature so that it is always displayed.
+Whenever the map is not orientated North (non-zero bearing) the compass appears. When reset to north, it disappears. The `autoHideDisabled` view modifier allows you to disable the auto-hide feature so that it is always displayed.
 
-When the compass is tapped, the map orients back to north (zero bearing).
+When the compass is tapped, the map orients back to north (zero bearing).
```
dfeinzimer, line 46:
```suggestion
When the compass is tapped, the map orients back to north (zero bearing).
```

Sources/ArcGISToolkit/Components/Compass/Compass.swift
```diff
@@ -32,17 +32,30 @@ public struct Compass: View {
     /// The width and height of the compass.
     private var size: CGFloat = 44
     
+    /// An action to perform when the compass is tapped. Note if `mapViewProxy` is non-`nil`
+    /// this will not be invoked.
+    private var action: (() -> Void)?
+
```
dfeinzimer, line 38:
```suggestion
    
```

Sources/ArcGISToolkit/Components/Compass/Compass.swift
```diff
@@ -32,17 +32,30 @@ public struct Compass: View {
     /// The width and height of the compass.
     private var size: CGFloat = 44
     
+    /// An action to perform when the compass is tapped. Note if `mapViewProxy` is non-`nil`
+    /// this will not be invoked.
```
dfeinzimer, line 36: Using the [Note callout](https://developer.apple.com/library/archive/documentation/Xcode/Reference/xcode_markup_formatting_ref/Note.html#//apple_ref/doc/uid/TP40016497-CH39-SW1) makes it a little more noticeable in my opinion. Not a big deal either way, take or leave.

```suggestion
    /// An action to perform when the compass is tapped.
    ///
    /// - Note: If `mapViewProxy` is non-`nil` this will not be invoked.
```

mhdostal: I removed that sentence completely, as the new initializers will prevent that scenario.

dfeinzimer (commented):
Looks good. Just some build errors in the tests now.

mhdostal:
> Just some build errors in the tests now.

Thanks for catching that! All fixed now.

dfeinzimer (approved):

dfeinzimer:
@mhdostal should this be cherry-picked into `v.next` as well?

mhdostal:
> @mhdostal should this be cherry-picked into `v.next` as well?

Yes. It should have gone into `v.next` the first time instead of main.

Issue #247, opened 2023-01-13
Support localization


njarecha:
This [PR](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/243) adds support for localization for toolkit. However, there are more strings needs to moved in the file and get translated later on.

dfeinzimer:
Useful initializers:
 - [`String(localized:comment:)`](https://developer.apple.com/documentation/swift/string/init(localized:table:bundle:locale:comment:))
- [`Text(_:format:)`](https://developer.apple.com/documentation/swiftui/text/init(_:format:))
- [`Int.formatted(_:)`](https://developer.apple.com/documentation/swift/int/formatted(_:)-36hqc)

PR #330, opened 2023-05-11
Improve Toolkit localization

dfeinzimer:
Related #247

### DIFF
--- a/Examples/ExamplesApp/Info.plist
+++ b/Examples/ExamplesApp/Info.plist
@@ -12,8 +12,40 @@
 	<string>6.0</string>
 	<key>CFBundleLocalizations</key>
 	<array>
+		<string>ar</string>
+		<string>ca</string>
+		<string>cs</string>
+		<string>da</string>
+		<string>de</string>
+		<string>el</string>
 		<string>en</string>
 		<string>es</string>
+		<string>fi</string>
+		<string>fr</string>
+		<string>he</string>
+		<string>hr</string>
+		<string>hu</string>
+		<string>id</string>
+		<string>it</string>
+		<string>ja</string>
+		<string>ko</string>
+		<string>nb</string>
+		<string>nl</string>
+		<string>pl</string>
+		<string>pt-BR</string>
+		<string>pt-PT</string>
+		<string>ro</string>
+		<string>ru</string>
+		<string>sk</string>
+		<string>sv</string>
+		<string>th</string>
+		<string>tr</string>
+		<string>uk</string>
+		<string>vi</string>
+		<string>zh-Hans-CN</string>
+		<string>zh-Hans</string>
+		<string>zh-Hant-HK</string>
+		<string>zh-Hant-TW</string>
 	</array>
 	<key>CFBundleName</key>
 	<string>$(PRODUCT_NAME)</string>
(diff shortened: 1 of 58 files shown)


Examples/ExamplesApp/Info.plist
```diff
@@ -10,6 +10,43 @@
 	<string>$(PRODUCT_BUNDLE_IDENTIFIER)</string>
 	<key>CFBundleInfoDictionaryVersion</key>
 	<string>6.0</string>
+	<key>CFBundleLocalizations</key>
```
dfeinzimer, line 13: These values referenced from the Cocoa supported languages.

Even though the Examples app itself won't be localized, adding these allows us to test component localization in the app.

Sources/ArcGISToolkit/Components/Authentication/TrustHostViewModifier.swift
```diff
@@ -38,15 +38,34 @@ struct TrustHostViewModifier: ViewModifier {
...
+                Button(role: .cancel) {
                     challenge.resume(with: .cancel)
+                } label: {
+                    Text("Cancel", bundle: .module)
                 }
             } message: { _ in
-                Text("Dangerous: The certificate provided by '\(challenge.host)' is not signed by a trusted authority.")
+                Text(
+                    "Dangerous: The certificate provided by '\(challenge.host)' is not signed by a trusted authority.",
+                    bundle: .module
```
mhdostal, line 67: Because there's a variable that may not be obvious, maybe a comment:
```suggestion
                    bundle: .module,
                    comment: "A warning that the host service (challenge.host) is providing a potentially unsafe certificate."
```
Or something like that.

Sources/ArcGISToolkit/Components/BasemapGallery/BasemapGallery.swift
```diff
@@ -214,8 +214,11 @@ extension AlertItem {
     /// - Parameter loadBasemapError: The load basemap error.
     init(loadBasemapError: Error) {
         self.init(
-            title: "Error loading basemap.",
-            message: "\((loadBasemapError as? ArcGISError)?.details ?? "The basemap failed to load for an unknown reason.")"
+            title: String(localized: "Error loading basemap.", bundle: .module),
+            message: String(
+                localized: "\((loadBasemapError as? ArcGISError)?.details ?? "The basemap failed to load for an unknown reason.")",
```
mhdostal, line 219: Error strings from the SDK/Core won't be localized, so I don't think this will work. Probably need a local string variable and just localized the 2nd bit if it gets there.

Sources/ArcGISToolkit/Components/Popups/PopupMedia/Images/AsyncImageView.swift
```diff
@@ -67,7 +73,7 @@ struct AsyncImageView: View {
                     Image(systemName: "exclamationmark.circle")
                         .aspectRatio(contentMode: .fit)
                         .foregroundColor(.red)
-                    Text("An error occurred loading the image: \(error.localizedDescription).")
+                    Text("\(loadingErrorMessage): \(error.localizedDescription).")
```
mhdostal, line 76: Is there a reason this was broken out, instead of just localizing the whole thing? The placeholder (error.localizedDescription), might be in a different place for some languages.

mhdostal (changes requested):

Sources/ArcGISToolkit/Components/Popups/PopupMedia/Images/AsyncImageView.swift
dfeinzimer: > The placeholder (error.localizedDescription), might be in a different place for some languages.

That's a good point, I've re-included the variable portion.

mhdostal (approved):

PR #332, opened 2023-05-16
[Scalebar] Localization

### LINKED ISSUE #247: Support localization

dfeinzimer:
Related #247 

Localizes the scalebar

Preview of an Arabic localization (right-to-left language sample):
Screenshot 2023-05-16 at 14 28 22


Sources/ArcGISToolkit/Components/Scalebar/ScalebarViewModel.swift
```diff
@@ -56,8 +56,10 @@ final class ScalebarViewModel: ObservableObject {
         let numberString = numberFormatter.string(
             from: NSNumber(value: altMapLength)
         ) ?? ""
-        let bottomUnitsText = " \(altDisplayUnits.abbreviation)"
-        let label = "\(numberString)\(bottomUnitsText)"
+        let label = String.totalLengthLabel(
+            length: numberString,
+            unitLabel: altDisplayUnits.localizedAbbreviation
+        )
```
philium, line 62: Have you considered using a [formatted measurement](https://developer.apple.com/documentation/foundation/measurement#3820415) as the label? That would simplify this code and might reduce the number of localized strings needed in the toolkit.

rolson: That's what I was thinking as well. I thought I recalled we had a way to go from Esri Linear Units to Measurements and localizing Measurements could be done by iOS.

philium: We do. They're not in the documentation as DocC doesn't currently support documenting members of types in a different module, but we have `class UnitAngle.fromAngularUnit(_:)`, `class UnitArea.fromAreaUnit(_:)`, and `class UnitLength.fromLinearUnit(_:)` for this very purpose.

dfeinzimer: Implement - its a much cleaner approach, thank you both.

rolson: 🥇

philium: There's a reason I referred you to the `formatted` methods on `Measurement` itself: "When working in Swift, use `formatted` methods directly on the types you want to format" (see [Data Formatting](https://developer.apple.com/documentation/foundation/data_formatting)). That's what we should do. For starters, add a convenience initializer to `Measurement`:

```swift
extension Measurement where UnitType == UnitLength {
    init(value: Double, linearUnit: LinearUnit) {
        self.init(value: value, unit: .fromLinearUnit(linearUnit))
    }
}
```

Then instead of having a method on `String` to format the values, just add a reusable format style:

```swift
extension FormatStyle where Self == Measurement<UnitLength>.FormatStyle {
    static var scaleMeasurement: Self {
        .measurement(width: .abbreviated, usage: .asProvided)
    }
}
```

FYI: That `measurement` method takes an optional `numberFormatStyle` that can be used to customize the format of the number part of the measurement.

Then, use those at the point of use:

```swift
// In 'ScalebarViewModel.getter:alternateUnit'

let measurement = Measurement(value: altMapLength, linearUnit: altDisplayUnits)
let label = measurement.formatted(.scaleMeasurement)

// In 'ScalebarViewModel.updateLabels()'

let segmentText: String
if index == numSegments - 1, let displayUnit {
    let measurement = Measurement(
        value: segmentMapLength,
        linearUnit: displayUnit
    )
    segmentText = measurement.formatted(.scaleMeasurement)
} else {
    segmentText = segmentMapLength.formatted(.number)
}
```

dfeinzimer: Yes thank you. I processed the feedback from your second comment but then forgot to process your first. Thank you for the suggestions. Done now.

Sources/ArcGISToolkit/Components/Popups/PopupView.swift
```diff
@@ -65,7 +65,14 @@ public struct PopupView: View {
                     case .success(_):
                         PopupElementScrollView(popupElements: popup.evaluatedElements)
                     case .failure(let error):
-                        Text("Popup evaluation failed: \(error.localizedDescription)", bundle: .module)
+                        Text(
+                            "Popup evaluation failed: \(error.localizedDescription)",
+                            bundle: .module,
+                            comment: """
+                                     An error message shown when a popup cannot be shown. The
```
mhdostal, line 72: Just to avoid "shown"-"shown"...
```suggestion
                                     An error message shown when a popup cannot be displayed. The
```

mhdostal (changes requested):

mhdostal:
If this is just for the scalar, why are there so many `Sources/ArcGISToolkit/Resources/en.lproj/Localizable.strings` strings?

dfeinzimer:
> If this is just for the scalar, why are there so many `Sources/ArcGISToolkit/Resources/en.lproj/Localizable.strings` strings?

Those changes actually came from #334, which was already reviewed and merged into this branch. Sorry for the confusion, I probably should have left that PR in draft until this was merged.

Sources/ArcGISToolkit/Components/Popups/PopupView.swift
mhdostal: @dfeinzimer You OK with this change?

mhdostal (approved):

dfeinzimer:
Resolved 1 merge conflict, ready for review again

mhdostal (approved):

PR #334, opened 2023-05-18
Add starter english localizable strings

### LINKED ISSUE #333: Add initial set of English localizable strings
Related #247 

Given `Product` -> `Export Localizations` is [currently broken](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/243#discussion_r1084658404) for Swift Packages, @njarecha has pointed out that we should manually add the current set of localizable strings to `en.lproj/Localizable.strings` so that translators can use this file to generate localizations for other languages in the mean time.

Given that we're not sure when Apple will address the bug, it makes sense to do this now as apart of the localization effort.

dfeinzimer:
This branch is based off of and to be merged following #332 

Closes #333 
Related #247 

Adds starter localization strings. See #333 for context.


njarecha (approved):
Looks good

mhdostal (approved):

PR #338, opened 2023-05-19
Update basemap gallery code snippet

zkline101:
A map property can no longer use the `@StateObject` property wrapper b/c `Map` no longer conforms to `ObservableObject`.

This PR can go into `main` because it applies to the current toolkit code.

### DIFF
--- a/Documentation/BasemapGallery/README.md
+++ b/Documentation/BasemapGallery/README.md
@@ -77,7 +77,7 @@ Selecting a basemap with a spatial reference that does not match that of the geo
 ### Basic usage for displaying a `BasemapGallery`.
 
 ```swift
-@StateObject var map = Map(basemapStyle: .arcGISImagery)
+@State var map = Map(basemapStyle: .arcGISImagery)
 
 var body: some View {
     MapView(map: map)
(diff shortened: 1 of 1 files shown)


Documentation/BasemapGallery/README.md
```diff
@@ -77,7 +77,7 @@ Selecting a basemap with a spatial reference that does not match that of the geo
 ### Basic usage for displaying a `BasemapGallery`.
 
 ```swift
-@StateObject var map = Map(basemapStyle: .arcGISImagery)
+@State var map = Map(basemapStyle: .arcGISImagery)
```
zkline101, line 80:
```suggestion
@State private var map = Map(basemapStyle: .arcGISImagery)
```

dfeinzimer (commented):
Changes look good. I'm not sure about merging into `main` though. This would be the first time I think I've seen a PR merge into `main` outside of a release. Perhaps we should wait for @mhdostal's word on it?

mhdostal:
I'm OK this going into main, as it's only for the repo doc and not code. That way anyone looking at the repo will get the correct doc, without having to switch to `v.next`. Otherwise, yes, `v.next` is the preferred branch mid-release.

dfeinzimer (approved):

PR #345, opened 2023-06-01
Use ellipsis character in localized strings

philium:
The ellipsis character (…) should be used instead of three dots (...) in localized strings.


mhdostal (approved):

mhdostal:
What's the key command to type the `ellipsis` character?

dfeinzimer (approved):
Good catch, thank you Phil.

PR #353, opened 2023-06-09
Add custom module property

philium:
This property first looks for a bundle with the toolkit's identifier. This prevents a crash on accessing a resource when the toolkit has been built into a binary framework.


dfeinzimer (approved):
Thanks @philium, did you build the toolkit into a binary framework and experience a crash?

philium:
Yes.

zkline101 (approved):

PR #356, opened 2023-06-15
Fix toolkit localization export

dfeinzimer:
@njarecha discovered that using `NSLocalizedString` here was breaking `xcodebuild -exportLocalizations` and causing that string to be the only one exported.

Screenshot 2023-06-15 at 15 34 34

### DIFF
--- a/Sources/ArcGISToolkit/Components/Popups/PopupMedia/Images/AsyncImageViewModel.swift
+++ b/Sources/ArcGISToolkit/Components/Popups/PopupMedia/Images/AsyncImageViewModel.swift
@@ -110,7 +110,7 @@ extension LoadImageError: LocalizedError {
         return String(
             localized: "The URL could not be reached or did not contain image data",
             bundle: .toolkitModule,
-            comment: "No Data"
+            comment: "Description of error thrown when a remote image could not be loaded."
         )
     }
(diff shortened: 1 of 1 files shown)


njarecha (dismissed):

Sources/ArcGISToolkit/Components/Popups/PopupMedia/Images/AsyncImageViewModel.swift
```diff
@@ -107,8 +107,8 @@ struct LoadImageError: Error {
 
 extension LoadImageError: LocalizedError {
     public var errorDescription: String? {
-        return NSLocalizedString(
-            "The URL could not be reached or did not contain image data",
+        return String(
+            localized: "The URL could not be reached or did not contain image data",
             bundle: .toolkitModule,
             comment: "No Data"
```
philium, line 113: I recently (re)watched [Build global apps: Localization by example](https://developer.apple.com/wwdc22/10110). In it, the presenter explained that a good localized string comment has three components:

image

I would suggest expanding this comment to include both what the string is used for and the context. Perhaps something like "Description of error thrown when a remote image could not be loaded."

mhdostal (approved):

PR #357, opened 2023-06-17
Make `content` parameter view builder

philium:
Doing so simplifies caller code in some cases. Also, remove the `escaping` attribute since the closure doesn't actually escape.

### DIFF
--- a/Documentation/FloatingPanel/README.md
+++ b/Documentation/FloatingPanel/README.md
@@ -65,7 +65,7 @@ The floating panel is displayed via a view modifier that allows you to set the c
 MapView(
     map: map
 )
-.floatingPanel() {
+.floatingPanel {
     List(1..<21) { Text("\($0)") }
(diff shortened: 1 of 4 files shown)


dfeinzimer (approved):
In support of the new Forms component I've actually been evaluating how to make FP initialize its content lazily (only upon presentation) to bring its behavior closer to a native sheet so that a user doesn't have to conditionally pass content.

I think this'll involve relocating the closure's [current call site](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/77e79160766dea68a21e3e7447dadfb95a146167/Sources/ArcGISToolkit/Components/FloatingPanel/FloatingPanelModifier.swift#L54) down into the FP's body, at which point I think it technically _will_ (?) become escaping.

It probably makes sense to make this change in the mean time.

mhdostal (approved):

PR #359, opened 2023-06-20
Update English strings

dfeinzimer:
Updates the English localized strings file to the generated one that has been provided to the localization team.

### DIFF
--- a/Sources/ArcGISToolkit/Resources/en.lproj/Localizable.strings
+++ b/Sources/ArcGISToolkit/Resources/en.lproj/Localizable.strings
@@ -1,8 +1,20 @@
+/* No comment provided by engineer. */
+"(%lld)" = "(%lld)";
+
+/* No comment provided by engineer. */
+"%@" = "%@";
+
+/* No comment provided by engineer. */
+"%@ %@" = "%@ %@";
+
+/* No comment provided by engineer. */
+"%lf" = "%lf";
+
 /* A label declaring the number of starting points selected for a utility network trace. */
 "%lld selected" = "%lld selected";
 
 /* An alert message indicating that a certificate is required to access
- content on a remote host. The variable is the host that prompted the challenge. */
+content on a remote host. The variable is the host that prompted the challenge. */
 "A certificate is required to access content on %@." = "A certificate is required to access content on %@.";
 
 /* A button to add new utility trace starting points. */
@@ -14,6 +26,9 @@
 /* A button allowing users to view a list of all sites defined in a floor aware map. */
 "All sites" = "All sites";
 
+/* No comment provided by engineer. */
+"All Sites" = "All Sites";
+
 /* A message describing the outcome of clearing all utility network trace results. */
 "All the trace inputs and results will be lost." = "All the trace inputs and results will be lost.";
 
@@ -21,13 +36,13 @@
 "Allow" = "Allow";
 
 /* No comment provided by engineer. */
-"Attributes" = "Attributes";
-
-/* No comment provided by engineer. */
 "An error occurred loading the image: %@." = "An error occurred loading the image: %@.";
 
 /* No comment provided by engineer. */
 "Attachments" = "Attachments";
+
+/* No comment provided by engineer. */
+"Attributes" = "Attributes";
 
 /* No comment provided by engineer. */
 "Authentication Required" = "Authentication Required";
@@ -60,9 +75,9 @@
 "Color" = "Color";
 
 /* An compass description to be read by a screen reader describing the
- current heading. The first variable being a degree value and the
- second being a corresponding cardinal direction (north, northeast,
- east, etc.). */
+current heading. The first variable being a degree value and the
+second being a corresponding cardinal direction (north, northeast,
+east, etc.). */
 "Compass, heading %lld degrees %@" = "Compass, heading %lld degrees %@";
 
 /* No comment provided by engineer. */
@@ -77,7 +92,7 @@
 /* No comment provided by engineer. */
 "Delete" = "Delete";
 
-/* No comment provided by engineer. */
+/* A button to close the bookmark selection menu. */
 "Done" = "Done";
 
 /* No comment provided by engineer. */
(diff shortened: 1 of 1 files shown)


mhdostal (dismissed):

Sources/ArcGISToolkit/Resources/en.lproj/Localizable.strings
yo1995, line 1: image

For some reasons, many textual files that come from [other team](https://github.com/Esri/arcgis-maps-sdk-swift/pull/4#discussion_r1151224324) have 755 file permissions. In addition to us fixing them, how can we let them know this is an issue? 🤔

Sources/ArcGISToolkit/Resources/en.lproj/Localizable.strings
```diff
@@ -1,8 +1,20 @@
...
+"(%lld)" = "(%lld)";
+
+/* No comment provided by engineer. */
+"%@" = "%@";
+
+/* No comment provided by engineer. */
+"%@ %@" = "%@ %@";
+
+/* No comment provided by engineer. */
+"%lf" = "%lf";
```
njarecha, line 11: Where is these used? Shouldn't there some text with these parameters?

philium: These are do to how certain `Text` views are created. A search for `"\(` finds (almost?) all of them. For example, "%@" is the result of `Text` being initialized like so: `Text("\(...)")`. That is done both in `AttachmentList` and `AttachmentPreview` to display the attachment size:

```swift
Text("\(attachmentModel.attachment.size.formatted(.byteCount(style: .file)))")
```

Instead, `Text.init(_:format:)` should be used instead:

```swift
Text(Int64(attachmentModel.attachment.size), format: .byteCount(style: .file))
```

If a string used to create a `Text` view should not be localized, use `Text.init(verbatim:)`.

Sources/ArcGISToolkit/Resources/en.lproj/Localizable.strings
dfeinzimer: This is a file @njarecha created and was then uploaded/retrieved from the SMB share. Does it create some issue?

Sources/ArcGISToolkit/Resources/en.lproj/Localizable.strings
njarecha: @dfeinzimer I just saw your commit. Here, nothing to localize. The parameters are returned as it is so you should use `Text.init(verbatim:)` for these cases.

dfeinzimer: @njarecha I've gotten it down to two. I'm not sure where they may be coming from.

philium: One of the remaining ones is coming from this:

```swift
Text((item.result as? Double).map { "\($0)" } ?? "N/A")
```

It probably would be best to use an if statement there:

```swift
if let result = item.result as? Double {
    Text(result, format: ...)
} else {
    Text("N/A")
}
```

The other one must be coming from here somehow:

```swift
Text(
    "\(viewModel.pendingTrace.startingPoints.count, specifier: "%lld") selected",
    bundle: .toolkitModule,
    comment: "A label declaring the number of starting points selected for a utility network trace."
)
```

For that, the specifier shouldn't be necessary. See one of Apple's recent WWDC sessions on localization for how to localize plural strings.

dfeinzimer: Thanks Phil, that took care of the first one. Removing the the specifier didn't seem to make any difference for the second. I poked around some more but haven't been able to find the source.

philium: Found it. The last one is coming from the title of a disclosure group (used in two places:

```swift
DisclosureGroup("(\(viewModel.selectedTrace?.elementResults.count ?? 0))", ...)
...
DisclosureGroup("(\(viewModel.selectedTrace?.utilityFunctionTraceResult?.functionOutputs.count ?? 0))", ...)
```

The solution is to use `DisclosureGroup.init(isExapnded:content:label:)` and in the label closure, return an instance of `Text` using `Text(_:format:)`.

yo1995: 💰 like a treasure hunt!

dfeinzimer: Yes, that was it, thank you again Phil!

Sources/ArcGISToolkit/Resources/en.lproj/Localizable.strings
```diff
@@ -128,6 +134,9 @@
 /* No comment provided by engineer. */
 "Mode" = "Mode";
 
+/* Shorthand for Not Available */
+"N/A" = "N/A";
```
njarecha, line 138: Should we use `Not Available` here? Not sure if there is an easy way to localize `N/A` to other languages.

philium: That seems like a good idea considering it has [various meanings](https://en.wikipedia.org/wiki/N/A):

> N/A or sometimes n/a is a common [abbreviation](https://en.wikipedia.org/wiki/Abbreviation) in [tables](https://en.wikipedia.org/wiki/Table_(information)) and lists for the phrase not applicable,[[1]](https://en.wikipedia.org/wiki/N/a#cite_note-ODA-1)[[2]](https://en.wikipedia.org/wiki/N/a#cite_note-Foley-2)[[3]](https://en.wikipedia.org/wiki/N/a#cite_note-Kay-3)[[4]](https://en.wikipedia.org/wiki/N/a#cite_note-Elia-4) not available,[[3]](https://en.wikipedia.org/wiki/N/a#cite_note-Kay-3)[[4]](https://en.wikipedia.org/wiki/N/a#cite_note-Elia-4) not assessed, [[5]](https://en.wikipedia.org/wiki/N/a#cite_note-5) or no answer.

njarecha (approved):
Thank you for making all the changes!

mhdostal (approved):

dfeinzimer:
Thanks all! @njarecha I've also updated the localization team's copy on the network share export folder.

njarecha:
@dfeinzimer Thank you!

PR #360, opened 2023-06-21
[Form] Initial implementation

dfeinzimer:
Adds the initial Form component implementation to a Form staging branch.


Sources/ArcGISToolkit/Components/Form/FormInputs/MultiLineTextEntry.swift
```diff
@@ -0,0 +1,48 @@
...
+    @State private var text: String
+    
+    /// A `TextAreaFeatureFormInput` which acts as a configuration.
+    let input: TextAreaFeatureFormInput
+    
+    /// Creates a view for text entry spanning multiple lines.
+    /// - Parameters:
+    ///   - text: The current text value.
+    ///   - input: A `TextAreaFeatureFormInput` which acts as a configuration.
+    init(text: String?, input: TextAreaFeatureFormInput) {
```
mhdostal, line 29: At some point after we consume the Core API I'd like to see this simplified to take only a `FieldFormElement`. The element will have the input type and a method to get (and hopefully set) the value, which makes will make coding the various input types much easier.

mhdostal (changes requested):

Sources/ArcGISToolkit/Components/Form/FormElementHeader.swift
```diff
@@ -0,0 +1,29 @@
...
+struct FormElementHeader: View {
+    /// The form element the header is for.
+    let element: FieldFeatureFormElement
+    
+    var body: some View {
+        Text(element.label)
+            .font(.headline)
+        Text(element.description)
+            .font(.footnote)
+            .foregroundColor(.secondary)
```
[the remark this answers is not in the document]
dfeinzimer: I've moved it down below into a new footer view but this will be further refined in the forthcoming validation PR.

Sources/ArcGISToolkit/Components/Form/Form.swift
```diff
@@ -0,0 +1,78 @@
...
+                                input: `input`
+                            )
+                        case let `input` as TextAreaFeatureFormInput:
+                            MultiLineTextEntry(
+                                text: attributes?[element.fieldName] as? String,
+                                input: `input`
+                            )
+                        default:
+                            EmptyView()
+                        }
```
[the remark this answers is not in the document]
dfeinzimer: Done, see newly added `FormView.makeElement(_:)`, `FormView.makeFieldElement(_:)`, and `FormView.makeGroupElement(_:)`.

Sources/ArcGISToolkit/Components/Form/FormView.swift
```diff
@@ -0,0 +1,109 @@
...
+    
+    public var body: some View {
+        ScrollView {
+            VStack(alignment: .leading) {
+                FormHeader(title: formDefinition?.title)
+                Divider()
+                ForEach(formDefinition?.formElements ?? [], id: \.element?.label) { container in
+                    if let element = container.element {
+                        makeElement(element)
+                    }
```
mhdostal, line 46: Right now, the elements are pretty squashed together. For now, this is OK, but during the validation UI work, we should add spacing/dividers to better separate elements.

mhdostal (approved):

Sources/ArcGISToolkit/Components/Form/FormInputs/MultiLineTextEntry.swift
```diff
@@ -0,0 +1,58 @@
...
+    let input: TextAreaFeatureFormInput
+    
+    /// Creates a view for text entry spanning multiple lines.
+    /// - Parameters:
+    ///   - element: The form element that corresponds to this text field.
+    ///   - text: The current text value.
+    ///   - input: A `TextAreaFeatureFormInput` which acts as a configuration.
+    init(element: FieldFeatureFormElement, text: String?, input: TextAreaFeatureFormInput) {
+        self.element =  element
+        self.text = text ?? ""
```
philium, line 35: FYI: State shouldn't be taken as an initializer parameter and an initializer parameter should not be used to set state. From the documentation for [`State`](https://developer.apple.com/documentation/swiftui/state):

> Like you do for a [StateObject](https://developer.apple.com/documentation/swiftui/stateobject), declare [State](https://developer.apple.com/documentation/swiftui/state) as private to prevent setting it in a memberwise initializer, which can conflict with the storage management that SwiftUI provides. Unlike a state object, **always initialize state by providing a default value in the state’s declaration** [emphasis added], as in the above examples. Use state only for storage that’s local to a view and its subviews.

dfeinzimer: This is set to be addressed in #365

dfeinzimer: Understood. I read through some relevant threads ([1](https://forums.swift.org/t/why-swiftui-state-property-can-be-initialized-inside-init-this-other-way/62772/33), [2](https://forums.swift.org/t/default-init-internal-needs-a-workaround-enough-is-enough/40931/50)) on the Swift forums to try to deepen my understanding.

In the current state of the Form component it is expected that the `MultiLineTextEntry` and `SingleLineTextEntry` views will always be initialized with the same text values. Their parent view (`FormView`) is not notified of changes made by a user to the text value and we're not updating the feature's attributes.

> If the view is re-initialized with a different value for one of those parameters corresponding to a state property, the state property won't actually be updated to reflect that new value

This is coming in our next iteration of the Form component work. We'll be persisting any edits into the feature's attributes. At this point we'll likely revise how `MultiLineTextEntry` and `SingleLineTextEntry` are initialized because then `FormView` _will_ be concerned with edits made.

PR #361, opened 2023-06-27
[Form] Add text entry length validation UI

dfeinzimer:


Sources/ArcGISToolkit/Components/Form/FormElementHeader.swift
```diff
@@ -20,7 +20,14 @@ struct FormElementHeader: View {
...
     var body: some View {
-        Text(element.label)
-            .font(.headline)
+        Group {
+            if element.required {
+                Text(verbatim: "\(element.label) *")
+            } else {
+                Text(element.label)
+            }
+        }
```
mhdostal, line 29: This seems easier to read and more consider to me.
```suggestion
        Text(verbatim: "\(element.label + (element.required ? " *" : ""))")
```

Sources/ArcGISToolkit/Components/Form/FormInputs/MultiLineTextEntry.swift
```diff
@@ -16,14 +16,26 @@ import SwiftUI
 
 /// A view for text entry spanning multiple lines.
 struct MultiLineTextEntry: View {
+    @Environment(\.formElementPadding) var padding
```
mhdostal, line 19: `padding` confused me below because it's the same name as `.padding`. I think we should specify it's `elementPadding`.
```suggestion
    @Environment(\.formElementPadding) var elementPadding
```

Sources/ArcGISToolkit/Components/Form/FormTextEntryStyle.swift
```diff
@@ -15,20 +15,18 @@ import SwiftUI
...
+struct FormTextEntryStyle: ViewModifier {
     func body(content: Content) -> some View {
         content
-            .padding(2)
-            .overlay {
-                RoundedRectangle(cornerRadius: 5)
-                    .stroke(.secondary.opacity(0.5), lineWidth: 0.5)
-            }
+            .padding(4)
+            .background(Color(uiColor: .tertiarySystemFill))
```
mhdostal, line 22: I like the tertiary color.

Sources/ArcGISToolkit/Components/Form/FormView.swift
```diff
@@ -18,6 +18,8 @@ import SwiftUI
 /// Forms allow users to edit information about GIS features.
 /// - Since: 200.2
 public struct FormView: View {
+    @Environment(\.formElementPadding) var padding
```
mhdostal, line 21:
```suggestion
    @Environment(\.formElementPadding) var elementPadding
```

Sources/ArcGISToolkit/Components/Form/TextEntryFooter.swift
```diff
@@ -0,0 +1,114 @@
...
+                case .emptyWhenRequired:
+                    requiredText
+                case .tooLong:
+                    maximumText
+                case .tooShort:
+                    minimumText
+                }
+            } else if !description.isEmpty {
+                Text(description)
+            } else if description.isEmpty && isFocused {
```
mhdostal, line 53: don't need to check `description.isEmpty` as that was just checked in the previous `if`.
```suggestion
            } else if isFocused {
```

mhdostal: The design from Field Maps has the concept of "never focused", and doesn't show the constraint text until it has had the keyboard focus. I don't see that taken into account here.

mhdostal (changes requested):
Looks really good, just a few comments.

Sources/ArcGISToolkit/Components/Form/TextEntryFooter.swift
dfeinzimer: Good point, I think latest commit should take this into account.

"Character Constraint" seems a bit ambiguous in the design and is the difference between the two places where I see "Never Focused" come into play.

Screenshot 2023-06-30 at 17 28 27
Screenshot 2023-06-30 at 17 28 38

Because min and max length are both non-optional, it seems that there is always a character constraint and so I'm now interpreting "Character Constraint" to only apply when text field has never reached it's minimum and never again after that.

mhdostal (approved):

Sources/ArcGISToolkit/Components/Form/TextEntryFooter.swift
mhdostal: Ok. Thanks.

PR #364, opened 2023-06-30
Add UI Testing App with UI Test Target + 3 starter tests

dfeinzimer:
Related #362, Apollo 47

 - Adds a UI Testing App with a UI Test Target and 3 starter tests:
   - Basemap Gallery general usage test
     - Opens the Basemap Gallery and selects two of the available basemaps. Selects a third incompatible basemap and verifies that an error message is shown.
   - Bookmarks general usage test
     - Opens the bookmark picker and verifies the expected bookmarks are present and that `Bookmarks.onSelectionChanged(perform:)` works as expected.
   - Floor Filter test
     - Selects a site, facility, changes the selected level and collapses the picker

Issue #362, opened 2023-06-28
UI Testing


dfeinzimer:
Toolkit UI Testing

We can approach this from two ways:
1. Add a new app that contains a UI testing target
  - Pro: Better seperation of purpose
  - Con: More overhead with new app to manage  
2. Add a UI testing target to the existing samples app
  - Con: We may need to introduce accessibility code into the examples that'll add extra complexity

mhdostal:
We'll also need to amend the current test designs to include UI testing.

dfeinzimer:
> We'll also need to amend the current test designs to include UI testing.

Would that be adding a new UI test design spec for each component in the common toolkit repo?

PR #367, opened 2023-07-11
Add UI Test Runner target to the Examples app

dfeinzimer:
Related https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/362, Apollo 47

This PR serves as an alternative to #364.

#364 creates an entire new testing app with a UI testing target. While writing the starter tests within, I noticed that the test views weren't requiring any differences from the example views we've already written in the Examples app. For that reason, I've opened this PR which simply adds a UI testing target to the existing examples app.


mhdostal:
This is certainly cleaner and more convenient than the original PR. However, I'm wondering if we're not trying to solve two separate issues now with the Examples. They are mainly to showcase the component and demonstrate how to use it. Two somewhat related things:

- Are the examples, as they're defined, enough to fully test the component? Meaning, is there component functionality that we'd want to test which is not exposed by the example?

- Given that the examples are for showcasing/demonstrating the component, I don't want to start "cluttering" them in order to get more test coverage.

Thoughts?

rolson:
> Are the examples, as they're defined, enough to fully test the component?

No.

> Given that the examples are for showcasing/demonstrating the component, I don't want to start "cluttering" them in order to get more test coverage.

I strongly agree.

The two goals should not be conflated. Tests are tests, examples are to show the minimum code required to use a component. I don't see these goals as compatible.

dfeinzimer:
It's a good point and yes I can see this as an issue. When we go to write tests for the Form component we may want to use a number of different forms with differing fields and configurations.

mhdostal:
@rolson Thanks for the comments, I think we should keep the two separate, examples and tests. Yes, for the simpler components the test view and example view may be very similar, but for more complex components they will diverge quite a bit.

dfeinzimer:
Closing in favor of #364

PR #368, opened 2023-07-14
[FormView] Add form view model and support for persisting text field edits

dfeinzimer:
- Adds a form view model to provide communication between a `FormView` and its host; adapts the implementation to use the model.
- Adds support to undo and submit local changes via the new model.
- Modifies the example view to make use of the new cancel and submit functionality.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Form/FormInputs/SingleLineTextEntry.swift
+++ b/Sources/ArcGISToolkit/Components/Form/FormInputs/SingleLineTextEntry.swift
@@ -18,11 +18,13 @@ import SwiftUI
 struct SingleLineTextEntry: View {
     @Environment(\.formElementPadding) var elementPadding
     
+    @EnvironmentObject var model: FormViewModel
+    
     /// A Boolean value indicating whether or not the field is focused.
     @FocusState private var isFocused: Bool
     
     /// The current text value.
-    @State private var text: String
+    @State private var text = ""
     
     /// The form element that corresponds to this text field.
     private let element: FieldFeatureFormElement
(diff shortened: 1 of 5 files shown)


Sources/ArcGISToolkit/Components/Form/FormViewModel.swift
```diff
@@ -0,0 +1,71 @@
...
+
+public class FormViewModel: ObservableObject {
+    /// The geodatabase which holds the table and feature being edited in the form.
+    @Published private var database: ServiceGeodatabase?
+    
+    /// The featured being edited in the form.
+    @Published private(set) var feature: ArcGISFeature?
+    
+    /// The structure of the form.
+    @Published var formDefinition: FeatureFormDefinition?
```
mhdostal, line 26: when does this get populated?

Sources/ArcGISToolkit/Components/Form/FormInputs/SingleLineTextEntry.swift
```diff
@@ -62,5 +63,11 @@ struct SingleLineTextEntry: View {
             input: input
         )
         .padding([.bottom], elementPadding)
+        .onAppear {
+            text = model.feature?.attributes[element.fieldName] as? String ?? ""
```
mhdostal, line 67: Once we switch to the core implementation, this would simply be `input.value`, as core will handle everything else.

Sources/ArcGISToolkit/Components/Form/FormViewModel.swift
```diff
@@ -0,0 +1,71 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ArcGIS
+import FormsPlugin
+import SwiftUI
+
+public class FormViewModel: ObservableObject {
```
mhdostal, line 18: I think having the model is a good step. Part of the "behavior" task is to figure out how to update a visible field element value when that value changes as the result of an expression evaluation. The model approach may be the way to go, where each element has a small model class with the `@State` property representing the value; that value would be updated (if necessary) after every successful expression evaluation.

mhdostal (commented):
This look good so far.

mhdostal (changes requested):
The more I think about this, the more I think the "FormViewModel" will simply be the "FeatureForm" API object.

mhdostal (approved):

PR #372, opened 2023-07-20
[FormView] Add date picker support

dfeinzimer:
- Adds date picker support to the Form View component

Finalized design pending

### DIFF
--- a/Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
+++ b/Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
@@ -42,7 +42,7 @@ struct DateTimeEntry: View {
                     .datePickerStyle(.graphical)
             } else {
                 Group {
-                    TextField(element.label, text: Binding(get: { String(describing: date) }, set: { _ in }))
+                    TextField(element.label, text: Binding(get: { formattedDate }, set: { _ in }))
                         .formTextEntryStyle()
                         .disabled(true)
                 }
@@ -79,6 +79,14 @@ struct DateTimeEntry: View {
             DatePicker(selection: $date, in: ...max, displayedComponents: displayedComponents) { EmptyView() }
         } else {
             DatePicker(selection: $date, displayedComponents: displayedComponents) { EmptyView() }
+        }
+    }
+    
+    var formattedDate: String {
+        if input.includeTime {
+            return date.formatted(.dateTime.day().month().year().hour().minute())
+        } else {
+            return date.formatted(.dateTime.day().month().year())
         }
     }
     
--- a/Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
+++ b/Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
@@ -78,12 +78,10 @@ struct DateTimeEntry: View {
     /// - Note: Secondary foreground color is used across entry views for consistency.
     @ViewBuilder var dateDisplay: some View {
         HStack {
-            TextField(
-                element.label,
-                text: Binding { date == nil ? "" : formattedDate } set: { _ in },
-                prompt: .noValue.foregroundColor(.secondary)
-            )
-            .disabled(true)
+            Text(formattedDate ?? .noValue)
+                .foregroundColor(displayColor)
+            
+            Spacer()
             
             if isEditing {
                 todayOrNowButton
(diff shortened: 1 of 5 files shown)


dfeinzimer:
After reviewing the Apps Team's implementation, I'm going back to matching the spec, that is showing the textual date along with the picker at the same time.

mhdostal:
I can open two date pickers at the same time; I think that when a field looses focus, the picker should go away:
image

Also, the date picker in the Apple apps that show the calendar view ( as in Reminders, Calendar) allow you to toggle the picker display be tapping on the date in the field. They also show the date in the accent color when the picker is open. (Only the Calendar app toggles the color, which is Red); the Reminders app always shows it in blue and doesn't toggle it). I like the color toggle.

Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
```diff
@@ -0,0 +1,208 @@
...
+        if let min = input.min, let max = input.max {
+            DatePicker(selection: Binding($date)!, in: min...max, displayedComponents: components) { }
+        } else if let min = input.min {
+            DatePicker(selection: Binding($date)!, in: min..., displayedComponents: components) { }
+        } else if let max = input.max {
+            DatePicker(selection: Binding($date)!, in: ...max, displayedComponents: components) { }
+        } else {
+            DatePicker(selection: Binding($date)!, displayedComponents: components) { }
+        }
+    }
```
mhdostal, line 120: This looks a little cluttered to me, what about putting the range calculation in a different method:
```suggestion
    /// Controls for making a specific date selection.
    @ViewBuilder var datePicker: some View {
        let components: DatePicker.Components = input.includeTime ? [.date, .hourAndMinute] : [.date]
        if let range = dateRange {
            DatePicker(selection: Binding($date)!, in: range, displayedComponents: components) { }
        } else {
            DatePicker(selection: Binding($date)!, displayedComponents: components) { }
        }
    }
    
    var dateRange: ClosedRange<Date>? {
        if let min = input.min, let max = input.max {
            return min...max
        } else if let min = input.min {
            return min...Date.distantFuture
        } else if let max = input.max {
            return Date.distantPast...max
        } else {
            return nil
        }
    }
```

mhdostal: I was wondering if the `nil` case above could be `return Date.distantPast...Date.distantFuture`; would that be equivalent? You could get rid of the `if let dateRange...` above.

mhdostal:
For the last date field, the one with a min/max date, the dates are goofy. The Form says they are: 07/01/1969 - 07/31/1969, but the form is displaying "February 1580"
image

The Unix dates, as they come through seem to be bad.

yo1995:
> The Form says they are: 07/01/1969 - 07/31/1969, but the form is displaying "February 1580" 

[Jeff Dean](https://github.com/JosXa/jeff-dean-bot/blob/bb16cff167204b6bb9f6726acadc22915ca36651/facts.txt#L57) and [Pope Gregory XIII](https://en.wikipedia.org/wiki/Gregorian_calendar) patched the space-time continuum after seeing those goofy dates, thence our relatively stable timeline nowadays 🤣

Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
dfeinzimer: Unless we need to support dates outside this range, I think we can go with that.

<p align=center>
Screenshot 2023-07-21 at 17 49 53
</p>

Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
```diff
@@ -0,0 +1,208 @@
...
+                Image(systemName: "calendar")
+                    .foregroundColor(.secondary)
+            } else {
+                ClearButton { date = nil }
+            }
+        }
+        .formTextEntryStyle()
+        .onTapGesture {
+            if date == nil { date = .now }
+            withAnimation { isEditing = true }
```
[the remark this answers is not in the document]
dfeinzimer: I've been able to figure a system out using a `@Published` property on the view model. It seems to be working well so far.

mhdostal (changes requested):

dfeinzimer:
Added support for handling read only dates, still ready for review

mhdostal (approved):

PR #373, opened 2023-07-21
Remove `LoginViewModel`

philium:
I ran across it again recently and realized that it is just an unnecessary level of indirection. So I removed it, moving all of its necessary stored properties onto the modifier itself and removing the properties that weren't needed.


Sources/ArcGISToolkit/Components/Authentication/LoginViewModifier.swift
```diff
@@ -0,0 +1,98 @@
+// Copyright 2022 Esri.
```
dfeinzimer, line 1:
```suggestion
// Copyright 2023 Esri.
```

philium: I didn't change it originally because it is still the original `Login.swift` file. But since it has been renamed and significantly changed, I suppose it is best to treat it as a new file.

dfeinzimer (approved):

mhdostal (approved):

PR #376, opened 2023-07-27
cancel the challenge if OAuth user credential creation is cancelled

njarecha:
Cancelling the challenge will give user a new `ArcGISChallengeCancellationError` with underlying error.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Authentication/Authenticator.swift
+++ b/Sources/ArcGISToolkit/Components/Authentication/Authenticator.swift
@@ -53,6 +53,9 @@ extension Authenticator: ArcGISAuthenticationChallengeHandler {
             do {
                 return .continueWithCredential(try await OAuthUserCredential.credential(for: configuration))
             } catch is CancellationError {
+                // If user cancels the creation of OAuth user credential then catch the
+                // cancellation error and cancel the challenge. This will make the request which
+                // issued the challenge fail with `ArcGISChallengeCancellationError`.
                 return .cancel
             }
         } else {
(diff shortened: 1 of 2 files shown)


Sources/ArcGISToolkit/Components/Authentication/Authenticator.swift
```diff
@@ -50,7 +50,11 @@ extension Authenticator: ArcGISAuthenticationChallengeHandler {
         
         // Create the correct challenge type.
         if let configuration = oAuthUserConfigurations.first(where: { $0.canBeUsed(for: challenge.requestURL) }) {
-            return .continueWithCredential(try await OAuthUserCredential.credential(for: configuration))
+            do {
+                return .continueWithCredential(try await OAuthUserCredential.credential(for: configuration))
+            } catch is CancellationError {
+                return .cancel
+            }
```
rolson, line 57: I don't understand why this change is needed

njarecha: So the request will fail with `ArcGISChallengeCancellationError with underlying error` instead of `CancellationError`.

rolson: should It be catching `ArcGISChallengeCancellationError` then?

rolson: Never mind, I think I understand. I do think a comment would be good here probably

njarecha: done

rolson (approved):

dfeinzimer (approved):

njarecha:
Thanks @rolson and @dfeinzimer!

PR #381, opened 2023-08-01
Authentication Alerts -> Authentication Sheets

### LINKED ISSUE #315: Authentication challenges can be lost
Originally reported by @carl8382 

> With the authenticator in the toolkit challenges can be completely lost if there is an alert (or sheet) being presented. The entire challenge basically goes into the void and I don’t know of a way to get it back.

- We've considered removing this `ZStack` which would result in alerts being displayed directly on the content rather than a sibling but it can cause other unintended behavior. For example, if the authentication challenge is triggered from a sheet displayed over a map view in a navigation view, the navigation link will automatically back out.
https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/30be9eee14d650c0bd9d4a8abd552aca244ec80e/Sources/ArcGISToolkit/Components/Authentication/AuthenticatorModifier.swift#L31-L42

- Another idea is to simply install the authenticator modifier on top of any view that may trigger an auth challenge.

- Presenting alerts from the top most presented view controller was recommended against.

- A single alert installed at the app level paired with an alert class that can could be accessed anywhere through an `@EnvironmentObject` is promising but sheets would still need to install their own alert modifier.

rolson:
This should partially fix #315.

iOS is more flexible with Sheets than alerts in two ways:

- A sheet can present another sheet on top of an existing sheet in certain situations
- In other situations, the operating system actually queues up sheets if you try to present one over another, and the second  sheet shows up after the first is dismissed.

With both of these in mind, I think I we can improve that toolkit issue if we move away from alerts and move to sheets for prompts.

For instance, we can prompt like images below.

I've tested this with authentication challenges coming in when you have a sheet showing.  You can get either behavior described above (sheet over sheet or queued up) depending on how you structure the app.

If you have a sheet showing and you do not want the queuing behavior, then you must embed a call to the `.authenticator(...)` view modifier from the content showing in the sheet.  This will allow the challenge sheet to be shown over top of the existing sheet. For example:

```swift
struct ContentView: View {
    @EnvironmentObject var authenticator: Authenticator
    @State var isSheetPresented: Bool = false
    
    var body: some View {
        VStack {
            AuthenticationExampleView()
            Button("sheet!") {
                isSheetPresented = true
            }
            .padding()
        }
        .sheet(isPresented: $isSheetPresented) {
            AuthenticationExampleView()
            // This causes the challenge to show within the context of a visible sheet
            // only while it's visible.
                .authenticator(authenticator)
        }
    }
}
```

If an authentication challenge comes while the app is showing an Alert, then the authentication challenge will be lost. This shouldn't be too much of a problem because alerts are more transient than sheets in most app workflows. 

Hopefully Alerts are something that we can deal with better at some point.  

With this in mind, I think this is a worthwhile improvement.

Also, with that branch you do not need to use the `._oAuthWebViewSheet()` modifier anywhere. This PR includes that as part of the `.authenticator` modifier.

<details>

<summary> Username / Password entry </summary>

</details>

<details>
<summary> Browse For Certificate Prompt </summary>

</details>

<details>
<summary> Server Trust Prompt </summary>

</details>


Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
```diff
@@ -181,14 +184,22 @@ struct CertificatePickerViewModifier: ViewModifier {
...
                 viewModel: viewModel
             )
+            .task {
+                // Present the prompt right away.
+                // Setting it after initialization allows it to animate.
+                // However, this needs to happen after a slight delay or
+                // it doesn't show.
+                try? await Task.sleep(nanoseconds: 250_000_000)
+                viewModel.showPrompt = true
+            }
```
dfeinzimer, line 202: I see this task is used 3x throughout. Would a custom modifier potentially be better? The doc comment will loose the context but that could potentially be alleviated by good documentation of the new modifier: e.g.

```swift
/// Performs the provided action after a slight delay.
/// - Tip: Occasionally delaying allows a sheet's presentation animation to work correctly.
/// - Parameters:
///   - delay: The delay to wait before allow the task to proceed, in nanoseconds. Defaults to 1/4 second.
///   - action: The action to perform after the delay.
func delayedTask(delay: UInt64 = 250_000_000, action: @escaping () -> Void) -> some View {
    task {
        try? await Task.sleep(nanoseconds: delay)
        action()
    }
}
```

rolson: I think that's a great idea to explore. Let me look into that.

rolson:
@dfeinzimer - if you want to try this out I have a test app here: https://devtopia.esri.com/ryan3374/AuthenticatorFunctionalSuite.

There is a bug that I found that I need to fix: if you dismiss a sheet by swiping it doesn't cancel the challenge.  I need to look into that.

Sources/ArcGISToolkit/Utility/CredentialInputView.swift
```diff
@@ -14,59 +14,145 @@
...
+    
+    /// The cancel action.
+    let cancelAction: CredentialInputSheetView.Action
+    
+    /// The continue action.
+    let continueAction: CredentialInputSheetView.Action
+    
+    @ViewBuilder func body(content: Content) -> some View {
+        content
+            .sheet(isPresented: isPresented) {
```
dfeinzimer, line 93: > There is a bug that I found that I need to fix: if you dismiss a sheet by swiping it doesn't cancel the challenge. I need to look into that.

Yes, I'd actually noticed this but forgot to mention it, sorry! Something like this may fix it. It's just unclear what the proper handler arguments would be.

```suggestion
            .sheet(isPresented: isPresented) {
                cancelAction.handler("", "")
            } content: {
```

Sources/ArcGISToolkit/Utility/CredentialInputView.swift
```diff
@@ -228,68 +283,13 @@ extension CredentialInputView {
...
-            )
-        }
+extension CredentialInputSheetView {
+    /// A configuration for an alert action.
+    struct Action {
+        /// The title of the action.
+        let title: String
+        
+        /// The block to execute when the action is triggered.
+        let handler: (String, String) -> Void
```
dfeinzimer, line 293: Should we document the purpose of the two parameters?

```suggestion
        let handler: (String, String) -> Void
```

Sources/ArcGISToolkit/Components/Authentication/CertificatePickerViewModifier.swift
rolson: I think the title of the method needs to be delayedOnAppear because the action is not async so it's not really the same as task.  I incorporated your idea with some slight modifications.

Sources/ArcGISToolkit/Utility/CredentialInputView.swift
rolson: I thin the better thing to do is to prevent the interactive dismiss. We can make the user deal with the challenge explicitly.

Sources/ArcGISToolkit/Utility/CredentialInputView.swift
```diff
@@ -14,59 +14,146 @@
...
+        Text("test")
+            .credentialInput(
+                isPresented: .constant(true),
+                fields: .usernamePassword,
+                message: "You must sign in to access 'arcgis.com'",
+                title: "Authentication Required",
+                cancelAction: .init(
+                    title: "Cancel",
+                    handler: { _, _ in
+
```
dfeinzimer, line 58:
```suggestion
```

dfeinzimer (approved):

nick-esri:
@rolson just chiming in with some design feedback, per your request via @carl8382 ....

Although technically possible, I don't think moving away from alerts is preferable here. 

Considerations:
- While both sheets and alerts are modal, the alerts we're currently using feel the most natural when it comes to sign in challenges, certificate requests and warnings: it provides critical information to the user, and appropriate actions. 
- The alert supports all we need to handle the information we want to display: a title, informative text, text fields and action buttons. 
- The sheet examples you posted use a lot of space to present the same information, and feels non-standard for the use case. 
- On the iPad the default presentation for a sheet is a page sheet, which again might feel like too much space for a small number of controls.

rolson:
There is some minor follow up work on this that needs to take place for some minor UI preferences. That will be done in a follow up PR.

dfeinzimer (approved):

rolson:
@dg0yal the only string that changed that I am aware of is this:

image

dfeinzimer:
> @dg0yal the only string that changed that I am aware of is this:
> 
> image

@rolson I will merge this into the localization fixes branch I'm working on.

PR #377, opened 2023-07-27
Core Forms implementation

mhdostal:
This switches the FormView to use the new Core implementation of Forms and removed FormsPlugin (RIP).

This should not be merged until after we branch to release and the Core Client YAML update PR gets merged.

PR #382, opened 2023-08-02
Form View - Add `FormViewTests` text UI tests

dfeinzimer:
Adds the three Form View text UI cases

Case `1_3`'s design specifies to verify the helper text turns red. From what I've found there's no way to check properties of UI elements such as a text's color. I researched a few routes including bringing over the image comp. logic from the SDK but that turned out to be a significant amount of code. Vision has a [computeDistance(_:to:)](https://developer.apple.com/documentation/vision/vnfeatureprintobservation/3182823-computedistance) method but I ran into [an issue](https://developer.apple.com/forums/thread/721760#:~:text=Am%20also%20having%20the%20same%20issue%20mentioned%20twice%20above%20(the%2019.737993%20magic%20number).) while experimenting with it. The other alternate would be to write a view model and these tests as unit tests instead but that also seems like a lot just to verify text color.

### DIFF
--- a/UI Test Runner/UI Test Runner/TestViews/FormViewTestView.swift
+++ b/UI Test Runner/UI Test Runner/TestViews/FormViewTestView.swift
@@ -16,6 +16,8 @@ import ArcGISToolkit
 import SwiftUI
 
 struct FormViewTestView: View {
+    @Environment(\.verticalSizeClass) var verticalSizeClass
+    
     /// The `Map` displayed in the `MapView`.
     @State private var map = Map(url: .sampleData)!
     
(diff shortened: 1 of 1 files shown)


Examples/Examples/FormExampleView.swift
```diff
@@ -44,22 +44,25 @@ struct FormExampleView: View {
...
-                            .presentationBackgroundInteraction(.enabled(upThrough: .medium))
-                            .presentationDetents([.medium])
-                    } else {
-                        FormView()
-                            .padding()
-                    }
-                    #if targetEnvironment(macCatalyst)
-                    Button("Dismiss") {
-                        isPresented = false
+                    Group {
```
dfeinzimer, line 47: The `.environmentObject(formViewModel)` is moved into the sheet, otherwise it is not made available in the hierarchy in Mac Catalyst. Opening the form resulted in a crash.

UI Test Runner/UITests/FormViewTests.swift
```diff
@@ -0,0 +1,70 @@
...
+        app.launch()
+        
+        let formViewTestsButton = app.buttons["Form View Tests"]
+        let inputValidationLabel = app.staticTexts["InputValidation"]
+        let singleLineNoValueNoPlaceholderNoDescriptionTitle = app.staticTexts[
+            "Single Line No Value, Placeholder or Description"
+        ]
+        let singleLineNoValueNoPlaceholderNoDescriptionTextField = app
+            .scrollViews.otherElements.containing(.staticText, identifier: "InputValidation")
+            .children(matching: .textField).element(boundBy: 1)
```
dfeinzimer, line 29: This query will be simplified in an upcoming PR.

UI Test Runner/UITests/FormViewTests.swift
```diff
@@ -0,0 +1,70 @@
...
+        )
+        
+        // Scroll to the target form element.
+        while !(singleLineNoValueNoPlaceholderNoDescriptionTitle.isHittable) {
+            app.scrollViews.firstMatch.swipeUp(velocity: 750)
+        }
+        
+        XCTAssertTrue(
+            singleLineNoValueNoPlaceholderNoDescriptionTextField.isHittable,
+            "The target text field wasn't found within 30 seconds."
```
mhdostal, line 49: This could be 5, but if it changes again, maybe we just say the following?
```suggestion
            "The target text field wasn't found within the expected time."
```

dfeinzimer:
Taking this back to draft until #377 is merged. No need to review again just yet. The example view changes may be able to drop off entirely.

dfeinzimer:
The test view has been updated to work with the new Core API. Ready for review once again.

PR #384, opened 2023-08-03
Form View - Add remaining text UI test cases

dfeinzimer:
- Adds test cases `1_2` and `1_3`
- Revises `1_1` to align with the two new cases.

**More:** Case `1_3`'s design specifies to verify the helper text turns red. From what I've found there's no way to check properties of UI elements such as a text's color. I researched a few routes including bringing over the image comp. logic from the SDK but that turned out to be a significant amount of code. `Vision` has a [`computeDistance(_:to:)`](https://developer.apple.com/documentation/vision/vnfeatureprintobservation/3182823-computedistance) method but I ran into [an issue](https://developer.apple.com/forums/thread/721760#:~:text=Am%20also%20having%20the%20same%20issue%20mentioned%20twice%20above%20(the%2019.737993%20magic%20number).) while experimenting with it. The other alternate would be to write a view model and these tests as unit tests instead but that also seems like a lot just to verify text color.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Form/FormInputs/SingleLineTextEntry.swift
+++ b/Sources/ArcGISToolkit/Components/Form/FormInputs/SingleLineTextEntry.swift
@@ -51,8 +51,10 @@ struct SingleLineTextEntry: View {
         HStack {
             TextField(element.label, text: $text, prompt: Text(element.hint ?? "").foregroundColor(.secondary))
                 .focused($isFocused)
+                .accessibilityIdentifier("\(element.label) Text Field")
             if !text.isEmpty {
                 ClearButton { text.removeAll() }
+                    .accessibilityIdentifier("\(element.label) Clear Button")
             }
         }
         .formTextEntryStyle()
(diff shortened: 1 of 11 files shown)


Sources/ArcGISToolkit/Components/Form/FormInputs/SingleLineTextEntry.swift
```diff
@@ -51,8 +51,10 @@ struct SingleLineTextEntry: View {
         HStack {
             TextField(element.label, text: $text, prompt: Text(element.hint ?? "").foregroundColor(.secondary))
                 .focused($isFocused)
-            if isFocused && !text.isEmpty {
```
dfeinzimer, line 54: Turns out that was misaligned with the design, the clear button should be visible whenever the field has a value.

Sources/ArcGISToolkit/Components/Form/FormInputs/SingleLineTextEntry.swift
```diff
@@ -51,8 +51,10 @@ struct SingleLineTextEntry: View {
...
+                .accessibilityIdentifier("\(element.label) Text Field")
```
dfeinzimer, line 54: These have been added for testing purposes only.

mhdostal (approved):

Sources/ArcGISToolkit/Components/Form/FormInputs/SingleLineTextEntry.swift
philium: @mhdostal You're thinking of [`accessibilityLabel`](https://developer.apple.com/documentation/objectivec/nsobject/1615181-accessibilitylabel), which is localized.

PR #391, opened 2023-08-09
Form View - `NavigationView` removal

dfeinzimer:
Related Apollo # 104

- Removes the problematic `NavigationView` from `FormView`'s hierarchy.
- Removes the "Done" button from the keyboard toolbar for multiline text fields ~and adds a circular check mark button (a la Android) that clears focus~


philium:
> a la Android

I don't think we should be borrowing any design elements that are unique to Android unless absolutely necessary. On iOS, text fields in forms don't usually have clear buttons because editing ends when the form is scrolled. An example of this for a multiline text field is the notes field of the Contact view in Address Book. Is that not the case (scrolling the form doesn't end editing)? I can't remember if that is a default (built-in) behavior or not, but if it isn't, we could add it. Better than using an Android UI design element.

dfeinzimer:
> Is that not the case (scrolling the form doesn't end editing)? I can't remember if that is a default (built-in) behavior or not,

Yes, that is the case. @mhdostal, any thoughts on using scrolling instead of the check?

mhdostal:
> Yes, that is the case. @mhdostal, any thoughts on using scrolling instead of the checkmark?

If the scrolling works and that follows what other Apple apps to, then that's what we should do. I don't like the checkmark.  In any event, I think the specific issue of "Done" vs. "Scrolling" is not important at the moment. I'd create an issue and we can get to it later.

dfeinzimer:
> I'd create an issue and we can get to it later.

Will remove the done button and revisit scroll-away later as apart of Apollo 128.

mhdostal (approved):

PR #380, opened 2023-07-31
FloatingPanel Keyboard Avoidance

dfeinzimer:
| Before | After |
|---|---|
| <p>https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/assets/16397058/846270dd-ab92-49fd-942a-1f5d68ef01dd</p>  | <p>https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/assets/16397058/dbaa7dff-1f00-4863-b9ab-a47ee81385bb</p>  |

PR #394, opened 2023-08-14
Repair `FloatingPanel` bottom content padding

dfeinzimer:
This replaces a line that was erroneously removed in #380. Without it, content towards the bottom of the FloatingPanel can collide with the Home Bar or the panel's handle divider.

First reported by @des12437 during release testing

|   | HomeBar  | Panel Handle Divider |
|---|---|---|
| Padding  |  homebar  | divider |
| No Padding  |  Screenshot 2023-08-14 at 09 07 03  |  Screenshot 2023-08-14 at 09 07 27 |


des12437 (approved):

dfeinzimer:
> Looks good. Is there a case where the padding should be zero based on the previous implementation?

Good question, I had considered that and I don't believe taking `isPresented` into account is needed.

mhdostal (approved):

Issue #328, opened 2023-05-09
BasemapGallery - Clone basemap prior to setting to avoid "object already owned" error.


mhdostal:
When the client passes in a list of base maps, and if multiple web maps are used with the BasemapGallery, you can get the "object already owned error". The solution is to clone the basemap prior to setting the new basemap.

Cloning is NOT necessary when the user passes in a portal, as the portal basemaps are reloaded every time the basemap gallery is displayed.

dfeinzimer:
@mhdostal I'm trying to repro this to adjust the fix for #1126 but I'm not understanding the description:

> and if multiple web maps are used with the BasemapGallery

Do you recall what the steps to reproduce were?

mhdostal:
In `BasemapGalleryExampleView`, create a second map property:
```
    @State private var map2: Map = {
        let map = Map(basemapStyle: .arcGISImagery)
        map.initialViewpoint = Viewpoint(
            center: Point(x: -93.258133, y: 44.986656, spatialReference: .wgs84),
            scale: 1_000_000
        )
        return map
    }()
```

Then, create a `.task` on the `MapView` in the body of the example:

```
            .task {
                map2.basemap = basemaps.first?.basemap
                try? await map2.load()
            }
```

Then run the example and set the basemap of the original map to the first basemap in the list. If there is no `.clone()` call, you get the `Object already owned` error.

PR #395, opened 2023-08-14
BasemapGallery test fix

mhdostal:
Fixes the "init" test in response to changes in #328


dfeinzimer (commented):
The test passes but then a few moments later I get a crash in core, still looking

Screenshot 2023-08-14 at 10 15 11

mhdostal:
> The test passes but then a few moments later I get a crash in core in response handling, still looking.

Weird. I was running all the tests and I didn't get a crash. If I run just the BasemapGallery tests or simply the `Init` test, it crashes.

dfeinzimer (approved):
For what its worth, seems to crash on `main`/`200.1` so I don't think this is a regression. Should we give this its own issue?

mhdostal:
> Should we give this its own issue?

Yes, it's own issue. If it's not a regression, then we'll fix it next release.

PR #397, opened 2023-08-14
Fix compiler warning in BookmarksTest

mhdostal:
https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/398

### DIFF
--- a/Tests/ArcGISToolkitTests/BookmarksTests.swift
+++ b/Tests/ArcGISToolkitTests/BookmarksTests.swift
@@ -57,20 +57,14 @@ final class BookmarksTests: XCTestCase {
             set: {_isPresented = $0 }
         )
         
-        var actionBookmark: Bookmark?
-        let action: ((Bookmark) -> Void) = {
-            actionBookmark = $0
-        }
-        var bookmarks = Bookmarks(
-            isPresented: isPresented,
-            geoModel: map
-        )
-        bookmarks.selectionChangedAction = action
+        var selectedBookmark: Bookmark?
+        let bookmarks = Bookmarks(isPresented: isPresented,geoModel: map)
+            .onSelectionChanged { selectedBookmark = $0 }
         XCTAssertTrue(_isPresented)
         let firstBookmark = try XCTUnwrap(map.bookmarks.first)
         bookmarks.selectBookmark(firstBookmark)
         XCTAssertFalse(_isPresented)
-        XCTAssertEqual(actionBookmark?.viewpoint, map.bookmarks.first?.viewpoint)
+        XCTAssertEqual(selectedBookmark, map.bookmarks.first)
     }
     
     /// Asserts that the list properly handles a selection when provided a viewpoint.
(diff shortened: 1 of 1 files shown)


Tests/ArcGISToolkitTests/BookmarksTests.swift
```diff
@@ -72,7 +72,7 @@ final class BookmarksTests: XCTestCase {
         let firstBookmark = try XCTUnwrap(map.bookmarks.first)
         bookmarks.selectBookmark(firstBookmark)
         XCTAssertFalse(_isPresented)
-        wait(for: [expectation], timeout: 1.0)
+        await fulfillment(of: [expectation], timeout: 1.0)
```
philium, line 75: FYI: This API was [introduced in Xcode 14.3](https://developer.apple.com/documentation/xcode-release-notes/xcode-14_3-release-notes#Testing), which is above the minimum Xcode version for this project.

mhdostal: @philium Thanks for catching that. This will have to wait then.

philium: Since Xcode 14.3 shipped with a newer version of the Swift compiler, you could use conditional compilation:

```swift
#if compiler(>=5.8)
        await fulfillment(of: [expectation], timeout: 1.0)
#else
        wait(for: [expectation], timeout: 1.0)
#endif
```

However, the whole reason for the new API is that the old API could cause a test to deadlock when used in concurrent test methods. So it might be better to avoid using expectations in this method for now.

mhdostal:
I've replaced the expectation with another method. Please re-review.

dfeinzimer (dismissed):

Tests/ArcGISToolkitTests/BookmarksTests.swift
```diff
@@ -59,9 +56,10 @@ final class BookmarksTests: XCTestCase {
             get: { _isPresented },
             set: {_isPresented = $0 }
         )
+        
+        var actionBookmark: Bookmark?
         let action: ((Bookmark) -> Void) = {
```
philium, line 61: Declaring a variable for this isn't necessary. Also, since the stored value is the selected bookmark, I would suggest calling the variable that stores the bookmark `selectedBookmark`. In addition, just compare the bookmarks themselves rather than their viewpoints. Lastly, `bookmarks` doesn't need to be mutable. It would be better to test using the public modifier. Putting that all together:

```swift
var selectedBookmark: Bookmark?
let bookmarks = Bookmarks(isPresented: isPresented,geoModel: map)
    .onSelectionChanged { selectedBookmark = $0 }
XCTAssertTrue(_isPresented)
let firstBookmark = try XCTUnwrap(map.bookmarks.first)
bookmarks.selectBookmark(firstBookmark)
XCTAssertFalse(_isPresented)
XCTAssertEqual(selectedBookmark, map.bookmarks.first)
```

mhdostal: Good stuff, thanks!

philium: You're welcome!

philium (approved):

PR #404, opened 2023-08-15
Floating Panel - Presentation/usability improvements

dfeinzimer:
Closes Apollo # 103

- Changes the Floating Panel to lazily evaluate the content only upon presentation. Currently the content will be evaluated as soon as the view containing the Floating Panel is loaded, meaning users have to do odd things like this to prevent content from being evaluated too early:
https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/f75f901995fc3efc74671c896286571df55b980a/Examples/Examples/PopupExampleView.swift#L75-L78
- Simple content such as `Text` that doesn't greedily fill its container width-wise would cause the panel to shrink horizontally when the panel is hiding away. This is also fixed and new demo content has been added to demonstrate the fix.

### DIFF
--- a/Examples/Examples/FloatingPanelExampleView.swift
+++ b/Examples/Examples/FloatingPanelExampleView.swift
@@ -39,26 +39,30 @@ struct FloatingPanelExampleView: View {
             viewpoint: initialViewpoint
         )
         .floatingPanel(selectedDetent: $selectedDetent, isPresented: isPresented) {
-            switch demoContent {
+            switch demoContent! {
             case .list:
                 FloatingPanelListDemoContent(selectedDetent: $selectedDetent)
             case .textField:
                 FloatingPanelTextFieldDemoContent(selectedDetent: $selectedDetent)
-            case .none:
-                EmptyView()
             }
         }
         .toolbar {
-            ToolbarItem(placement: .navigationBarTrailing) {
-                Menu {
-                    Button("List") {
-                        demoContent = .list
+            ToolbarItemGroup(placement: .navigationBarTrailing) {
+                if demoContent != nil {
+                    Button("Dismiss") {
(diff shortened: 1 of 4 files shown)


Sources/ArcGISToolkit/Components/FloatingPanel/FloatingPanelModifier.swift
```diff
@@ -51,7 +51,7 @@ public extension View {
                 horizontalAlignment: horizontalAlignment,
                 isPresented: isPresented,
                 maxWidth: maxWidth,
-                panelContent: content()
```
dfeinzimer, line 54: This is the other main piece of the change. The content was being evaluated right here in the modifier with no regards to the value of `isPresented`.

Examples/Examples/FloatingPanelExampleView.swift
```diff
@@ -39,26 +39,35 @@ struct FloatingPanelExampleView: View {
             viewpoint: initialViewpoint
         )
         .floatingPanel(selectedDetent: $selectedDetent, isPresented: isPresented) {
-            switch demoContent {
+            switch demoContent! {
```
mhdostal, line 42: Instead of having `demoContent` be optional and force-unwrapping it here, can we just default it to something, maybe `.list`?

mhdostal (changes requested):

Examples/Examples/FloatingPanelExampleView.swift
dfeinzimer: Yes, I've reverted this change. It wasn't 100% necessary.

mhdostal (approved):

yo1995 (approved):

PR #407, opened 2023-08-17
Cleanup `FloatingPanel`

philium:
Changes binding properties to use `Binding` as an attribute, eliminating the need to go through the `wrappedValue` property. That also allows us to use the compiler-generated memberwise initializer, so the handwritten one has been removed. Lastly, removes erroneous `public` attribute from the `body` property.


mhdostal (approved):

dfeinzimer (approved):
Thank you!

PR #412, opened 2023-08-24
Simplify `FormView` example

dfeinzimer:
Simplifies the `FormView` example to only demonstrate hosting it with a Floating Panel.

~Leaving in draft at the moment as there're current issues with the core forms implementation.~


dfeinzimer:
Marking ready for review as forms are operational again with core branch `ivan4258/form_fixes`

PR #420, opened 2023-08-30
Localized string key improvements

dfeinzimer:
Related Apollo 119

- Adds comments for the remaining localized string keys missing "engineer comments" resolving ambiguites highlighted by the translations team.

cc @dg0yal

### DIFF
--- a/Sources/ArcGISToolkit/Components/Popups/PopupView.swift
+++ b/Sources/ArcGISToolkit/Components/Popups/PopupView.swift
@@ -113,8 +113,12 @@ public struct PopupView: View {
                         )
                     }
                 } else {
-                    VStack(alignment: .center) {
-                        Text("Evaluating popup expressions…", bundle: .toolkitModule)
+                    HStack(alignment: .center, spacing: 10) {
+                        Text(
+                            "Evaluating popup expressions",
+                            bundle: .toolkitModule,
+                            comment: "A label indicating popup expressions are being evaluated."
+                        )
                         ProgressView()
                     }
                     .frame(maxWidth: .infinity)
--- a/Sources/ArcGISToolkit/Components/Search/SearchView.swift
+++ b/Sources/ArcGISToolkit/Components/Search/SearchView.swift
@@ -128,11 +128,15 @@ public struct SearchView: View {
     
     @Environment(\.horizontalSizeClass) var horizontalSizeClass
     @Environment(\.verticalSizeClass) var verticalSizeClass
-
+    
     /// The string shown in the search view when no user query is entered.
     /// Defaults to "Find a place or address". Note: this is set using the
     /// `prompt` modifier.
-    private var prompt = String(localized: "Find a place or address", bundle: .toolkitModule)
+    private var prompt = String(
+        localized: "Find a place or address",
+        bundle: .toolkitModule,
+        comment: "A hint as to what a user can search for within a search bar."
+    )
     
     /// Determines whether a built-in result view will be shown. Defaults to `true`.
     /// If `false`, the result display/selection list is not shown. Set to false if you want to hide the results
--- a/Sources/ArcGISToolkit/Components/Search/SearchView.swift
+++ b/Sources/ArcGISToolkit/Components/Search/SearchView.swift
@@ -223,8 +223,9 @@ public struct SearchView: View {
                         "Repeat Search Here",
                         bundle: .toolkitModule,
                         comment: """
-                                  A button to show when a user has panned the map away from the
-                                  original search location.
+                                  A label for button to show when a user has panned the map away
+                                  from the original search location. 'Here' is in reference to the
+                                  current visible extent of the map or scene.
                                   """
                     )
                 }
--- a/Sources/ArcGISToolkit/Components/Search/SearchView.swift
+++ b/Sources/ArcGISToolkit/Components/Search/SearchView.swift
@@ -135,7 +135,7 @@ public struct SearchView: View {
     private var prompt = String(
         localized: "Find a place or address",
         bundle: .toolkitModule,
-        comment: "A hint as to what a user can search for within a search bar."
+        comment: "A hint as to what the user can search for within a search bar."
     )
     
     /// Determines whether a built-in result view will be shown. Defaults to `true`.
@@ -227,7 +227,7 @@ public struct SearchView: View {
                         "Repeat Search Here",
                         bundle: .toolkitModule,
                         comment: """
-                                  A label for button to show when a user has panned the map away
+                                  A label for button to show when the user has panned the map away
                                   from the original search location. 'Here' is in reference to the
                                   current visible extent of the map or scene.
                                   """
--- a/Sources/ArcGISToolkit/Components/Search/SearchView.swift
+++ b/Sources/ArcGISToolkit/Components/Search/SearchView.swift
@@ -135,7 +135,7 @@ public struct SearchView: View {
     private var prompt = String(
         localized: "Find a place or address",
         bundle: .toolkitModule,
-        comment: "A hint for user what to search in a search bar."
+        comment: "A hint for the user on what to search for in a search bar."
     )
     
     /// Determines whether a built-in result view will be shown. Defaults to `true`.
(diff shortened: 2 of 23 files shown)


dfeinzimer:
@njarecha I was trying to re-run `xcodebuild -exportLocalizations -sdk <SDK> -localizationPath <PATH> -exportLanguage en` to validate the manually updated strings file in this PR but I was again unable to get it to generate. Do you recall the exact command you used to get the generation to succeed?

dfeinzimer:
@njarecha Disregard above, got it to work by using Swift 5.9. It must be an issue with Swift 5.7.

`xcodebuild -exportLocalizations -sdk iphoneos17.0 -localizationPath ./Sources/ArcGISToolkit/Resources/en.lproj  -exportLanguage en` writes an `.xcloc` to `localizationPath` containing a generated strings file.

Sources/ArcGISToolkit/Resources/en.lproj/Localizable.strings
```diff
@@ -68,205 +62,221 @@ second being a corresponding cardinal direction (north, northeast,
...
+/* A label for a button to show new utility network trace configuration options. */
 "New trace" = "New trace";
 
-/* No comment provided by engineer. */
+/* A label indicating that no bookmarks are available for selection. */
 "No bookmarks" = "No bookmarks";
 
-/* No comment provided by engineer. */
+/* A statement that no utility trace configurations are available. */
 "No configurations available" = "No configurations available";
```
dfeinzimer, line 152: The translators enquired if this string should have a period. I believe it is okay without one but I'm open to other interpretations.

Sources/ArcGISToolkit/Resources/en.lproj/Localizable.strings
```diff
@@ -68,205 +62,221 @@ second being a corresponding cardinal direction (north, northeast,
...
 
-/* No comment provided by engineer. */
+/* A statement that no sites or facilities with names matching a filter phrase were found. */
 "No matches found" = "No matches found";
```
dfeinzimer, line 155: The translators enquired if this string should have a period. I believe it is okay without one but I'm open to other interpretations.

Sources/ArcGISToolkit/Resources/en.lproj/Localizable.strings
```diff
@@ -68,205 +62,221 @@ second being a corresponding cardinal direction (north, northeast,
...
 
 /* A message to show when there are no results or suggestions. */
 "No results found" = "No results found";
```
dfeinzimer, line 158: The translators enquired if this string should have a period. I believe it is okay without one but I'm open to other interpretations.

Sources/ArcGISToolkit/Components/Popups/PopupView.swift
```diff
@@ -113,8 +113,12 @@ public struct PopupView: View {
...
-                    VStack(alignment: .center) {
-                        Text("Evaluating popup expressions…", bundle: .toolkitModule)
+                    HStack(alignment: .center, spacing: 10) {
+                        Text(
+                            "Evaluating popup expressions",
+                            bundle: .toolkitModule,
+                            comment: "A label indicating popup expressions are being evaluated."
+                        )
                         ProgressView()
                     }
```
dfeinzimer, line 123: The translators reported garbled characters in their translation software on this string. We can try alternatives like using the unicode value for the horizontal ellipsis instead but with the `ProgressView()` in an `HStack` right next to this I think we're okay without the `…` all together.

Screenshot 2023-08-30 at 16 54 14

Sources/ArcGISToolkit/Resources/en.lproj/Localizable.strings
njarecha: No strong opinion here but ok to add a period as it is used in a `Text` as a message and not a title of the button etc.

Sources/ArcGISToolkit/Resources/en.lproj/Localizable.strings
njarecha: No strong opinion here but ok to add a period as it is used in a `Text` as a message and not a title of the button etc.

Sources/ArcGISToolkit/Resources/en.lproj/Localizable.strings
njarecha: No strong opinion here but ok to add a period as it is used in a `Text` as a message and not a title of the button etc.

njarecha:
> @njarecha Disregard above, got it to work by using Swift 5.9. It must be an issue with Swift 5.7.
xcodebuild -exportLocalizations -sdk iphoneos17.0 -localizationPath ./Sources/ArcGISToolkit/Resources/en.lproj  -exportLanguage en writes an .xcloc to localizationPath containing a generated strings file.

If you include `exportLanguage en` then it will export the existing file and won't generate from he source. Try, `xcodebuild -exportLocalizations -sdk iphoneos17.0 -localizationPath  /Users/<username>/Desktop/ArcGISToolkit_localizations` and compare.

njarecha (commented):
Some suggestions

Sources/ArcGISToolkit/Components/Search/SearchView.swift
```diff
@@ -223,8 +227,9 @@ public struct SearchView: View {
                         "Repeat Search Here",
```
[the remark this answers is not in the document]
dfeinzimer: I think that seems right, no express reason for all the words to be capitalized.

Sources/ArcGISToolkit/Components/Search/SearchView.swift
```diff
@@ -220,11 +224,12 @@ public struct SearchView: View {
...
                 } label: {
                     Text(
-                        "Repeat Search Here",
+                        "Repeat search here",
                         bundle: .toolkitModule,
                         comment: """
-                                  A button to show when a user has panned the map away from the
-                                  original search location.
+                                  A label for button to show when the user has panned the map away
+                                  from the original search location. 'Here' is in reference to the
```
njarecha, line 231:
```suggestion
                                  from the original search location. 'here' is in reference to the
```

njarecha (changes requested):
Please update English strings file with the changes you made as per my earlier review.

njarecha (approved):
Looks good

mhdostal (changes requested):
Is it possible to move the "cancel" strings into the Strings.swift extension, like "value" and "field"? There are a lot of these:

```
                        localized: "Cancel",
                        bundle: .toolkitModule,
                        comment: "A label for a button to cancel an operation."
```

dfeinzimer:
> Is it possible to move the "cancel" strings into the Strings.swift extension, like "value" and "field"? There are a lot of these:

Good catch, done. I didn't realize how many "Cancel"s there were.

mhdostal (approved):

njarecha (approved):

dfeinzimer:
Thanks @njarecha and @mhdostal!

Sources/ArcGISToolkit/Components/Search/SearchView.swift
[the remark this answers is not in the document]
philium: There is if it is a button title, which it looks like it is. [The HIG says](https://developer.apple.com/design/human-interface-guidelines/buttons#Content) to use title style capitalization for button titles.

Sources/ArcGISToolkit/Extensions/String.swift
```diff
@@ -0,0 +1,43 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+extension String {
```
philium, line 16: Instead of putting these on `String`, consider that most—if not all—uses of these either directly or indirectly create an instance of `Text`. So if they were moved into an extension of `Text`, then they would be even simpler to use:

```swift
extension Text {
    static var cancel: Self {
        ...
    }
}
```

It might take a little refactoring in some places (when the string is passed around before creating the text, change the type of the parameter/property for the string to `Text`), but it is worth investigating at least.

dfeinzimer: #421

PR #422, opened 2023-09-01
Adjust search repeat button title

dfeinzimer:
Resolves [this](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/420#discussion_r1313466033) feedback


Sources/ArcGISToolkit/Resources/en.lproj/Localizable.strings
```diff
@@ -199,7 +199,7 @@ variable provides additional data. */
 /* A label for button to show when the user has panned the map away
 from the original search location. 'here' is in reference to the
```
philium, line 200:
```suggestion
from the original search location. 'Here' is in reference to the
```

The comment will also need to be updated in code.

philium (approved):

dfeinzimer:
Thanks for all of your reviews today Phil!

PR #424, opened 2023-09-07
`FormView` - Add DateTime picker UI tests, revise test case 1

dfeinzimer:
Apollo 94

- Adds UI tests for `FormView` DateTime fields
- Revises existing test case 1 tests

### DIFF
--- a/Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
+++ b/Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
@@ -187,7 +187,14 @@ struct DateTimeEntry: View {
     /// The button to set the date to the present time.
     var todayOrNowButton: some View {
         Button {
-            date = .now
+            let now = Date.now
+            if dateRange.contains(now) {
+                date = now
+            } else if now > dateRange.upperBound {
+                date = dateRange.upperBound
+            } else {
+                date = dateRange.lowerBound
+            }
         } label: {
             input.includeTime ? Text.now : .today
         }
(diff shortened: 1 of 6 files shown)


mhdostal (approved):

dfeinzimer:
This is ready for re-review. 

I found that switching the presentation over to a FloatingPanel destabilized the first test cases so I revised those. I also found that `testCase_1_3` had missing steps which I've now added. Lastly, I discovered that we don't need to explicitly scroll around the form. We can `tap()` elements to bring them into view. Anything that may be offscreen due to platform/panel size differences can still be tested with `.exists`.

Sources/ArcGISToolkit/Components/Form/FormView.swift
```diff
@@ -19,9 +19,6 @@ import SwiftUI
 public struct FormView: View {
     @Environment(\.formElementPadding) var elementPadding
     
-    /// The model for this form view.
-    @EnvironmentObject var model: FormViewModel
-
```
dfeinzimer, line 24: This property is unused for the time being.

Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
```diff
@@ -182,10 +187,18 @@ struct DateTimeEntry: View {
...
         Button {
-            date = .now
+            let now = Date.now
+            if dateRange.contains(now) {
+                date = now
+            } else if now > dateRange.upperBound {
+                date = dateRange.upperBound
+            } else {
+                date = dateRange.lowerBound
+            }
```
dfeinzimer, line 197: I found that the "Today"/"Now" button needs to be clamped to the date range.

mhdostal (approved):

PR #429, opened 2023-09-09
`FormView` - Switch to `FieldFormElement.value` from feature attributes + other minor fixes

dfeinzimer:
- Switches the date/time single and multi line form input fields to get their value from `FieldFormElement.value`
- Adds guards in `onAppear` bodies to prevent unnecessarily calling `setAttributeValue` with the value that's already stored.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
+++ b/Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
@@ -63,11 +63,11 @@ struct DateTimeEntry: View {
             if element.value.isEmpty {
                 date = nil
             } else {
-                date = DateFormatter.arcGISDateFormatter.date(from: element.value)
+                date = try? Date(element.value, strategy: Date.ParseStrategy.arcGISDateParseStrategy)
             }
         }
         .onChange(of: date) { newDate in
-            guard let currentDate = DateFormatter.arcGISDateFormatter.date(from: element.value),
+            guard let currentDate = try? Date(element.value, strategy: Date.ParseStrategy.arcGISDateParseStrategy),
                   newDate != currentDate else {
                 return
             }
@@ -208,12 +208,14 @@ struct DateTimeEntry: View {
     }
 }
 
-private extension DateFormatter {
-    static let arcGISDateFormatter: DateFormatter = {
-        let dateFormatter = DateFormatter()
-        dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss"
-        return dateFormatter
-    }()
+private extension Date.ParseStrategy {
+    /// A parse strategy for date/time strings with a yyyy-MM-dd'T'HH:mm:ss format.
+    static var arcGISDateParseStrategy: Self {
+        .fixed(
+            format: "\(year: .defaultDigits)-\(month: .defaultDigits)-\(day: .defaultDigits)T\(hour: .defaultDigits(clock: .twentyFourHour, hourCycle: .zeroBased)):\(minute: .defaultDigits):\(second: .defaultDigits)",
+            timeZone: .current
+        )
+    }
 }
 
 private extension String {
--- a/Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
+++ b/Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
@@ -60,9 +60,9 @@ struct DateTimeEntry: View {
         }
         .padding([.bottom], elementPadding)
         .onAppear {
-            if let date = featureForm?.feature.attributes[element.fieldName] as? Date {
-                self.date = date
-            }
+            let dateFormatter = DateFormatter()
+            dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSZ"
+            self.date = dateFormatter.date(from: element.value)
         }
         .onChange(of: date) { newDate in
             //TODO: add `required` property to API
--- a/Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
+++ b/Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
@@ -60,7 +60,11 @@ struct DateTimeEntry: View {
         }
         .padding([.bottom], elementPadding)
         .onAppear {
-            self.date = DateFormatter.arcGISDateFormatter.date(from: element.value)
+            if element.value.isEmpty {
+                date = nil
+            } else {
+                date = DateFormatter.arcGISDateFormatter.date(from: element.value)
+            }
         }
         .onChange(of: date) { newDate in
             //TODO: add `required` property to API
--- a/Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
+++ b/Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
@@ -67,7 +67,7 @@ struct DateTimeEntry: View {
             }
         }
         .onChange(of: date) { newDate in
-            guard let currentDate = try? Date(element.value, strategy: Date.ParseStrategy.arcGISDateParseStrategy),
+            guard let currentDate = try? Date(element.value, strategy: .arcGISDateParseStrategy),
                   newDate != currentDate else {
                 return
             }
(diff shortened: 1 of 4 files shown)


Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
```diff
@@ -60,9 +60,9 @@ struct DateTimeEntry: View {
         }
         .padding([.bottom], elementPadding)
         .onAppear {
-            if let date = featureForm?.feature.attributes[element.fieldName] as? Date {
-                self.date = date
-            }
+            let dateFormatter = DateFormatter()
+            dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSZ"
```
mhdostal, line 64: Is there any advantage to making `dateFormatter` a property?

UI Test Runner/UITests/FormViewTests.swift
```diff
@@ -306,12 +306,12 @@ final class FormViewTests: XCTestCase {
         
         XCTAssertEqual(
             footer.label,
-            "Required"
+            "Date Entry is Required"
```
dfeinzimer, line 309: This seems to have changed in the test data since Friday.

UI Test Runner/UITests/FormViewTests.swift
```diff
@@ -306,12 +306,12 @@ final class FormViewTests: XCTestCase {
...
         )
         
         XCTAssertTrue(
-            calendarImage.isHittable,
-            "The calendar image isn't hittable."
+            calendarImage.exists,
+            "The calendar image doesn't exist."
```
dfeinzimer, line 314: Hittable was working okay on Friday but seems to not be now. Checking for existence should be more stable.

Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
```diff
@@ -202,6 +204,13 @@ struct DateTimeEntry: View {
     }
 }
 
+private extension DateFormatter {
+    static let arcGISDateFormatter: DateFormatter = {
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss"
+        return dateFormatter
+    }()
+}
```
philium, line 213: Apple's [Data Formatting](https://developer.apple.com/documentation/foundation/data_formatting) document recommends only using formatters in Objective-C. In Swift, there are (mostly) equivalent formatting APIs that don't require creating a formatter directly. I have found one case where I could only do what I wanted using a formatter, but that is a rare exception. In this case, I think you could use [`Date.init(_:strategy:)`](https://developer.apple.com/documentation/foundation/date/3798897-init) to do the same thing without using a formatter.

Using the new formatter APIs involves a change of mindset. Before those new APIs came along, creating formatters to format data was ubiquitous. Now, we have to train ourselves not to reach for a formatter when needing to format data and instead, search for the replacement API that does the same thing, but in a more Swift-y way.

mhdostal (approved):

Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
mhdostal: @philium Good point, thanks for the reminder!

dfeinzimer: @philium I've not been able to find a usage of `Date.init(_:strategy:)` that'll work with the date strings we have.

The format of the dates we're parsing is: `1969-07-07T20:17:00`.

I've tried a number of combinations of modifiers on `Date.ISO8601FormatStyle`.  More specifically, I've not been able to find a suitable `Date.ISO8601FormatStyle.DateTimeSeparator` for the `T`.

dfeinzimer:
@mhdostal I added one more [commit](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/429/commits/14fe2ece66a3ef84ce03593192a2a5478cf837bd). I noticed we were calling `setAttributeValue` in the `onAppear`s, passing in the value that's already stored.

Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
philium: You need to create the formatter yourself using a fixed format:

```swift
let strategy: Date.ParseStrategy = .fixed(
    format: "\(year: .defaultDigits)-\(month: .defaultDigits)-\(day: .defaultDigits)T\(hour: .defaultDigits(clock: .twentyFourHour, hourCycle: .zeroBased)):\(minute: .defaultDigits):\(second: .defaultDigits)",
    timeZone: .current
)
try Date("1969-07-07T20:17:00", strategy: strategy) // 1969-07-08 03:17:00 +0000
```

dfeinzimer:
Applied the Date.ParseStrategy feedback, this is still ready for review.

Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
```diff
@@ -60,11 +60,17 @@ struct DateTimeEntry: View {
         }
         .padding([.bottom], elementPadding)
         .onAppear {
-            if let date = featureForm?.feature.attributes[element.fieldName] as? Date {
-                self.date = date
+            if element.value.isEmpty {
+                date = nil
+            } else {
+                date = try? Date(element.value, strategy: Date.ParseStrategy.arcGISDateParseStrategy)
```
mhdostal, line 66: See the changes to the extension as well...
```suggestion
                date = try? Date(element.value, strategy: .arcGISDateParseStrategy)
```

Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
```diff
@@ -60,11 +60,17 @@ struct DateTimeEntry: View {
...
             }
         }
         .onChange(of: date) { newDate in
+            guard let currentDate = try? Date(element.value, strategy: Date.ParseStrategy.arcGISDateParseStrategy),
```
mhdostal, line 70:
```suggestion
            guard let currentDate = try? Date(element.value, strategy: .arcGISDateParseStrategy),
```

Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeEntry.swift
```diff
@@ -202,6 +208,15 @@ struct DateTimeEntry: View {
     }
 }
 
+private extension Date.ParseStrategy {
```
mhdostal, line 211: The magic is here:

```suggestion
private extension ParseStrategy where Self == Date.ParseStrategy {
```

UI Test Runner/UITests/FormViewTests.swift
```diff
@@ -609,6 +613,16 @@ final class FormViewTests: XCTestCase {
...
 
+private extension Date.ParseStrategy {
+    /// A parse strategy for date/time strings with a yyyy-MM-dd'T'HH:mm:ss format.
+    static var arcGISDateParseStrategy: Self {
+        .fixed(
+            format: "\(year: .defaultDigits)-\(month: .defaultDigits)-\(day: .defaultDigits)T\(hour: .defaultDigits(clock: .twentyFourHour, hourCycle: .zeroBased)):\(minute: .defaultDigits):\(second: .defaultDigits)",
+            timeZone: .current
+        )
+    }
+}
```
mhdostal, line 624: I tried to remove this and use the one in the Toolkit, but I couldn't get it to compile. I made the toolkit one internal, added `@testable import ArcGISToolkit` and `import Foundation` to this file, but kept getting this error:
```
Undefined symbol: _$s10Foundation13ParseStrategyP13ArcGISToolkitAA4DateVABVRszrlE010arcGISDatebC0AGvgZ
```

mhdostal (changes requested):

UI Test Runner/UITests/FormViewTests.swift
dfeinzimer: Yes, I had tried this exact thing to prevent the repeated code myself and after a long build time it failed and I got the same error. You'd think it should be able to work.

mhdostal: 🤷

mhdostal (approved):

UI Test Runner/UITests/FormViewTests.swift
philium: This shouldn't be necessary. A test shouldn't try to duplicate exactly what the code being tested does. Correct me if I'm wrong, but the test is just using the strategy to compute the expected date from a string. Instead, just get the time interval of the calculated date and then replace the date calculation with a date created from that time interval. Or, if desired, create a date from date components. But unless I'm missing something, the test does not and should not be using a date parse strategy.

PR #430, opened 2023-09-13
fly-over

rolson:
- abstracted out an ArcGISSceneViewController which is a SceneView that can be used with UIKit
- some renames
- removed some duplicate properties
- cleaned up some Task code (used detached)
- created an augmented reality examples category

### DIFF
--- a/Sources/ArcGISToolkit/Components/AR/ARGeoView.swift
+++ b/Sources/ArcGISToolkit/Components/AR/ARGeoView.swift
@@ -23,14 +23,19 @@ public enum ARGeoViewTrackingMode {
 
 public struct ARGeoView: UIViewControllerRepresentable {
     private let scene: ArcGIS.Scene
-    private let renderVideoFeed: Bool
     private let cameraController: TransformationMatrixCameraController
     private let trackingMode: ARGeoViewTrackingMode
+    private let renderVideoFeed: Bool
+    private let graphicsOverlays: [GraphicsOverlay]
+    private let analysisOverlays: [AnalysisOverlay]
     
     public func makeUIViewController(context: Context) -> ViewController {
         let viewController = ViewController(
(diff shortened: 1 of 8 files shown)


rolson:
please hold off for a bit, I'm not happy with some of the renames.

rolson:
ok, this is ready for review now.

Sources/ArcGISToolkit/Components/AR/ARGeoView.swift
```diff
@@ -0,0 +1,384 @@
...
+        }
+        
+        var sceneViewController: ArcGISSceneViewController
+        
+        public init(
+            scene: ArcGIS.Scene,
+            cameraController: TransformationMatrixCameraController,
+            graphicsOverlays: [GraphicsOverlay] = [],
+            analysisOverlays: [AnalysisOverlay] = [],
+            renderVideoFeed: Bool = true
```
des12437, line 279:
```suggestion
            analysisOverlays: [AnalysisOverlay] = [],
            trackingMode: ARGeoViewTrackingMode = .ignore,
            renderVideoFeed: Bool
```
The `ARGeoView` initializer sets the tackingMode, so that value should be passed here as well. Also `renderVideoFeed` does not have a default argument in the `ARGeoView` initializer, so that default argument should be removed here.

des12437 (commented):
Looks good. There are more changes needed to fully support flyover, such as setting the space and atmosphere effect in the `ARGeoView` initializer, but those changes can be made in another PR.

Sources/ArcGISToolkit/Components/AR/ARGeoView.swift
rolson: parameters with defaults should be last.

des12437 (approved):

PR #432, opened 2023-09-15
Organize extensions

dfeinzimer:
Organizes extensions made in the Toolkit into ~3~ 4 folders:

Screenshot 2023-09-14 at 18 03 35

*Updated*


philium:
FYI: `String` is provided by the `Swift` module, not `Foundation`. So if you're sorting types by module, which it looks like you are, `String.swift` should go in a directory called `Swift`: `Extensions/Swift/String.swift`.

dfeinzimer:
> `String.swift` should go in a directory called `Swift`: `Extensions/Swift/String.swift`

Thanks Phil, I'll make that change. Would the same go for [Result.swift](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/v.next/Sources/ArcGISToolkit/Extensions/Result.swift)? I'm not sure but it seems like it might.

philium:
> Would the same go for [Result.swift](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/v.next/Sources/ArcGISToolkit/Extensions/Result.swift)?

Yes. [`Result`](https://developer.apple.com/documentation/swift/result) is also provided by the Swift standard library, so it is also part of the `Swift` module.

PR #435, opened 2023-09-19
`FormView` - Scroll to break focus

dfeinzimer:
Closes Apollo 161

This introduces behavior where scrolling the form will cause any single line or multi line text entry field to loose focus if it currently has it.

~I'm not entirely set on this method because it seems to introduce a very slight delay on when the `ScrollView`'s main gesture recognizer picks up the gesture, causing a very slight jump when scrolling first starts.~


dfeinzimer:
@mhdostal Let me know your thoughts on this. I'm not 100% sold on it for the reason in the description.

I figure we can either:
- merge as is and see what response it gets in the **FeatureFormsMicro** app
- merge with an added note about future optimization potential
- close this out and continue the issue for later

mhdostal:
Unfortunately, this is iOS 16-only: https://developer.apple.com/documentation/swiftui/view/scrolldismisseskeyboard(_:)/

dfeinzimer:
> Unfortunately, this is iOS 16-only: https://developer.apple.com/documentation/swiftui/view/scrolldismisseskeyboard(_:)/

Ah nice find. In this case would you be in support of using that modifier and making this behavior iOS 16.0+ only?

mhdostal:
This does solve the scrolling problem (with the slight delay you mention), but doesn't solve the problem in Apollo 161 where you can resize the floating panel to the point where it's completely behind the keyboard.

mhdostal:
> > Unfortunately, this is iOS 16-only: https://developer.apple.com/documentation/swiftui/view/scrolldismisseskeyboard(_:)/
> 
> Ah nice find. In this case would you be in support of using that modifier and making this behavior iOS 16.0+ only?

If it works to solve the scrolling issue, yes. I suspect we may need something different for the floating panel detent issue.

dfeinzimer:
@mhdostal this is ready for re-review. It now also modifies the `FloatingPanel` to trigger a keyboard close when the handle is dragged. This should resolve the issue of an orphaned keyboard blocking the form.

When approved I can cherry pick the the `FloatingPanel` changes into `v.next` to keep the `Forms` branch clean.

Examples/Examples/FormViewExampleView.swift
```diff
@@ -16,7 +16,8 @@ import ArcGISToolkit
 import SwiftUI
 
 struct FormViewExampleView: View {
-    @Environment(\.verticalSizeClass) var verticalSizeClass
```
dfeinzimer, line 19: This property is unused.

Sources/ArcGISToolkit/Components/FloatingPanel/FloatingPanel.swift
```diff
@@ -127,6 +131,14 @@ struct FloatingPanel<Content>: View where Content: View {
             .onChanged {
                 let deltaY = $0.location.y - (latestDragGesture?.location.y ?? $0.location.y)
                 let proposedHeight = height + ((isCompact ? -1 : +1) * deltaY)
+                
+                if shouldSendResign {
+                    if proposedHeight < height {
```
mhdostal, line 136: Should we _always_ resign first responder, not just when the user shortens the panel but anytime the handle is dragged? In practice, whenever I tried to get in a state where I could grab the handle and make it bigger, I couldn't as displaying the keyboard always maximized the height of the panel.

dfeinzimer: Thats a good point. Yes I think doing that should be fine.

mhdostal (approved):

PR #436, opened 2023-09-19
Fly over AR View and example

rolson:
This implements the AR fly over view (`FlyOverSceneView`) an an example.

You will need branch `ryan3***4/flyOver` of the SDK as well in order to build and test this.

I think flyover is pretty much complete based on the done criteria:

- there is a demo
- You can pinch to zoom (change scale)
- You can tap to identify

Note: Double tap to zoom does not work. My guess is this is because of the use of the TransformationMatrixCameraController.

Still todo, can be done later:
- Need to figure out preference bug
- We can easily switch to reality kit if we wanted

The approach taken in this PR is that the user passes in the scene view. This is so they can set it up exactly how they want to. If we didn't do this then we would need to duplicate the whole scene view API onto the FlyOverSceneView. 
When we get the scene view that the user passes us, we then override a few things (camera controller, view drawing mode, etc).

The user can perform operations (ie. identify, etc) because we give them access to the proxy through the scene view ViewBuilder closure.

Usage of `FlyOverSceneView`:

```swift
    var body: some View {
        FlyoverSceneView(
            initialCamera: initialCamera,
            translationFactor: 3_000
        ) { _ in
            SceneView(scene: scene)
        }
    }
```


Examples/Examples/FlyoverExampleView.swift
```diff
@@ -0,0 +1,50 @@
...
+        heading: 40,
+        pitch: 90,
+        roll: 0
+    )
+    
+    var body: some View {
+        FlyoverSceneView(
+            initialCamera: initialCamera,
+            translationFactor: 3_000,
+            clippingDistance: 60_000
```
des12437, line 38:
```suggestion
            translationFactor: 3_000
```
Clipping distance should not be required in the initializer as it's not needed for a Flyover workflow. However it may be good to support modifying or setting the clipping distance in a future iteration in case it's needed to support a user's specific workflow.

rolson: Great, thank you

Sources/ArcGISToolkit/Components/AR/FlyoverSceneView.swift
```diff
@@ -0,0 +1,140 @@
...
+                            orientation: lastGoodDeviceOrientation
+                        )
+                    }
+                    .onAppear {
+                        arViewProxy.session?.run(configuration)
+                    }
+                    .onDisappear {
+                        arViewProxy.session?.pause()
+                    }
+                    .opacity(0.0)
```
des12437, line 74:
```suggestion
                    .opacity(0)
```

des12437 (approved):
Looks good.

rolson:
Instead of opacity, I added a modifier to the ARSwiftUIView to hide the video feed

des12437:
The tap to identify and pan gestures are not working with these changes. I'm looking into the exact change that caused the issue.

rolson:
Correct. It is because the ARSCNView was shown over the SceneView, eating the touches. I've fixed it. Just needed to disable the AR view.

des12437 (dismissed):

des12437 (dismissed):

des12437:
Should `RealityKit.ARView` be integrated in place of `ARKit.ARSCNView` in a separate PR? If so, I can work on that once this PR is merged.

des12437 (approved):

pgruenler (approved):
LGTM

PR #438, opened 2023-09-21
Tabletop AR and example

des12437:
This PR implements tabletop AR by visualizing horizontal planes and performing hit testing when a user taps on a plane to place an ArcGIS scene.

Usage of `TableTopSceneView`:
``` swift
var body: some View {
       TableTopSceneView(
            anchorPoint: anchorPoint,
            translationFactor: 1_000,
            clippingDistance: 400
        ) { _ in
            SceneView(scene: scene)
        }
    }
```


Sources/ArcGISToolkit/Components/AR/ARSwiftUIView.swift
```diff
@@ -44,6 +46,22 @@ struct ARSwiftUIView {
         view.videoFeedIsHidden = true
         return view
     }
+    
+    func onAddNode(
```
rolson, line 50: needs doc

Sources/ArcGISToolkit/Components/AR/ARSwiftUIView.swift
```diff
@@ -44,6 +46,22 @@ struct ARSwiftUIView {
...
+        perform action: @escaping (SCNSceneRenderer, SCNNode, ARAnchor) -> Void
+    ) -> Self {
+        var view = self
+        view.onAddNodeAction = action
+        return view
+    }
+    
+    func onUpdateNode(
```
rolson, line 58: needs doc

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
```diff
@@ -0,0 +1,241 @@
...
+    /// Updates the last good device orientation.
+    func updateLastGoodDeviceOrientation() {
+        // Get the device orientation, but don't allow non-landscape/portrait values.
+        let deviceOrientation = UIDevice.current.orientation
+        if deviceOrientation.isValidInterfaceOrientation {
+            lastGoodDeviceOrientation = deviceOrientation
+        }
+    }
+    
+    func visualizePlane(_ renderer: SCNSceneRenderer, didAdd node: SCNNode, for anchor: ARAnchor) {
```
rolson, line 127: needs doc

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
```diff
@@ -0,0 +1,241 @@
...
+        
+        // Create a custom object to visualize the plane geometry and extent.
+        let plane = Plane(anchor: planeAnchor)
+        
+        // Add the visualization to the ARKit-managed node so that it tracks
+        // changes in the plane anchor as plane estimation continues.
+        node.addChildNode(plane)
+    }
+    
+    func updatePlane(_ renderer: SCNSceneRenderer, didUpdate node: SCNNode, for anchor: ARAnchor) {
```
rolson, line 139: needs doc

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
```diff
@@ -0,0 +1,241 @@
...
+        
+        // `SCNPlane` is vertically oriented in its local coordinate space, so
+        // rotate it to match the orientation of `ARPlaneAnchor`.
+        node.eulerAngles.x = -.pi / 2
+
+        super.init()
+
+        node.opacity = 0.25
+        guard let material = node.geometry?.firstMaterial
+            else { fatalError("SCNPlane always has one material") }
```
rolson, line 175: if there is always a geometry and always a first material then it should be force unwrapped (`!`). If there are cases where geometry or firstMaterial are `nil` then it should be `guard let else { return `nil` }` and make the init failable.

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
```diff
@@ -0,0 +1,241 @@
...
+        fatalError("init(coder:) has not been implemented")
+    }
+}
+
+private extension SceneViewProxy {
+    /// Sets the initial transformation used to offset the originCamera.  The initial transformation is based on an AR point determined via existing plane hit detection from `screenPoint`.  If an AR point cannot be determined, this method will return `false`.
+    ///
+    /// - Parameter screenPoint: The screen point to determine the `initialTransformation` from.
+    /// - Returns: The `initialTransformation`.
+    /// - Since: 200.3
```
rolson, line 193: not a public method, don't need since
```suggestion
```

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
```diff
@@ -0,0 +1,241 @@
...
+    func setInitialTransformation(
```
rolson, line 194: This doesn't set the initial transformation. It returns an initial transformation. Therefore it should be named as such. Maybe something like `initialTransformation(for:...)`

Also it doesn't seem to require the SceneViewProxy, so this method should extend SceneViewProxy. Maybe it should extend ARViewProxy.

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
```diff
@@ -0,0 +1,241 @@
...
+        let initialTransformation = TransformationMatrix.identity.subtracting(matrix)
+
+        return initialTransformation
+    }
+    
+    /// Internal method to perform a hit test operation to get the transformation matrix representing the corresponding real-world point for `screenPoint`.
+    ///
+    /// - Parameter screenPoint: The screen point to determine the real world transformation matrix from.
+    /// - Returns: An `TransformationMatrix` representing the real-world point corresponding to `screenPoint`.
+    func internalHitTest(using screenPoint: CGPoint, for arViewProxy: ARSwiftUIViewProxy) -> TransformationMatrix? {
```
rolson, line 212: I wouldn't include the prefix "internal" unless it's needed. I don't think it is in this case. 

Also this doesn't seem to need to extend SceneViewProxy. It might make more sense extending ARSwiftUIProxy.

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
```diff
@@ -0,0 +1,241 @@
...
+    /// The last portrait or landscape orientation value.
+    @State private var lastGoodDeviceOrientation = UIDeviceOrientation.portrait
+    @State private var arViewProxy = ARSwiftUIViewProxy()
+    @State private var sceneViewProxy: SceneViewProxy?
+    @State private var initialTransformation: TransformationMatrix? = nil
+    
+    private let cameraController: TransformationMatrixCameraController
+    private let sceneViewBuilder: (SceneViewProxy) -> SceneView
+    private let configuration: ARWorldTrackingConfiguration
+    private var didSetTransforamtion: Bool {
```
rolson, line 29: it seems like this should be named `initialTransformationIsSet`

rolson (commented):
Maybe not for the initial PR, but soon after I think we need a way to reshow the planes and hide the scene view. This could be done with a view modifier potentially.

Examples/Examples/TableTopExampleView.swift
```diff
@@ -0,0 +1,61 @@
...
+        scene.baseSurface.navigationConstraint = .unconstrained
+        scene.baseSurface.opacity = 0
+        
+        return scene
+    }()
+    
+    private let anchorPoint = Point(
+        x: -122.68350326165559,
+        y: 45.53257485106716,
+        z: 10.189046687446535,
```
rolson, line 41:
```suggestion
```

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
```diff
@@ -0,0 +1,249 @@
...
+                .onUpdateNode { _, node, anchor in
+                    updatePlane(with: node, for: anchor)
+                }
+                .onAppear {
+                    arViewProxy.session?.run(configuration)
+                }
+                .onDisappear {
+                    arViewProxy.session?.pause()
+                }
+                .gesture(
```
rolson, line 88: shouldn't this just be a tap gesture?

```swift

                .onTapGesture {
                    ...
                }
```

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
```diff
@@ -0,0 +1,249 @@
...
+    private let cameraController: TransformationMatrixCameraController
+    private let sceneViewBuilder: (SceneViewProxy) -> SceneView
+    private let configuration: ARWorldTrackingConfiguration
+    private var initialTransformationIsSet: Bool {
+        initialTransformation != nil
+    }
+    
+    /// Creates a fly over scene view.
+    /// - Parameters:
+    ///   - anchorPoint: The anchor point of the ArcGIS Scene.
```
rolson, line 35: It might be good to explain "anchor point" a bit.

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
des12437: `onTapGesture()` does not provide the screen point in iOS 15 and the screen point is needed for hit testing.

rolson: So I wonder if we write our own onTap that does provide the screen point and internally use the `if available` and if not iOS 16 then in that place use the DragGesture. Then eventually the workaround can be removed. [I did that here](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/v.next/Sources/ArcGISToolkit/Extensions/SwiftUI/View.swift#L62-L71) for presentation detents.

Examples/Examples/TableTopExampleView.swift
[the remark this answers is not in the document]
rolson: or not pass the param I mean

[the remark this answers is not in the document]
des12437: The z value does not need to be specified here. Will remove.

pgruenler (changes requested):

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
des12437: I added a `onSingleTapGesture()` modifier similar to the one on `MapView`/`SceneView` which uses UIKit to detect the single tap rather than SwiftUI.

des12437:
I based this branch on the changes in #443 since both PRs modify the `draw`/`updateCamera` method.

Sources/ArcGISToolkit/Components/AR/ARSwiftUIView.swift
```diff
@@ -0,0 +1,165 @@
...
+        view.onSingleTapGesture = action
+        return view
+    }
+    
+    /// Hides the video feed for the AR view.
+    func videoFeedHidden() -> Self {
+        var view = self
+        view.videoFeedIsHidden = true
+        return view
+    }
```
rolson, line 79: You can remove this if not used.

```suggestion
```

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
```diff
@@ -0,0 +1,265 @@
...
+        configuration.planeDetection = [.horizontal]
+    }
+    
+    public var body: some View {
+        ZStack {
+            ARSwiftUIView(proxy: arViewProxy)
+                .onDidUpdateFrame { _, frame in
+                    guard let sceneViewProxy else { return }
+                    updateLastGoodDeviceOrientation()
+                    sceneViewProxy.updateCamera(
```
rolson, line 67: Since `sceneViewProxy.updateCamera` is now used outside of just the flyoverSceneView, we should move that extension into it's own file.

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
```diff
@@ -0,0 +1,265 @@
...
+                }
+                .onAddNode { _, node, anchor in
+                    addPlane(with: node, for: anchor)
+                }
+                .onUpdateNode { _, node, anchor in
+                    updatePlane(with: node, for: anchor)
+                }
+                .onSingleTapGesture { screenPoint in
+                    guard let sceneViewProxy,
+                          !initialTransformationIsSet else { return }
```
rolson, line 86: This is purely stylistic, so take it or leave it. When the guard statement is multiple lines I prefer the else on a separate line.  Same suggestion applies to other multi-line guard statements below.

```suggestion
                    guard let sceneViewProxy,
                          !initialTransformationIsSet
                    else { return }
```

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
```diff
@@ -0,0 +1,265 @@
...
+            translationX: Double(worldTransform.columns.3.x),
+            translationY: Double(worldTransform.columns.3.y),
+            translationZ: Double(worldTransform.columns.3.z)
+        )
+        
+        return hitTestMatrix
+    }
+}
+
+private extension SceneViewProxy {
```
rolson, line 226: if we will need this for world scale then I would move it into the same file as the updateCamera extension method. Or that can be done when world scale needs it. It might only need the setFieldOfView method

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
```diff
@@ -0,0 +1,265 @@
...
+        if let extentGeometry = plane.node.geometry as? SCNPlane {
+            extentGeometry.width = CGFloat(planeAnchor.extent.x)
+            extentGeometry.height = CGFloat(planeAnchor.extent.z)
+            plane.node.simdPosition = planeAnchor.center
+        }
+    }
+}
+
+/// A helper class to visualize a plane found by ARKit.
+class Plane: SCNNode {
```
rolson, line 162:
```suggestion
private class Plane: SCNNode {
```

Sources/ArcGISToolkit/Components/AR/ARSwiftUIView.swift
```diff
@@ -0,0 +1,165 @@
...
+    /// Sets the closure to call when a new node has been added to the scene.
+    func onAddNode(
+        perform action: @escaping (SCNSceneRenderer, SCNNode, ARAnchor) -> Void
+    ) -> Self {
+        var view = self
+        view.onAddNodeAction = action
+        return view
+    }
+    
+    /// Sets the closure to call when the scene's nodes are update.
```
rolson, line 56:
```suggestion
    /// Sets the closure to call when the scene's nodes are updated.
```

Sources/ArcGISToolkit/Components/AR/ARSwiftUIView.swift
```diff
@@ -0,0 +1,165 @@
...
+    }
+    
+    /// Sets a closure to perform when a single tap occurs on the AR view.
+    func onSingleTapGesture(
+        perform action: @escaping (CGPoint) -> Void
+    ) -> Self {
+        var view = self
+        view.onSingleTapGesture = action
+        return view
+    }
```
rolson, line 72: This is fine, but it's not ideal because we won't be reminded to remove it when we no longer support iOS 16. I would prefer a general `View` extension like the presentation detents one.

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
des12437: Agreed, however theses errors occur when moving `updateCamera()` into a separate file:
Screenshot 2023-09-26 at 11 51 59 AM


Screenshot 2023-09-26 at 11 53 22 AM

rolson: I'm not seeing that. Can you try branch `ryan/tabletop2` please?

des12437: Yes, is that branch published?

des12437: The code compiles without errors on `ryan/tabletop2`. Thanks.

rolson: I'm not sure what the difference was. Maybe an import statement?

des12437: No it was because I placed the file under the components folder rather than the extensions folder.

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
des12437: I think only `setFieldOfView()` is needed for world scale, but the method can be moved later if necessary.

Sources/ArcGISToolkit/Components/AR/ARSwiftUIView.swift
```diff
@@ -0,0 +1,132 @@
...
+import SwiftUI
+
+typealias ARViewType = ARSCNView
+
+/// A SwiftUI version of an AR view.
+struct ARSwiftUIView {
+    /// The closure to call when the AR view renders.
+    private(set) var onDidUpdateFrameAction: ((ARSession, ARFrame) -> Void)?
+    private(set) var onAddNodeAction: ((SCNSceneRenderer, SCNNode, ARAnchor) -> Void)?
+    private(set) var onUpdateNodeAction: ((SCNSceneRenderer, SCNNode, ARAnchor) -> Void)?
```
rolson, line 24: can you doc these please?

Sources/ArcGISToolkit/Components/AR/ARSwiftUIView.swift
```diff
@@ -0,0 +1,132 @@
...
+
+import ARKit
+import SwiftUI
+
+typealias ARViewType = ARSCNView
+
+/// A SwiftUI version of an AR view.
+struct ARSwiftUIView {
+    /// The closure to call when the AR view renders.
+    private(set) var onDidUpdateFrameAction: ((ARSession, ARFrame) -> Void)?
```
rolson, line 22: I guess the doc here is out of date now that we chose a new delegate method.

Sources/ArcGISToolkit/Components/AR/ARSwiftUIView.swift
```diff
@@ -0,0 +1,132 @@
...
+        return view
+    }
+}
+
+extension ARSwiftUIView: UIViewRepresentable {
+    func makeUIView(context: Context) -> ARViewType {
+        let arView = ARViewType()
+        arView.delegate = context.coordinator
+        arView.session.delegate = context.coordinator
+        proxy?.arView = arView
```
rolson, line 69:
```suggestion
        // Set the AR view on the proxy.
        proxy?.arView = arView
```

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
```diff
@@ -0,0 +1,288 @@
...
+        if deviceOrientation.isValidInterfaceOrientation {
+            lastGoodDeviceOrientation = deviceOrientation
+        }
+    }
+    
+    /// Visualizes a new node added to the scene as an AR Plane.
+    /// - Parameters:
+    ///   - node: The node to be added to the scene.
+    ///   - anchor: The anchor position of the node.
+    func addPlane(with node: SCNNode, for anchor: ARAnchor) {
```
rolson, line 130: can this be private?

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
```diff
@@ -0,0 +1,288 @@
...
+        // Add the visualization to the ARKit-managed node so that it tracks
+        // changes in the plane anchor as plane estimation continues.
+        node.addChildNode(plane)
+    }
+    
+    /// Visualizes a node updated in scene as an AR Plane.
+    /// - Parameters:
+    ///   - node: The node to be updated in the scene.
+    ///   - anchor: The anchor position of the node.
+    func updatePlane(with node: SCNNode, for anchor: ARAnchor) {
```
rolson, line 145: can this be private?

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
```diff
@@ -0,0 +1,288 @@
...
+                        action(screenPoint.location)
+                    }
+            )
+        }
+    }
+}
+
+/// A helper class to visualize a plane found by ARKit.
+private class Plane: SCNNode {
+    let node: SCNNode
```
rolson, line 186: can you doc this please?

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
```diff
@@ -0,0 +1,288 @@
...
+    
+    init?(anchor: ARPlaneAnchor) {
```
rolson, line 188: can you doc this please?

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
```diff
@@ -0,0 +1,288 @@
...
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+}
+
+private extension ARSwiftUIViewProxy {
+    /// Performs a hit test operation to get the transformation matrix representing the corresponding real-world point for `screenPoint`.
+    /// - Parameter screenPoint: The screen point to determine the real world transformation matrix from.
+    /// - Returns: A `TransformationMatrix` representing the real-world point corresponding to `screenPoint`.
+    func hitTest(using screenPoint: CGPoint) -> TransformationMatrix? {
```
rolson, line 218: "at" might be more descriptive in this case than "using".

```suggestion
    func hitTest(at screenPoint: CGPoint) -> TransformationMatrix? {
```

rolson (commented):
Looking good. Couple minor comments.

pgruenler (commented):
Looks good. Just left on comment.

rolson (approved):

pgruenler (approved):

des12437:
Thank you both.

PR #439, opened 2023-09-22
Flyover: update delegate method

rolson:
Use new delegate method to update the transformation matrix. 

2 reasons:
This allows us to switch from reality kit to ARKit easily with this code:

```swift
//typealias ARViewType = RealityKit.ARView
typealias ARViewType = ARSCNView
```

And it also allows our draw method to rely on frame state that is passed directly to the delegate method instead of querying it from the session.

### DIFF
--- a/Sources/ArcGISToolkit/Components/AR/FlyoverSceneView.swift
+++ b/Sources/ArcGISToolkit/Components/AR/FlyoverSceneView.swift
@@ -101,7 +101,7 @@ extension SceneViewProxy {
     ) {
         let cameraTransform = frame.camera.transform
         
-        // Rotate camera transform 90 degrees in the XY plane.
+        // Rotate camera transform 90 degrees counter-clockwise in the XY plane.
         let transform = simd_float4x4.init(
             cameraTransform.columns.1,
             -cameraTransform.columns.0,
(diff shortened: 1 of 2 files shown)


des12437 (dismissed):
Looks good.

rolson:
The transform needed to be rotated for the interface orientation. I've fixed that. Please review

Sources/ArcGISToolkit/Components/AR/FlyoverSceneView.swift
```diff
@@ -88,54 +87,84 @@ public struct FlyoverSceneView: View {
...
+
+private extension ARCamera {
+    /// The transform rotated for a particular device orientation.
+    /// - Parameter orientation: The device orientation that the transform is appropriate for.
+    /// - Precondition: 'orientation.isValidInterfaceOrientation'
+    func transform(for orientation: UIDeviceOrientation) -> simd_float4x4 {
+        precondition(orientation.isValidInterfaceOrientation)
+        switch orientation {
+        case .portrait:
+            // Rotate camera transform 90 degrees counter-clockwise in the XY plane.
```
des12437, line 143: `landscapeRight` and `portraitUpsideDown` could have similar documentation.

des12437 (approved):
Rendering looks correct for each type of device orientation.

rolson:
closing in favor of #443

PR #440, opened 2023-09-22
Flyover: use auto (pulse) draw instead of manual drawing

rolson:
I'm not sure why we are manually drawing and why we had that requirement before (@mhdostal?).  But all we are doing is updating the transformation matrix camera controller and the field of view. The scene view's normal pulse/draw should work fine and it does in my testing.

### DIFF
--- a/Sources/ArcGISToolkit/Components/AR/FlyoverSceneView.swift
+++ b/Sources/ArcGISToolkit/Components/AR/FlyoverSceneView.swift
@@ -54,11 +54,10 @@ public struct FlyoverSceneView: View {
             SceneViewReader { sceneViewProxy in
                 sceneViewBuilder(sceneViewProxy)
                     .cameraController(cameraController)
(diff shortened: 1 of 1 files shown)


des12437 (approved):
The rendering looks correct with automatic drawing, although it would be good to know why manual drawing was used previously.

pgruenler:
If we are drawing on the pulse could we also update the camera on the pulse? If I'm understanding this right, we are updating the camera every frame and using it every pulse. It seems like we should do both either on the pulse or for every frame.

rolson:
The pulse is more like a "figure out what changed". So I think simply updating the camera controller will allow the pulse to do its job. Also, they are both happening on the same thread (main), so the camera controller will change and then pulse will be called in the same context.

pgruenler (approved):

rolson:
closing in favor of #443

PR #441, opened 2023-09-22
`FormView` - Add Radio Button support

dfeinzimer:
- Adds Radio Button support to FormView and 1 associated test case.
  - Adds an initializer to `ComboBoxInput` to support fallback initialization from a `RadioButtonInput`.


Sources/ArcGISToolkit/Components/Form/FormInputs/RadioButtonsInput.swift
```diff
@@ -0,0 +1,170 @@
...
+                    if input.noValueOption == .show {
+                        makeRadioButtonRow(
+                            placeholderValue,
+                            selectedValue == nil,
+                            !codedValues.isEmpty,
+                            useNoValueStyle: true
+                        ) {
+                            selectedValue = nil
+                        }
+                    }
```
dfeinzimer, line 78: The design states that if the field "Doesn't have a value, and this field is not required" then the no value option is shown. 

What's unclear is if the no value option hides itself once the field does have a value. It seems that it would be better to always show the no value option as long as `noValueOption == .show`.

Sources/ArcGISToolkit/Components/Form/FormInputs/RadioButtonsInput.swift
```diff
@@ -0,0 +1,170 @@
...
+    
+    /// A Boolean value indicating whether the date selection was cleared when a value is required.
+    @State private var requiredValueMissing = false
+    
+    /// The selected option.
+    @State private var selectedValue: CodedValue?
+    
+    /// A Boolean value indicating whether the current value doesn't exist as an option in the domain.
+    ///
+    /// In this scenario a ``ComboBoxInput`` should be used instead.
```
mhdostal, line 32:
```suggestion
    /// A Boolean value indicating whether a `ComboBoxInput`` should be used instead. This will be `true` if
    /// the current value doesn't exist as an option in the domain
```

Sources/ArcGISToolkit/Components/Form/FormInputs/RadioButtonsInput.swift
```diff
@@ -0,0 +1,170 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ArcGIS
+import SwiftUI
+
+struct RadioButtonsInput: View {
```
mhdostal, line 18: Needs doc.

mhdostal (commented):
Just a couple of comments

Sources/ArcGISToolkit/Components/Form/FormInputs/RadioButtonsInput.swift
dfeinzimer: Added

mhdostal (approved):

PR #442, opened 2023-09-22
`FormView` - General refactoring

dfeinzimer:
- Removes `FieldFormElement.isRequired`. This was a temporary extension and is no longer needed.
- Removes unneeded comments and cleans up the `FormView` example. Renames the example view itself.


Examples/Examples/FormViewExampleView.swift
```diff
@@ -112,17 +109,51 @@ extension FormExampleView {
     }
 }
 
-private extension FormExampleView {
-    /// A Boolean value indicating whether the form controls should be shown directly in the form's presenting container.
-    var useControlsInForm: Bool {
-        verticalSizeClass == .compact ||
-        UIDevice.current.userInterfaceIdiom == .mac ||
-        UIDevice.current.userInterfaceIdiom == .pad
-    }
```
dfeinzimer, line 121: For purposes of the example, showing the Cancel and Submit buttons within the toolbar should be fine.

Examples/ExamplesApp/Examples.swift
```diff
@@ -39,7 +39,7 @@ extension ExampleList {
             AnyExample("Bookmarks", content: BookmarksExampleView()),
             AnyExample("Compass", content: CompassExampleView()),
             AnyExample("Floor Filter", content: FloorFilterExampleView()),
-            AnyExample("Form", content: FormExampleView()),
+            AnyExample("FormView", content: FormViewExampleView()),
```
dfeinzimer, line 42: Seems like it'd be more correct to use the component's symbol name here (no space).

Should the other entries be changed to match? I can take care of that separately on `v.next`.

mhdostal (approved):
This looks good. Let's wait on renaming the other view files/labels. Maybe create an issue for it.

PR #443, opened 2023-09-23
Fly over: do not use ARView

rolson:
This PR will supersede the other 2 fly over PRs (#439, #440). I will close those.

This has a few notable changes:

- The ARView is not used, goes straight to ARSession 
- I have removed ARSwiftUIView, @des12437  can add it back in for tabletop.
- The view drawing mode is automatic (uses pulse)
- The pitch is changed so that it's more natural
- The FOV is not set not set on the scene view because it does not need to be since we are not showing the camera feed


Sources/ArcGISToolkit/Components/AR/FlyoverSceneView.swift
```diff
@@ -44,35 +45,22 @@ public struct FlyoverSceneView: View {
...
-                    }
-            }
+        SceneViewReader { sceneViewProxy in
+            sceneViewBuilder(sceneViewProxy)
+                .cameraController(cameraController)
+                .onAppear { session.start() }
+                .onDisappear { session.pause() }
+                .onChange(of: session.currentFrame) { frame in
+                    guard let frame else { return }
+                    sceneViewProxy.updateCamera(
```
des12437, line 58:
```suggestion
                    guard let frame else { return }
                    updateLastGoodDeviceOrientation()
                    sceneViewProxy.updateCamera(
```
This line was removed so the device orientation was not updating.

Sources/ArcGISToolkit/Components/AR/FlyoverSceneView.swift
```diff
@@ -86,55 +74,104 @@ public struct FlyoverSceneView: View {
...
+    func pause() {
+        session.pause()
+    }
+    
+    /// The latest AR frame.
+    @Published
+    var currentFrame: ARFrame?
+    
+    func session(_ session: ARSession, didUpdate frame: ARFrame) {
+        currentFrame = frame
```
des12437, line 107: The session captures video at a rate of 60 frames per second. Lowering the frame rate could improve the video quality by improving overall performance. The value `session.configuration?.videoFormat.framesPerSecond` is get-only. Maybe the value can be overridden or the [preferredFramesPerSecond](https://developer.apple.com/documentation/metalkit/mtkview/1536027-preferredframespersecond) could be specified on the metal view. When presenting the ARSCNView the frames can be set directly with SCNView [preferredFramesPerSecond](https://developer.apple.com/documentation/scenekit/scnview/1621205-preferredframespersecond). This change could be addressed in a separate PR.

des12437 (changes requested):

Sources/ArcGISToolkit/Components/AR/FlyoverSceneView.swift
rolson: I agree, not for this PR. Also might not be something we'd want to change. We'd need to see.

rolson:
Thank you both!

PR #445, opened 2023-09-25
Fix combine imports

dfeinzimer:
Fix Combine import errors as reported by @zkline101 

> JobManager.swift:88:26: warning: 'ObservableObject' aliases 'Combine.ObservableObject' and cannot be used here because 'Combine' was not imported by this file; this is an error in Swift 6

> LocatorSearchSource.swift:19:35: warning: 'ObservableObject' aliases 'Combine.ObservableObject' and cannot be used here because 'Combine' was not imported by this file; this is an error in Swift 6

### DIFF
--- a/Sources/ArcGISToolkit/Components/JobManager/JobManager.swift
+++ b/Sources/ArcGISToolkit/Components/JobManager/JobManager.swift
@@ -13,6 +13,7 @@
 
 import ArcGIS
 import BackgroundTasks
+import Combine
 import Foundation
 import OSLog
 import SwiftUI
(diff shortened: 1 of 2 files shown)


zkline101 (approved):

Sources/ArcGISToolkit/Components/JobManager/JobManager.swift
```diff
@@ -13,6 +13,7 @@
 
 import ArcGIS
 import BackgroundTasks
+import Combine
 import Foundation
 import OSLog
 import SwiftUI
```
philium, line 19: Is this still needed? Im guessing it was only used to get at `ObservableObject`. Now that `Combine` is imported, `SwiftUI` probably doesn't need to be.

dfeinzimer: I tried removing the SwiftUI import and tried a build which failed with the following errors. Should we be importing `UIKit` instead of `SwiftUI`?

Screenshot 2023-09-26 at 10 26 12

PR #448, opened 2023-09-27
AR: Improve TableTop AR plane visulaization

des12437:
This PR improves plane visualization in TableTop AR by creating a plane with convex polygon geometry which is more accurate since it uses all enclosing points with [ARSCNPlaneGeometry](https://developer.apple.com/documentation/arkit/arscnplanegeometry). The previous implementation visualizes a plane with less accuracy since the plane geometry is estimated based on the plane anchor's extent and center.

### DIFF
--- a/Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
+++ b/Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
@@ -76,7 +76,7 @@ public struct TableTopSceneView: View {
                     )
                 }
                 .onAddNode { renderer, node, anchor in
-                    addPlane(renderer: renderer, node: node, for: anchor)
+                    addPlane(renderer: renderer, node: node, anchor: anchor)
                 }
                 .onUpdateNode { _, node, anchor in
                     updatePlane(with: node, for: anchor)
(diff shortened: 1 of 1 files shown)


Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
```diff
@@ -125,38 +125,48 @@ public struct TableTopSceneView: View {
     
     /// Visualizes a new node added to the scene as an AR Plane.
     /// - Parameters:
+    ///   - renderer: The renderer for the scene.
     ///   - node: The node to be added to the scene.
     ///   - anchor: The anchor position of the node.
-    private func addPlane(with node: SCNNode, for anchor: ARAnchor) {
-        // Place content only for anchors found by plane detection.
+    private func addPlane(renderer: SCNSceneRenderer, node: SCNNode, for anchor: ARAnchor) {
```
rolson, line 131: Since the method doesn't read like a sentence we shouldn't use the proposition for the final argument. I didn't notice this last review.

```suggestion
    private func addPlane(renderer: SCNSceneRenderer, node: SCNNode, anchor: ARAnchor) {
```

rolson (approved):

pgruenler (approved):

des12437:
Thank you both.

PR #452, opened 2023-09-30
`FormView` - Add support for numeric text fields

dfeinzimer:
Adds support for numeric support in text inputs.


Sources/ArcGISToolkit/Components/Form/FormInputs/SingleLineTextInput.swift
```diff
@@ -77,10 +89,77 @@ struct SingleLineTextInput: View {
...
+                text.insert("-", at: text.startIndex)
+            }
+        } label: {
+            Image(systemName: "plus.forwardslash.minus")
+        }
+    }
+    
+    /// The range of valid values for a numeric input field.
+    var rangeDomain: RangeDomain? {
+        if let field = featureForm?.feature.table?.field(named: element.fieldName) {
```
dfeinzimer, line 147: As discussed in standup, the need for this should drop-off soon, perhaps before this is merged.

Sources/ArcGISToolkit/Components/Form/FormInputs/SingleLineTextInput.swift
```diff
@@ -62,10 +71,13 @@ struct SingleLineTextInput: View {
         }
         .formTextInputStyle()
         TextInputFooter(
-            currentLength: text.count,
+            text: text,
             isFocused: isFocused,
             element: element,
-            input: input
+            input: input,
+            rangeDomain: rangeDomain,
```
mhdostal, line 78: This can go away once we have the range domain on the element (`FieldFormElement.domain`).

Sources/ArcGISToolkit/Components/Form/FormInputs/SingleLineTextInput.swift
```diff
@@ -62,10 +71,13 @@ struct SingleLineTextInput: View {
...
+            isNumeric: isNumeric,
+            isDecimal: isDecimal
```
mhdostal, line 80: Is there a better way to do this instead of passing in both values? What about moving the `FieldType` extension containing the `isNumeric` and `isDecimal` properties to its own file and just passing in the field type to the footer? That would eliminate one argument. Thoughts?

Sources/ArcGISToolkit/Components/Form/FormInputs/SingleLineTextInput.swift
```diff
@@ -77,10 +89,77 @@ struct SingleLineTextInput: View {
...
+    var rangeDomain: RangeDomain? {
+        if let field = featureForm?.feature.table?.field(named: element.fieldName) {
+            return field.domain as? RangeDomain
+        } else {
+            return nil
+        }
+    }
+}
+
+private extension FieldType {
```
mhdostal, line 155: If we end up using this in more than one file, move it to it's own file in the Extension folder.

Sources/ArcGISToolkit/Components/Form/FormInputs/MultiLineTextInput.swift
```diff
@@ -112,3 +113,10 @@ struct MultiLineTextInput: View {
         }
     }
 }
+
+private extension MultiLineTextInput {
+    /// The field type of the text input.
+    var fieldType: FieldType {
+        featureForm!.feature.table!.field(named: element.fieldName)!.type!
+    }
+}
```
dfeinzimer, line 122: As just mentioned on Slack, seems it'd be better to have this as a property on `FieldFormElement` rather than having to resolve it at the input level manually with forced optionals. Note that it's also added in `SingleLineTextInput` for now.

mhdostal (approved):

PR #454, opened 2023-10-02
FlyOver: Use ARPositionalTrackingConfiguration

rolson:
According to apple doc:

> Enables 6 degrees of freedom tracking of the iOS device by running the camera at lowest possible resolution and frame rate. Use this configuration when you don't need to parse the camera feed, such as for example, virtual reality scenarios.

Switching to this should be more efficient. Trying it on the device, it seems to perform well and may even reduce the jumpiness.


des12437 (approved):
The jumpiness looks improved and the thermal state is improved as well in the Xcode energy report.

pgruenler (approved):

PR #457, opened 2023-10-03
Flyover AR: Do not allow user to specify pitch or roll

rolson:
- The pitch must be 90 and the roll must be 0 for flyover to work correctly. As such we shouldn't let the user specify those parameters.
- The initial camera needs to respond to changes, this fixes that and allows the initial camera and translation factor to be updated

### DIFF
--- a/Sources/ArcGISToolkit/Components/AR/FlyoverSceneView.swift
+++ b/Sources/ArcGISToolkit/Components/AR/FlyoverSceneView.swift
@@ -19,16 +19,16 @@ import ArcGIS
 public struct FlyoverSceneView: View {
     /// The AR session.
     @StateObject private var session = ObservableARSession()
+    /// The initial camera.
+    private let initialCamera: Camera
+    /// The translation factor.
+    private let translationFactor: Double
     /// The closure that builds the scene view.
     private let sceneViewBuilder: (SceneViewProxy) -> SceneView
     /// The camera controller that we will set on the scene view.
     @State private var cameraController: TransformationMatrixCameraController
     /// The current interface orientation.
     @State private var interfaceOrientation: InterfaceOrientation?
-    /// The initial camera.
-    let initialCamera: Camera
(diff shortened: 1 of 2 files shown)


Sources/ArcGISToolkit/Components/AR/FlyoverSceneView.swift
```diff
@@ -61,9 +118,20 @@ public struct FlyoverSceneView: View {
                         orientation: interfaceOrientation
                     )
                 }
+                .onChange(of: initialCamera) { initialCamera in
+                    cameraController.originCamera = initialCamera
+                }
+                .onChange(of: translationFactor) { translationFactor in
+                    cameraController.translationFactor = translationFactor
+                }
```
rolson, line 126: @des12437 can you look at if we need similar logic for the tabletop scene view? You can try adding a button that updates the anchor point and/or translation factor and see if the changes are propagated. I assume they won't be and similar logic will be required.

Sources/ArcGISToolkit/Components/AR/FlyoverSceneView.swift
```diff
@@ -61,9 +118,20 @@ public struct FlyoverSceneView: View {
...
                 .observingInterfaceOrientation($interfaceOrientation)
         }
     }
+    
+    func updateCameraController() {
+        cameraController = TransformationMatrixCameraController(originCamera: initialCamera)
+        cameraController.translationFactor = translationFactor
+    }
```
des12437, line 134: Is this method needed or used anywhere? It seems `.onChange(of: initialCamera)` and `.onChange(of: translationFactor)` update those properties already.

rolson: no, it's not needed. I left it in by accident.

des12437 (approved):

pgruenler (approved):

PR #460, opened 2023-10-06
TableTop AR: Add coaching overlay

des12437:
This PR adds a coaching overlay which replaces most help text prompts. The coaching overlay and single help text prompt "Tap a surface to place the scene" can be hidden with a view modifier.

### DIFF
--- a/Sources/ArcGISToolkit/Components/AR/ARSwiftUIView.swift
+++ b/Sources/ArcGISToolkit/Components/AR/ARSwiftUIView.swift
@@ -122,14 +122,14 @@ extension ARSwiftUIView {
 }
 
 /// A proxy for the ARSwiftUIView.
-class ARSwiftUIViewProxy {
+class ARSwiftUIViewProxy: NSObject, ARSessionProviding {
     /// The underlying AR view.
     /// This is set by the ARSwiftUIView when it is available.
-    fileprivate var arView: ARViewType?
+    fileprivate var arView: ARViewType!
     
     /// The AR session.
-    var session: ARSession? {
-        arView?.session
+    @objc dynamic var session: ARSession {
+        arView.session
     }
     
     /// Creates a raycast query that originates from a point on the view, aligned with the center of the camera's field of view.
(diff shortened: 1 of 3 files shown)


Sources/ArcGISToolkit/Components/AR/ARSwiftUIView.swift
```diff
@@ -143,10 +126,47 @@ class ARSwiftUIViewProxy {
...
+        uiView.goal = goal
+        uiView.setActive(active, animated: true)
+    }
+    
+    func makeCoordinator() -> Coordinator {
+        Coordinator()
+    }
+    
+    class Coordinator: NSObject, ARCoachingOverlayViewDelegate {}
+}
```
rolson, line 172: Can you doc this please

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
```diff
@@ -117,11 +118,26 @@ public struct TableTopSceneView: View {
...
                 .onAppear {
-                    arViewProxy.session?.run(configuration)
+                    arViewProxy.session.run(configuration)
                 }
                 .onDisappear {
-                    arViewProxy.session?.pause()
+                    arViewProxy.session.pause()
+                }
+            
+            ARCoachingOverlay(goal: .horizontalPlane)
```
rolson, line 127: Instead of messing with the opacity, I think I would just put this in an if block:

```swift
if !coachingOverlayIsHidden { 
    ...
}
```

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
```diff
@@ -170,12 +178,15 @@ public struct TableTopSceneView: View {
               let planeGeometry = ARSCNPlaneGeometry(device: device)
         else { return }
         
+        // Disable coaching overlay when a plane node is found.
+        coachingOverlayIsActive = false
```
rolson, line 182: does this ever need to be set to true again? Can nodes/planes go away?

des12437: No, the nodes are only removed with `node.removeFromParentNode()` when the ArcGIS Scene is added to the view, so the coaching overlay is not needed at that point. If we [support](https://developer.apple.com/documentation/arkit/arsessionobserver/2941046-sessionshouldattemptrelocalizati) relocalization then the coaching overlay may need to be shown again.

Sources/ArcGISToolkit/Components/AR/TableTopSceneView.swift
```diff
@@ -117,12 +118,28 @@ public struct TableTopSceneView: View {
...
+                    .sessionProvider(arViewProxy)
+                    .active(coachingOverlayIsActive)
+                    .allowsHitTesting(false)
+                    .overlay(alignment: .top) {
+                        if !helpText.isEmpty {
+                            Text(helpText)
+                                .frame(maxWidth: .infinity, alignment: .center)
+                                .padding(8)
+                                .background(.regularMaterial, ignoresSafeAreaEdges: .horizontal)
+                                .animation(.easeInOut, value: 1)
```
rolson, line 138: this animation isn't working for me. I would think it should be this:
```suggestion
                                .transition(.opacity)
```

but that doesn't seem to be working either.

rolson: [This works](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/461)

rolson (dismissed):

rolson (approved):

pgruenler (approved):

PR #468, opened 2023-10-10
AR Flyover: Do not require initial heading

rolson:
The initial heading isn't required if we use world alignment of gravity and heading. This simplifies the API.

PR #475, opened 2023-10-17
AR Flyover: Add initial heading parameter

CalebRas:
## Description

This PR adds an `initialHeading` parameter to the `FlyoverSceneView` initializers to allow the user to initialize the scene view looking a specific point/object just as they could with a `SceneView`. Passing a value of `nil` as the initial heading sets the heading of the scene to compass heading. This parameter is being re-added as it was removed in #468.

## Linked Issue(s)

- `swift/issues/4625`

## How To Test

- **Note**: As of creation, this PR requires `swift-daily` to compile.

### DIFF
--- a/Sources/ArcGISToolkit/Components/AR/FlyoverSceneView.swift
+++ b/Sources/ArcGISToolkit/Components/AR/FlyoverSceneView.swift
@@ -82,8 +82,8 @@ public struct FlyoverSceneView: View {
     /// be effectively viewed in augmented reality. One such property is the camera controller.
     public init(
         initialLocation: Point,
-        initialHeading: Double? = nil,
         translationFactor: Double,
+        initialHeading: Double? = nil,
         @ViewBuilder sceneView: @escaping (SceneViewProxy) -> SceneView
     ) {
         let camera = Camera(
--- a/Sources/ArcGISToolkit/Components/AR/FlyoverSceneView.swift
+++ b/Sources/ArcGISToolkit/Components/AR/FlyoverSceneView.swift
@@ -36,11 +36,11 @@ public struct FlyoverSceneView: View {
     /// - Parameters:
     ///   - initialLatitude: The initial latitude of the scene view's camera.
     ///   - initialLongitude: The initial longitude of the scene view's camera.
-    ///   - initialHeading: The initial heading of the scene view's camera. A value of `nil` means
-    ///   the scene view's heading will be initially oriented to compass heading.
     ///   - initialAltitude: The initial altitude of the scene view's camera.
     ///   - translationFactor: The translation factor that defines how much the scene view translates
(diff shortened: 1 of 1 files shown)


Sources/ArcGISToolkit/Components/AR/FlyoverSceneView.swift
```diff
@@ -70,16 +82,29 @@ public struct FlyoverSceneView: View {
     /// be effectively viewed in augmented reality. One such property is the camera controller.
     public init(
         initialLocation: Point,
+        initialHeading: Double? = nil,
         translationFactor: Double,
         @ViewBuilder sceneView: @escaping (SceneViewProxy) -> SceneView
```
des12437, line 87:
```suggestion
        translationFactor: Double,
        initialHeading: Double? = nil,
        @ViewBuilder sceneView: @escaping (SceneViewProxy) -> SceneView
```
Parameters with default arguments should be placed at the end of the parameter list.

des12437 (dismissed):

Sources/ArcGISToolkit/Components/AR/FlyoverSceneView.swift
```diff
@@ -104,7 +131,12 @@ public struct FlyoverSceneView: View {
         SceneViewReader { sceneViewProxy in
             sceneViewBuilder(sceneViewProxy)
                 .cameraController(cameraController)
-                .onAppear { session.start() }
+                .onAppear {
+                    if shouldOrientToCompass {
+                        session.configuration.worldAlignment = .gravityAndHeading
+                    }
+                    session.start()
+                }
```
rolson, line 139: Now that the view needs to know about the configuration, I think it would be cleaner to just pass the configuration to the start method. Then the configuration property on 163 goes away.

Sources/ArcGISToolkit/Components/AR/FlyoverSceneView.swift
```diff
@@ -21,12 +21,14 @@ public struct FlyoverSceneView: View {
...
-    /// A Boolean value indicating whether to orient the scene view's initial heading to compass heading.
-    private let shouldOrientToCompass: Bool
     /// The translation factor.
     private let translationFactor: Double
+    /// A Boolean value indicating whether to orient the scene view's initial heading to compass heading.
+    private let shouldOrientToCompass: Bool
     /// The closure that builds the scene view.
     private let sceneViewBuilder: (SceneViewProxy) -> SceneView
+    /// The configuration used for the AR session.
+    private let configuration: ARConfiguration
```
rolson, line 31: I don't think we need this. We just need to create it in onAppear and pass it to start(...)

rolson (approved):

des12437 (approved):

PR #480, opened 2023-10-19
`FormView` - Use new `updateValue`, `value` and `formattedValue` APIs

dfeinzimer:
Closes Apollo 116

Note this PR affected by some known issues:
Apollo 233
Apollo 240

### DIFF
--- a/Sources/ArcGISToolkit/Components/Form/FormInputs/ComboBoxInput.swift
+++ b/Sources/ArcGISToolkit/Components/Form/FormInputs/ComboBoxInput.swift
@@ -132,15 +132,15 @@ struct ComboBoxInput: View {
         .padding([.bottom], elementPadding)
         .onAppear {
             codedValues = featureForm!.codedValues(fieldName: element.fieldName)
-            selectedValue = codedValues.first { $0.name == element.value }
+            selectedValue = codedValues.first { $0.name == element.formattedValue }
         }
         .onChange(of: selectedValue) { newValue in
             guard newValue?.name != inputModel.value else {
                 return
             }
-
+            
             requiredValueMissing = element.isRequired && newValue == nil
-            featureForm?.feature.setAttributeValue(newValue?.code, forKey: element.fieldName)
+            try? element.updateValue(newValue?.code)
             model.evaluateExpressions()
         }
         .onChange(of: inputModel.value) { newValue in
--- a/Sources/ArcGISToolkit/Components/Form/FormInputs/ComboBoxInput.swift
+++ b/Sources/ArcGISToolkit/Components/Form/FormInputs/ComboBoxInput.swift
@@ -65,8 +65,12 @@ struct ComboBoxInput: View {
     /// - Parameters:
     ///   - element: The input's parent element.
     init(element: FieldFormElement) {
-        self.element = element
+        precondition(
+            element.input is ComboBoxFormInput,
+            "\(Self.self).\(#function) element's input must be \(ComboBoxFormInput.self)."
+        )
         
+        self.element = element
         let input = element.input as! ComboBoxFormInput
         self.noValueLabel = input.noValueLabel
         self.noValueOption = input.noValueOption
--- a/Sources/ArcGISToolkit/Components/Form/FormInputs/ComboBoxInput.swift
+++ b/Sources/ArcGISToolkit/Components/Form/FormInputs/ComboBoxInput.swift
@@ -24,6 +24,9 @@ struct ComboBoxInput: View {
     /// The model for the ancestral form view.
     @EnvironmentObject var model: FormViewModel
     
+    /// The model for the input.
+    @StateObject var inputModel: FormInputModel
+    
     /// The set of options in the combo box.
     @State private var codedValues = [CodedValue]()
     
@@ -39,9 +42,6 @@ struct ComboBoxInput: View {
     /// The selected option.
     @State private var selectedValue: CodedValue?
     
-    /// The feature form containing the input.
-    private var featureForm: FeatureForm?
-    
     /// The input's parent element.
     private let element: FieldFormElement
     
@@ -50,9 +50,6 @@ struct ComboBoxInput: View {
     
     /// The display state value for `nil` value options.
     private let noValueOption: FormInputNoValueOption
-    
-    /// The model for the input.
-    @StateObject var inputModel: FormInputModel
     
     /// A subset of coded values with names containing `filterPhrase` or all of the coded values
     /// if `filterPhrase` is empty.
@@ -66,12 +63,11 @@ struct ComboBoxInput: View {
     
     /// Creates a view for a combo box input.
     /// - Parameters:
-    ///   - featureForm: The feature form containing the input.
     ///   - element: The input's parent element.
-    ///   - input: The input configuration of the view.
-    init(featureForm: FeatureForm?, element: FieldFormElement, input: ComboBoxFormInput) {
-        self.featureForm = featureForm
+    init(element: FieldFormElement) {
         self.element = element
+        
+        let input = element.input as! ComboBoxFormInput
         self.noValueLabel = input.noValueLabel
         self.noValueOption = input.noValueOption
         
@@ -82,12 +78,10 @@ struct ComboBoxInput: View {
     
     /// Creates a view for a combo box input.
     /// - Parameters:
-    ///   - featureForm: The feature form containing the input.
     ///   - element: The input's parent element.
     ///   - noValueLabel: The text used to represent a `nil` value.
     ///   - noValueOption: The display state value for `nil` value options.
-    init(featureForm: FeatureForm?, element: FieldFormElement, noValueLabel: String, noValueOption: FormInputNoValueOption) {
-        self.featureForm = featureForm
+    init(element: FieldFormElement, noValueLabel: String, noValueOption: FormInputNoValueOption) {
         self.element = element
         self.noValueLabel = noValueLabel
         self.noValueOption = noValueOption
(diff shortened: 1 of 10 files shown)


Sources/ArcGISToolkit/Components/Form/FormInputs/ComboBoxInput.swift
```diff
@@ -131,21 +125,21 @@ struct ComboBoxInput: View {
...
         }
-        .onChange(of: selectedValue) { newValue in
-            guard newValue?.name != inputModel.value else {
-                return
+        .onChange(of: selectedValue) { selectedValue in
+            requiredValueMissing = inputModel.isRequired && selectedValue == nil
+            do {
+                try element.updateValue(selectedValue?.code)
+            } catch {
+                print(error.localizedDescription)
```
dfeinzimer, line 136: This is not handling any errors thrown yet. I will do that in a separate PR.

Sources/ArcGISToolkit/Components/Form/FormInputs/FormInputModel.swift
```diff
@@ -18,15 +18,18 @@ import SwiftUI
 /// A model for an input in a form.
 ///
 /// - Since: 200.3
-public class FormInputModel: ObservableObject {
+class FormInputModel: ObservableObject {
```
dfeinzimer, line 21: I don't think this needs/should be public, should it?

Sources/ArcGISToolkit/Components/Form/FormInputs/FormInputModel.swift
```diff
@@ -18,15 +18,18 @@ import SwiftUI
...
     
     /// A Boolean value indicating whether a value in the input is editable.
     @Published var isEditable: Bool
     
     /// The value of the input.
-    @Published var value: String
+    @Published var value: Any?
+    
+    /// The formatted value of the input.
+    @Published var formattedValue: String
```
dfeinzimer, line 32: `value` is now Any?. Because `Any` is not comparable, I've added a published `formattedValue: String` property. Inputs can instead now monitor this for changes.

Sources/ArcGISToolkit/Components/Form/FormInputs/SwitchInput.swift
```diff
@@ -33,32 +36,23 @@ struct SwitchInput: View {
     @State private var requiredValueMissing = false
     
     /// A Boolean value indicating whether the switch is toggled on or off.
-    @State private var switchState: Bool = false
+    @State private var isOn: Bool = false
```
dfeinzimer, line 39: `isOn` seems like a clearer name to me and matches the name of the Toggle's parameter.

Sources/ArcGISToolkit/Components/Form/FormViewModel.swift
```diff
@@ -27,7 +27,7 @@ public class FormViewModel: ObservableObject {
     @Published private var table: ServiceFeatureTable?
     
     /// The feature form.
-    @Published private var featureForm: FeatureForm?
+    @Published var featureForm: FeatureForm?
```
dfeinzimer, line 30: Provides access to all of the input views.

Sources/ArcGISToolkit/Components/Form/FormInputs/ComboBoxInput.swift
```diff
@@ -66,12 +63,11 @@ struct ComboBoxInput: View {
...
     /// - Parameters:
-    ///   - featureForm: The feature form containing the input.
     ///   - element: The input's parent element.
-    ///   - input: The input configuration of the view.
-    init(featureForm: FeatureForm?, element: FieldFormElement, input: ComboBoxFormInput) {
-        self.featureForm = featureForm
+    init(element: FieldFormElement) {
         self.element = element
+        
+        let input = element.input as! ComboBoxFormInput
```
mhdostal, line 70: Would this be good as a `precondition`? `precondition(element.input is ComboBoxFormInput)` to verify that the force-unwrapping will succeed and potentially give developers more information?

Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeInput.swift
```diff
@@ -66,30 +60,30 @@ struct DateTimeInput: View {
...
+        .onChange(of: model.focusedFieldName) { newFocusedFieldName in
+            isEditing = newFocusedFieldName == element.fieldName
+        }
         .onAppear {
-            if inputModel.value.isEmpty {
+            if inputModel.formattedValue.isEmpty {
                 date = nil
             } else {
-                date = try? Date(inputModel.value, strategy: .arcGISDateParseStrategy)
+                date = try? Date(inputModel.formattedValue, strategy: .arcGISDateParseStrategy)
```
mhdostal: For `DateTimeInput`s, `inputModel.value` is already a date.

image

Sources/ArcGISToolkit/Components/Form/FormInputs/ComboBoxInput.swift
```diff
@@ -131,21 +125,21 @@ struct ComboBoxInput: View {
...
-            requiredValueMissing = element.isRequired && newValue == nil
-            featureForm?.feature.setAttributeValue(newValue?.code, forKey: element.fieldName)
             model.evaluateExpressions()
         }
-        .onChange(of: inputModel.value) { newValue in
-            let codedValues = featureForm!.codedValues(fieldName: element.fieldName)
-            selectedValue = codedValues.first { $0.name == newValue }
+        .onChange(of: inputModel.formattedValue) { _ in
+            let codedValues = model.featureForm!.codedValues(fieldName: element.fieldName)
+            selectedValue = codedValues.first { $0.name == inputModel.formattedValue }
```
mhdostal, line 142: This doesn't cause an infinite loop with the `.onChange(of: selectedValue)`, does it?

Sources/ArcGISToolkit/Components/Form/FormInputs/FormInputModel.swift
mhdostal: No

Sources/ArcGISToolkit/Components/Form/FormInputs/FormInputModel.swift
mhdostal: Are `value` and `formattedValue` guaranteed to be in sync with each other? I would think so, so this should be fine.

mhdostal: So because `Any?` is not `Comparable`, you can't do `.onChange(of: inputModel.value)`?

Sources/ArcGISToolkit/Components/Form/FormInputs/SwitchInput.swift
```diff
@@ -90,54 +83,23 @@ struct SwitchInput: View {
             .disabled(!inputModel.isEditable)
             .padding([.bottom], elementPadding)
             .onAppear {
-                if element.value.isEmpty {
+                if element.formattedValue.isEmpty {
                     fallbackToComboBox = true
                 } else {
-                    switchState = isOn
+                    isOn = input.onValue.name == inputModel.formattedValue
```
mhdostal, line 89: You could compare codes here too, right? May be a bit faster (but not a big deal).

`input.onValue.code == inputModel.value`???

Or would you have to cast that to the correct code type?

Sources/ArcGISToolkit/Components/Form/TextInputFooter.swift
```diff
@@ -103,7 +103,7 @@ struct TextInputFooter: View {
         .foregroundColor(validationError == nil ? .secondary : .red)
         .onChange(of: text) { newText in
             if !hasPreviouslySatisfiedMinimum && !fieldType.isNumeric {
-                if newText.count >= lengthRange.lowerBound {
+                if newText.count >= lengthRange!.lowerBound {
```
mhdostal, line 106: Is this OK if `lengthRange` is optional and can be nil?

mhdostal (changes requested):

Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeInput.swift
mhdostal: Ha ha, looks like you already changed this one! Nice!

Sources/ArcGISToolkit/Components/Form/FormInputs/FormInputModel.swift
dfeinzimer: That's correct.

Sources/ArcGISToolkit/Components/Form/FormInputs/DateTimeInput.swift
dfeinzimer: Yep I had my eye on it and revisited this morning.

dfeinzimer:
> Are value and formattedValue guaranteed to be in sync with each other?

As far as I've seen so far, yes they are.

Sources/ArcGISToolkit/Components/Form/FormInputs/SwitchInput.swift
dfeinzimer: Correct, we'd need the cast which is why I'm depending on formattedValue to be the coded value's name.

Sources/ArcGISToolkit/Components/Form/FormInputs/ComboBoxInput.swift
dfeinzimer: We could do that, but seeing as the individual input views aren't public (at least for now) that error would be on our side, right?

I just gave it a try and we'd still need the force-unwrap (the precon. doesn't seem to alleviate that need) so we'd just be adding a line to each init.

That said, still happy to add them in if you like?

Sources/ArcGISToolkit/Components/Form/FormInputs/ComboBoxInput.swift
dfeinzimer: From what I've seen so far, no.

Sources/ArcGISToolkit/Components/Form/TextInputFooter.swift
dfeinzimer: Should be. Length ranges are only applicable to non-numeric text input so the `!fieldType.isNumeric` condition above on 105 should protect there.

dfeinzimer:
I think I've responded to all of your feedback. No code changes so far. Ready for re-review.

Sources/ArcGISToolkit/Components/Form/FormInputs/ComboBoxInput.swift
mhdostal: When the `precondition` fails, does the user see a better message than the 'unwrapping a nil' message? That would be the reason to put it in there

dfeinzimer: The default precondition error would look like:

> Precondition failed

I've added preconditions with a more helpful message:

> Precondition failed: SingleLineTextInput.init(element:) element's input must be TextBoxFormInput.

mhdostal (approved):
The `preconditions` are great, thanks!

PR #482, opened 2023-10-20
`FormView` - Example: Identify features from all layers

dfeinzimer:
This is just a small QOL enhancement so that the example identifies features on any layers. 

The identify code no longer needs to be adjusted depending on which operational layer the target feature is on.

Good maps to test this with are "Wind Turbines" ID ending `...2dba` and Redland's water hydrants `...abe4`

### DIFF
--- a/Examples/Examples/FormViewExampleView.swift
+++ b/Examples/Examples/FormViewExampleView.swift
@@ -104,7 +104,14 @@ extension FormViewExampleView {
            })
             .cancellationToNil()?
             .get()
-            .first {
+            .first(where: { result in
+                if let feature = result.geoElements.first as? ArcGISFeature,
+                   let _ = (feature.table?.layer as? FeatureLayer)?.featureFormDefinition {
+                    return true
+                } else {
+                    return false
+                }
+            }) {
             return identifyLayerResult.geoElements.first as? ArcGISFeature
         }
         return nil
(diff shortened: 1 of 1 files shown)


Examples/Examples/FormViewExampleView.swift
```diff
@@ -96,18 +96,16 @@ extension FormViewExampleView {
...
                 tolerance: 10
                )
            })
             .cancellationToNil()?
             .get()
-            .geoElements
-            .first as? ArcGISFeature {
-            return feature
+            .first {
+            return identifyLayerResult.geoElements.first as? ArcGISFeature
```
mhdostal, line 108: This will only return the first layer result that has `ArcGISFeature`s. We want the first layer result that has a feature form definition.
```suggestion
            .first(where: { result in
                if let feature = result.geoElements.first as? ArcGISFeature,
                   let _ = (feature.table?.layer as? FeatureLayer)?.featureFormDefinition {
                    return true
                } else {
                    return false
                }
            }) {
            return identifyLayerResult.geoElements.first as? ArcGISFeature
```

mhdostal (changes requested):

mhdostal (approved):

dfeinzimer:
Made one more change, the `let _ =` was causing a syntax warning.

mhdostal (approved):

PR #483, opened 2023-10-24
`FloatingPanel` - Improve keyboard avoidance, `FormView` -  Scroll to focused form element

dfeinzimer:
Closes Apollo 162

PR #484, opened 2023-10-24
`FormView` - Text input refactor

dfeinzimer:
Closes Apollo 176, 206

- Refactors `MultiLineTextInput` and `SingleLineTextInput` into one new unified view: `TextInput`.
- Improves adherence to design spec for read only text fields.
- Multiline text fields are now auto-expanded to the height necessary to show their full content.
- Known issue Apollo 269: Cursor in multiline text can become hidden behind keyboard


Sources/ArcGISToolkit/Components/Form/FormInputs/MultiLineTextInput.swift
```diff
@@ -63,36 +63,11 @@ struct MultiLineTextInput: View {
...
-                TextEditor(text: $text)
-                    .disabled(!inputModel.isEditable)
-            }
-            if isFocused && !text.isEmpty && inputModel.isEditable {
-                ClearButton { text.removeAll() }
-            }
+        if inputModel.isEditable {
+            textEditor
+        } else {
+            Text(text.isEmpty ? "--" : text)
```
mhdostal, line 69: Can you make sure Kotlin is doing this as well?

Sources/ArcGISToolkit/Components/Form/FormInputs/MultiLineTextInput.swift
```diff
@@ -128,4 +103,44 @@ private extension MultiLineTextInput {
...
+            TextEditor(text: $text)
+                .scrollContentBackgroundHidden()
+            if isFocused && !text.isEmpty && inputModel.isEditable {
+                ClearButton { text.removeAll() }
+            }
+        }
+        .background(.clear)
+        .focused($isFocused)
+        .foregroundColor(isPlaceholder ? .secondary : .primary)
+        .frame(minHeight: 75, maxHeight: 150)
```
mhdostal, line 119: Will this prevent the height from expanding to display all the text? In the issue we talked about always showing the entirety of the text, or am I misremembering?

Sources/ArcGISToolkit/Components/Form/FormInputs/MultiLineTextInput.swift
dfeinzimer: They will open an issue to add it as discussed in #apollo just now.

Sources/ArcGISToolkit/Components/Form/FormInputs/MultiLineTextInput.swift
dfeinzimer: Yes, my bad I see that you did mention it in 206. I think I have a solution but it will depend on the `FormViewModel.focusedElement` proposed in #483 in order to be usable so I'll take this back to draft in the mean time to get that merged first.

Sources/ArcGISToolkit/Components/Form/FormInputs/TextInput.swift
```diff
@@ -92,67 +114,99 @@ struct MultiLineTextInput: View {
...
-            }
-            if isFocused {
-                model.focusedElement = element
-            } else if model.focusedElement == element {
-                model.focusedElement = nil
-            }
+    }
+    
+    /// The range of valid values for a numeric input field.
+    var rangeDomain: RangeDomain? {
```
mhdostal, line 204: You should be able to check the `element.domain` now; I think Ivan fixed it.

Sources/ArcGISToolkit/Components/Form/FormInputs/TextInput.swift
```diff
@@ -0,0 +1,223 @@
...
+        )
+    }
+    
+    var body: some View {
+        InputHeader(label: element.label, isRequired: inputModel.isRequired)
+            .padding([.top], elementPadding)
+        if inputModel.isEditable {
+            textField
+        } else {
+            Text(text.isEmpty ? "--" : text)
```
mhdostal, line 67: Looks a little cramped, so I think it needs the same padding as in `formInputStyle()`. Not sure about adding the minimum frame height.
```suggestion
            Text(text.isEmpty ? "--" : text)
                .padding([.horizontal], 10)
                .padding([.vertical], 5)
```

mhdostal (changes requested):

Sources/ArcGISToolkit/Components/Form/FormInputs/TextInput.swift
dfeinzimer: Yup that seems to work good now.

mhdostal (approved):

PR #486, opened 2023-10-30
Dynamic Entity Arcade Callout

rolson:
The code was reviewed internally already. But this componentizes it and adds it to the toolkit.

- View.evaluateArcadeExpression view modifier
- View.onReceive view modifiers (now that this is utilized in 2 examples)
- Example for showing arcade expression in a callout for a dynamic entity


rolson:
Closing this as we are going to consider the evaluateArcadeExpression as a view modifier for the SDK instead.

PR #488, opened 2023-10-31
Popup simplification

rolson:
I noticed the popup code is more complex than it needed to be. It was converting the result of an async throwing method into a Result unnecessarily.


rolson:
All uses of `cancellationToNil()` are removed with these changes. I think we should deprecate it potentially. Thoughts?

dfeinzimer (approved):
Thanks for grabbing the tutorial steps too!

I'd be in favor of the deprecation.

mhdostal (approved):

PR #491, opened 2023-11-01
Deprecate cancellationToNil

rolson:
it's no longer used in our examples so deprecating it.


Sources/ArcGISToolkit/Extensions/Swift/Result.swift
```diff
@@ -25,6 +25,8 @@ public extension Result where Failure == Error {
     
     /// Converts the result to a `nil` in the case of a user cancelled error.
     /// - Returns: `Self` or `nil` if there was a cancellation error.
+    /// - Attention: Deprecated at 200.3.
+    @available(*, deprecated, message: "Check the 'failure' for 'CancellationError' instead.")
```
rolson, line 29:
```suggestion
    @available(*, deprecated, message: "Check the 'failure' case for 'CancellationError' instead.")
```

dfeinzimer (approved):

mhdostal (approved):

PR #492, opened 2023-11-01
Auto-refresh popup view for dynamic entities

rolson:
Auto-refresh popup view for dynamic entities.

This is hard to test because the SDK doesn't allow you to create popups for dynamic entities because of 2 issues. I added a workaround in the SDK and I can provide that to the reviewers of this PR. 

Toolkit example code for this is in `ryan/dynamicEntityPopup` branch if you want to try test it out.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Popups/PopupView.swift
+++ b/Sources/ArcGISToolkit/Components/Popups/PopupView.swift
@@ -41,6 +41,7 @@ import ArcGIS
 /// - Displays media (images and charts) full-screen.
 /// - Supports hyperlinks in text, media, and fields elements.
 /// - Fully supports dark mode, as do all Toolkit components.
+/// - Supports auto-refresh for popups where the geo element is a dynamic entity.
 ///
 /// **Behavior**
 ///
@@ -71,7 +72,7 @@ public struct PopupView: View {
     private var showCloseButton = false
     
     /// The result of evaluating the popup expressions.
-    @State private var evaluateExpressionsResult: Result<[PopupExpressionEvaluation], Error>?
+    @State private var evaluation: Evaluation?
 
     /// A binding to a Boolean value that determines whether the view is presented.
     private var isPresented: Binding<Bool>?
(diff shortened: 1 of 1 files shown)


rolson:
@mhdostal how hard would it be to preserve the expanded/collapsed state when the dynamic entity changes? Notice in the video below, when the entity changes, the collapsed section expands:

PopupDynamicEntity

mhdostal:
> @mhdostal how hard would it be to preserve the expanded/collapsed state when the dynamic entity changes? Notice in the video below, when the entity changes, the collapsed section expands:

It seems the entire list of popup elements are being redrawn, as the `PopupElementList()` is in a block dictated by the `@State private var evaluation: Evaluation?` var. The default value for the `DisclosureGroup.isExpanded` is `true`, which is why it's expanding when redrawn.

A solution would be to have a model object for each popup element which contained an `isExpanded` property. That would prevent the default state from being reset each redraw.  We've considered this for the `FormView`, as we're trying to set the initial state of a Group Element. I haven't tried this as we'd need to refactor some stuff in the `FormView`, but might this work?

rolson:
Interesting. I'm not sure if that would work, I'd have to think through it. Either way, it can happen outside this PR.

mhdostal:
> Either way, it can happen outside this PR.

We need to incorporate something like that into the `FeatureForm` work, so we'll take a look at porting our solution to the popup.

Sources/ArcGISToolkit/Components/Popups/PopupView.swift
```diff
@@ -126,13 +126,37 @@ public struct PopupView: View {
...
+            await evaluateExpressions()
+        }
+        .task(id: ObjectIdentifier(popup)) {
+            // If the popup is showing for a dynamic entity, then observe
+            // the changes and update the popup accordingly.
+            guard let dynamicEntity = popup.geoElement as? DynamicEntity else { return }
+            for await changes in dynamicEntity.changes {
+                if changes.dynamicEntityWasPurged {
+                    break
+                }
```
mhdostal, line 140: Just for my own information, what happens to the popup/feature on the map when the Dynamic Entity is purged?

mhdostal (approved):
Looks cool.

Sources/ArcGISToolkit/Components/Popups/PopupView.swift
rolson: it would simply stop updating the popup

mhdostal: Does the popup get dismissed or the feature representing the dynamic entity get removed?

rolson: The popup is not dismissed.

I believe that purged means that it was removed from rendering because it was purged from the datasource.

dfeinzimer (approved):

dg0yal:
This is pretty cool.
 You could build a lightweight dashboard just based on this

PR #496, opened 2023-11-02
Update `SearchViewModel`, fix nullability

yo1995:
Change nullability of the center diff value. With the recent changes in ~~`swift/pull/4674`~~ `swift/pull/4433` and others to handle the GeometryEngine exception & nullability, the `distance` method no longer return optional value.

image


rolson (approved):

rolson:
> did nullability change for that method between 200.1 and now? I thought I un-did all of the nullability changes.

I un-did all of the nullability changes I inadvertently introduced. But Ting pointed out a PR earlier in August which caused some changes.

dfeinzimer (approved):

PR #497, opened 2023-11-03
AR: Add unit tests

des12437:
Adds unit tests for FlyoverSceneView and TableTopSceneView.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/FlyoverSceneView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/FlyoverSceneView.swift
@@ -30,7 +30,7 @@ public struct FlyoverSceneView: View {
     /// The camera controller that we will set on the scene view.
     @State private var cameraController: TransformationMatrixCameraController
     /// The current interface orientation.
-    @State var interfaceOrientation: InterfaceOrientation?
+    @State private var interfaceOrientation: InterfaceOrientation?
     
     /// Creates a fly over scene view.
     /// - Parameters:
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/FlyoverSceneView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/FlyoverSceneView.swift
@@ -20,11 +20,11 @@ public struct FlyoverSceneView: View {
     /// The AR session.
     @StateObject private var session = ObservableARSession()
     /// The initial camera.
-    private let initialCamera: Camera
+    let initialCamera: Camera
     /// The translation factor.
-    private let translationFactor: Double
+    let translationFactor: Double
     /// A Boolean value indicating whether to orient the scene view's initial heading to compass heading.
-    private let shouldOrientToCompass: Bool
+    let shouldOrientToCompass: Bool
     /// The closure that builds the scene view.
     private let sceneViewBuilder: (SceneViewProxy) -> SceneView
     /// The camera controller that we will set on the scene view.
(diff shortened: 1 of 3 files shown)


Sources/ArcGISToolkit/Components/Augmented Reality/FlyoverSceneView.swift
```diff
@@ -187,3 +187,25 @@ private class ObservableARSession: NSObject, ObservableObject, ARSessionDelegate
...
+    }
+    
+    func getShouldOrientToCompass() -> Bool {
+        return shouldOrientToCompass
+    }
+    
+    func getInterfaceOrientation() -> InterfaceOrientation? {
+        return interfaceOrientation
+    }
+}
```
dfeinzimer, line 211: I can understand doing this for the camera controller and interface orientation because state properties are supposed to be private by convention but for the others, any reason to not just make the properties themselves internal?

rolson: If we need to test these, then the properties themselves should just be internal.

rolson (commented):
If we need to test these, then the properties themselves should just be internal.

Sources/ArcGISToolkit/Components/Augmented Reality/FlyoverSceneView.swift
des12437: > I can understand doing this for the camera controller and interface orientation because state properties are supposed to be private by convention but for the others, any reason to not just make the properties themselves internal?

No, I will make those properties internal.

dfeinzimer (approved):

Tests/ArcGISToolkitTests/ARTests.swift
```diff
@@ -0,0 +1,188 @@
...
+@testable import ArcGISToolkit
+
+@MainActor final class ARTests: XCTestCase {
+    func testFlyoverLocationInitWithDefaults() throws {
+        let sceneView = SceneView(scene: Scene())
+        
+        let view = FlyoverSceneView(
+            initialLocation: .init(
+                latitude: 0,
+                longitude: 0
```
rolson, line 26: it would probably be better to choose numbers other than 0 because 0 is sometimes a default and that could potentially imply "not-set"

Sources/ArcGISToolkit/Components/Augmented Reality/FlyoverSceneView.swift
```diff
@@ -187,3 +187,13 @@ private class ObservableARSession: NSObject, ObservableObject, ARSessionDelegate
...
+
+extension FlyoverSceneView {
+    func getCameraController() -> CameraController {
+        return cameraController
+    }
+    
+    func getInterfaceOrientation() -> InterfaceOrientation? {
+        return interfaceOrientation
+    }
+}
```
rolson, line 199: we can just make the state properties internal

dfeinzimer: Are we okay to forgo this recommendation from the `@State` docs?

> Declare state as private to prevent setting it in a memberwise initializer, which can conflict with the storage management that SwiftUI provides. [Source](https://developer.apple.com/documentation/swiftui/state#:~:text=an%20initial%20value.-,Declare%20state%20as%20private%20to%20prevent%20setting%20it%20in%20a%20memberwise%20initializer%2C%20which%20can%20conflict%20with%20the%20storage%20management%20that%20SwiftUI%20provides,-%3A)

rolson: I'm ok with it if we aren't setting it in a memberwise init. Otherwise I'm ok with not testing those two properties as I don't think it's worth adding a function to workaround it.

rolson: I'm ok either way

des12437: The properties are now internal since they are not set by the init, however, a better option could be to not test the state properties at all and create an `initialCamera

rolson: I'm ok not testing them if we feel we don't want to violate that Swift convention.

rolson (approved):
thank you!

dfeinzimer (approved):

des12437:
Thank you both.

PR #498, opened 2023-11-03
Initial expression evaluation code

mhdostal:
This allows the form display to continue even if there are initial expression evaluation errors.

Apollo 268

### DIFF
--- a/Sources/ArcGISToolkit/Components/Form/FormView.swift
+++ b/Sources/ArcGISToolkit/Components/Form/FormView.swift
@@ -70,11 +70,11 @@ public struct FormView: View {
             do {
                 isEvaluating = true
                 try await featureForm?.evaluateExpressions()
-                isEvaluating = false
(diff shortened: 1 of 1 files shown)


dfeinzimer (approved):
Is this something that'll stay this way or do we plan to require a successful initial evaluation at some point? Should we have an issue if so? At that time we could show an error instead of the endless spinner.

mhdostal:
> Is this something that'll stay this way or do we plan to require a successful initial evaluation at some point?

We need to figure out if clients will need access to any/all expression evaluation errors. If so, we would need to return those from `FeatureForm.evaluateExpressions()`.

I'm in the process of creating a discussion issue for that.

dfeinzimer:
> we would need to return those from `FeatureForm.evaluateExpressions()`.
> 
> I'm in the process of creating a discussion issue for that.

Sounds good.

PR #502, opened 2023-11-07
Add Privacy Manifest

zkline101:
Swift issue: #4177

These are the things that need to be declared in the privacy manifest if we collect any of this info:
* [Describing data use](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_data_use_in_privacy_manifests)
- [Required Use APIs](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api)

I only listed in the privacy manifest that we use user defaults for a qualified reason. I didn't see any other items we need to list, please confirm this as a part of reviewing this PR.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -36,6 +36,9 @@ let package = Package(
             name: "ArcGISToolkit",
             dependencies: [
                 .product(name: "ArcGIS", package: "arcgis-maps-sdk-swift")
+            ],
+            resources: [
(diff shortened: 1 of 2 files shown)


dfeinzimer (approved):
Looks good, thank you for handling

mhdostal (approved):

PR #503, opened 2023-11-08
AR: Fix Issue with tapping a surface to place a scene in Tabletop Mode on iOS 15

des12437:
Fixes the tap gesture bug in iOS 15 for TableTop AR by setting the minimum drag distance to zero.


pgruenler (dismissed):
Fix looks good. A comment or two about the gestures and minimum distance might be nice.

pgruenler (approved):

rolson (approved):

PR #513, opened 2023-11-27
`FormView` - Example: Add interface to confirm cancel

dfeinzimer:
This slightly adjusts the behavior of the `FormView` example to confirm and close the form when there is a map tap before opening the next form. This takes care of a condition where a new form isn't displayed properly when another form was already open.

| Before  | After  |
|---|---|
| <video src="https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/assets/16397058/1e5ff073-0369-4417-a6d0-1344803eb99e">  | <video src="https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/assets/16397058/60a45da3-386f-44d1-9e04-41d957c03858">  |

### DIFF
--- a/Examples/Examples/FormViewExampleView.swift
+++ b/Examples/Examples/FormViewExampleView.swift
@@ -73,16 +73,16 @@ struct FormViewExampleView: View {
                     FormView(featureForm: featureForm)
                         .padding([.horizontal])
                 }
-            
-                .alert("Cancel editing?", isPresented: $isCancelConfirmationPresented) {
-                    Button("Cancel editing", role: .destructive) {
-                        formViewModel.undoEdits()
-                        featureForm = nil
-                        isFormPresented = false
-                    }
-                    Button("Continue editing", role: .cancel) { }
+                .alert("Discard edits", isPresented: $isCancelConfirmationPresented) {
+                        Button("Discard edits", role: .destructive) {
+                            formViewModel.undoEdits()
+                            featureForm = nil
+                            isFormPresented = false
+                        }
+                        Button("Continue editing", role: .cancel) { }
+                } message: {
+                    Text("Updates to this feature will be lost.")
                 }
-            
                 .environmentObject(formViewModel)
                 .navigationBarBackButtonHidden(isFormPresented)
                 .toolbar {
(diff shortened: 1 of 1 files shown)


Examples/Examples/FormViewExampleView.swift
```diff
@@ -16,6 +16,9 @@ import ArcGISToolkit
 import SwiftUI
 
 struct FormViewExampleView: View {
+    /// A Boolean value indicating whether the alert confirming the user's intent to cancel is displayed.
+    @State private var cancelConfirmationIsShowing = false
```
mhdostal, line 20: To match the existing `isPresented` and the alert argument label (also `isPresented`), maybe follow that form here:
```suggestion
    @State private var isCancelConfirmationPresented = false
```

Examples/Examples/FormViewExampleView.swift
```diff
@@ -67,6 +74,15 @@ struct FormViewExampleView: View {
                         .padding([.horizontal])
                 }
             
+                .alert("Cancel editing?", isPresented: $cancelConfirmationIsShowing) {
+                    Button("Cancel editing", role: .destructive) {
+                        formViewModel.undoEdits()
+                        featureForm = nil
+                        isPresented = false
```
mhdostal, line 81: Could we have a `didSet` block on the `featureForm` property which set `isPresented`? That might be more clear. In that case I would also change the `Cancel` button block, [L96](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/2fcb74eaa53f9ea6982c21a6da228da82ac15c4e/Examples/Examples/FormViewExampleView.swift#L96) to just set `featureForm = nil`.

And the same for [L106](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/2fcb74eaa53f9ea6982c21a6da228da82ac15c4e/Examples/Examples/FormViewExampleView.swift#L106)

Examples/Examples/FormViewExampleView.swift
```diff
@@ -67,6 +74,15 @@ struct FormViewExampleView: View {
                         .padding([.horizontal])
                 }
             
+                .alert("Cancel editing?", isPresented: $cancelConfirmationIsShowing) {
+                    Button("Cancel editing", role: .destructive) {
```
mhdostal, line 78: The AGOL map viewer has this:

image

So instead of "Cancel editing", maybe "Discard edits" as that is more descriptive. We should coordinate with Kotlin as well to make sure we're in sync with the verbiage.

mhdostal (changes requested):

Examples/Examples/FormViewExampleView.swift
dfeinzimer: Per discussion in Apollo Kotlin currently has no interface for this and just dismisses the form.

mhdostal (approved):

PR #519, opened 2023-11-29
Bump Swift Tools Version

dfeinzimer:


rolson (approved):

rolson:
We we also then want to change the readme to say Xcode 15 is required? Currently says 14.1 or newer.

dfeinzimer:
> Do we also then want to change the readme to say Xcode 15 is required? Currently says 14.1 or newer.

@des12437 took care of it in #516.

mhdostal (approved):

Issue #523, opened 2023-11-29
Bookmarks list row label is only selectable


des12437:
Testing with iPhone 15 Pro simulator with iOS 17, the `Bookmarks` example menu list items can only be selected when the label is tapped. The entire row is not selectable. This could be an issue in the Bookmarks component.

https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/assets/117859673/0373459e-a6a5-4e4b-afb8-f51a720c688d

PR #525, opened 2023-11-30
Bookmarks improvements

### LINKED ISSUE #520: Bookmarks menu does not show on Mac Catalyst
The Bookmarks component does not fully present in the `Bookmarks` example on Mac Catalyst with the following console logs:

```
UIScene property of UINSSceneViewController was accessed before it was set.
Bound preference SizePreferenceKey tried to update multiple times per frame.
```

Screenshot 2023-11-29 at 1 43 32 PM

Occasionally the component does fully present:
Screenshot 2023-11-29 at 1 54 57 PM

dfeinzimer:
Closes #520 & #523.

This resolves two issues found during certification and is a refactor/simplification of the Bookmarks component as well as its example and tutorial.

- Component
  - Resolves #520 - now works as expected on Catalyst. I tested this on Ventura and it behaved as expected so it may be a regression with Xcode 15.0 or Catalyst on Sonoma.
  - Expands horizontal tappable (#523).
  - Will show an error now if loading the GeoModel failed.
- Example & Tutorial
  - Uses map view proxy to change viewpoint so the map is no longer reinitialized on bookmark/viewpoint changes.
 
|   | Before | After |
|---|---|---|
| iPhone | iPhone-before | iPhone-after |
| iPad | iPad-before | iPad-after |
| Catalyst | cat-before | cat-after |


des12437 (approved):
Verified bug fixes for each iOS version and refactor changes look good.

Sources/ArcGISToolkit/Components/Bookmarks/Bookmarks.swift
```diff
@@ -130,47 +100,128 @@ public struct Bookmarks: View {
...
+                .onChange(of: selectedBookmark) { selectedBookmark in
+                    if let selectedBookmark {
+                        selectBookmark(selectedBookmark)
                     }
                 }
-                .onSizeChange {
-                    listHeight = $0.height
+        } else if let loadingError {
+            makeErrorMessage(with: loadingError)
+        } else if geoModel != nil {
```
mhdostal, line 120: If there is a `geoModel`, but it does not have any bookmarks, it will always show the `loading` view. It will never fall through to the `noBookmarks` view. Maybe something like:
```suggestion
        } else if geoModel != nil, !geoModel.isLoadingComplete {
```

```
extension GeoModel {
    var isLoadingComplete: Bool {
        loadStatus == .loaded || loadStatus == .failed
    }
}
```

mhdostal (changes requested):

des12437 (approved):

mhdostal (approved):

dfeinzimer:
Thank you both!

PR #531, opened 2023-12-02
Add `EnvironmentValues.isPortraitOrientation` extension

dfeinzimer:
~This is a proposal to add a new public protocol named `CompactAware` which makes an `isCompact: Bool` property available in conforming types.~

I noticed the Toolkit has is `isCompact` properties declared in 8 different places. It seems common enough  to warrant a reusable property.


dfeinzimer:
Thanks to a suggestion from @philium this has been simplified even further into a light extension on `EnvironmentValues`. 

Now users only need to declare one `@Environment` property. 🎉

Examples/Examples/UtilityNetworkTraceExampleView.swift
```diff
@@ -111,3 +103,11 @@ private extension ArcGISCredential {
...
     }
 }
+
+private extension EnvironmentValues {
+    /// A Boolean value indicating whether this environment has a compact horizontal size class and
+    /// a regular vertical size class.
+    var isPortraitOrientation: Bool {
+        horizontalSizeClass == .compact && verticalSizeClass == .regular
+    }
+}
```
mhdostal, line 113: Since this is the only place this is used in the Example app, I think it's OK to have it here as a duplicate of what's in the Toolkit. If it's needed outside the Toolkit in more than one or two places, we should consider making the Toolkit version public.

mhdostal (approved):
A very elegant solution!

Examples/Examples/UtilityNetworkTraceExampleView.swift
philium: I would strongly suggest not making the toolkit version public, no matter how many places it is used in the examples. The toolkit is a collection of components, not convenience extensions. While it may be tempting to make code public just so that it can be re-used in examples, it comes at a cost. As public code, it becomes public API that needs to be tested and maintained indefinitely. As internal code, it can be tweaked/changed/removed as needed, when needed. In my opinion, it is far better, especially for something as trivial as this, to just copy it into the examples target (or wherever else it may be useful outside of the toolkit) rather than making it part of the toolkit's public API.

mhdostal: @philium That makes sense, thanks for the feedback!

PR #544, opened 2023-12-13
Task updates to model classes.

mhdostal:
Updates to Task creation and disposal for `isVisible`, `isRequired`, `isEditable`, and `value`.

### DIFF
--- a/Sources/ArcGISToolkit/Components/FormView/FormInputs/FormInputModel.swift
+++ b/Sources/ArcGISToolkit/Components/FormView/FormInputs/FormInputModel.swift
@@ -19,7 +19,7 @@ import SwiftUI
 /// A model for an input in a form.
 ///
 /// - Since: 200.4
-class FormInputModel: ObservableObject {
+@MainActor class FormInputModel: ObservableObject {
     /// A Boolean value indicating whether a value in the input is required.
     @Published var isRequired: Bool
     
@@ -37,7 +37,10 @@ class FormInputModel: ObservableObject {
     private var tasks = [Task<Void, Never>]()
     
     deinit {
-        clearTasks()
+        tasks.forEach { task in
+            task.cancel()
+        }
+        tasks.removeAll()
     }
     
     /// Initializes a form input model.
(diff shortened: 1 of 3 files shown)


mhdostal:
@dfeinzimer I've updated the code to use TaskGroups; please take a look.

Sources/ArcGISToolkit/Components/FormView/FormInputs/FormInputModel.swift
```diff
@@ -34,67 +34,50 @@ class FormInputModel: ObservableObject {
...
-    }
-    
-    /// A detached task observing changes in the required state.
-    private var observeIsRequiredTask: Task<Void, Never> {
-        Task.detached { [unowned self] in
-            for await isRequired in element.$isRequired {
-                await MainActor.run {
-                    self.isRequired = isRequired
+        observationTask = Task {
+            await withThrowingTaskGroup(of: Void.self) { group in
```
dfeinzimer, line 53: Can this just be `withTaskGroup`? Doesn't seem like anything will throw.

Same question for the other location in `FormViewModel`.

dfeinzimer (commented):
Works good. Just the one question.

dfeinzimer (approved):

Sources/ArcGISToolkit/Components/FormView/FormInputs/FormInputModel.swift
```diff
@@ -34,67 +34,50 @@ class FormInputModel: ObservableObject {
...
-        tasks.removeAll()
-    }
-    
-    /// A detached task observing changes in the required state.
-    private var observeIsRequiredTask: Task<Void, Never> {
-        Task.detached { [unowned self] in
-            for await isRequired in element.$isRequired {
-                await MainActor.run {
-                    self.isRequired = isRequired
+        observationTask = Task {
```
philium, line 52: Should this be a detached task? Otherwise the task is implicitly holding onto self strongly, preventing it from the model from going away and the task from being cancelled.

Sources/ArcGISToolkit/Components/FormView/FormViewModel.swift
```diff
@@ -88,10 +84,8 @@ import SwiftUI
         isEvaluating = true
         evaluateTask = Task {
```
philium, line 85: Should this be a detached task? Otherwise the task is implicitly holding onto self strongly, preventing it from the model from going away and the task from being cancelled.

Sources/ArcGISToolkit/Components/FormView/FormViewModel.swift
```diff
@@ -46,27 +46,23 @@ import SwiftUI
...
     
     /// Kick off tasks to monitor `isVisible` for each element.
-    func initializeIsVisibleTasks() {
-        featureForm.elements.forEach { element in
-            let newTask = Task.detached { [unowned self] in
-                for await _ in element.$isVisible {
-                    await MainActor.run {
-                        self.updateVisibleElements()
+    private func initializeIsVisibleTasks() {
+        isVisibleTask = Task {
```
philium, line 55: Should this be a detached task? Otherwise the task is implicitly holding onto self strongly, preventing it from the model from going away and the task from being cancelled.

mhdostal: @philium Is this better? I had to add an `await` and `self.` in `for element in await self.featureForm.elements`.

Is the use of `[unowned self]` OK or should it be something else?

```
    private func initializeIsVisibleTasks() {
        isVisibleTask = Task.detached { [unowned self] in
            await withTaskGroup(of: Void.self) { [unowned self] group in
                for element in await self.featureForm.elements {
                    group.addTask {
                        for await _ in element.$isVisible {
                            guard !Task.isCancelled else { return }
                            await self.updateVisibleElements()
                        }
                    }
                }
            }
        }
    }
```

philium: Using `unowned` might be alright, but it is probably best to use `weak`, just to be on the safe side. Remember that a captures a value as it is, so `[weak self]` is only needed on the outer closure (`Task.detached { [weak self] in ... }`), not on the call to `withTaskGroup`.

PR #545, opened 2023-12-14
`FeatureFormView` - Make form input views public, support custom layouts

dfeinzimer:
Ref Apollo 294 & 353

### DIFF
--- a/Sources/ArcGISToolkit/Components/FormView/FormInputs/ComboBoxInput.swift
+++ b/Sources/ArcGISToolkit/Components/FormView/FormInputs/ComboBoxInput.swift
@@ -18,7 +18,8 @@ import SwiftUI
 /// A view for numerical value input.
 ///
 /// This is the preferable input type for long lists of coded value domains.
-struct ComboBoxInput: View {
+/// - Since: 200.4
+public struct ComboBoxInput: View {
     @Environment(\.formElementPadding) var elementPadding
     
     /// The model for the ancestral form view.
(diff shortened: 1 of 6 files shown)


mhdostal:
@dfeinzimer This looks great. Does the keyboard avoidance/scrolling work OK with custom views? Also, you should keep the original initializer along with the overloaded one you added.

dfeinzimer:
> Does the keyboard avoidance/scrolling work OK with custom views?

Keyboard avoidance should work automatically, yes. Scrolling a focused field to the top of the form/screen would fall on the user as they would supply the ScrollView and the associated proxy/reader to achieve that.

> Also, you should keep the original initializer along with the overloaded one you added.

👍 I will keep both.

dfeinzimer:
@mhdostal Should we prepend `Form` onto the names of the new public views to communicate their intended usage? E.g. `ComboBoxInput` -> `FormComboBoxInput`

mhdostal:
> Should we prepend `Form` onto the names of the new public views to communicate their intended usage? E.g. `ComboBoxInput` -> `FormComboBoxInput`

I can see the reasoning to do that; it makes it explicit that it's part of the Feature Form. However, the associated SDK object is `ComboBoxFormInput`. I think that would be confusing having two objects with very similar names, except for the placement of the word "Form". I can't think of an alternative.

dfeinzimer:
Closing this as not currently planned/prioritized and the branch has fallen significantly behind. The changes will still be available in the [Files changed](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/545/files) tab.

PR #548, opened 2023-12-18
Task updates to model classes

mhdostal:
Updates to Task creation and disposal for isVisible, isRequired, isEditable, and value.

Using `weak self` as mentioned [here](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/544#discussion_r1428589553) caused this line `await self?.updateVisibleElements()` in the inner-most `for await` to give an error: `Reference to captured var 'self' in concurrently-executing code`. So I left it as `unowned self`.


Sources/ArcGISToolkit/Components/FormView/FormInputs/FormInputModel.swift
```diff
@@ -49,33 +49,33 @@ import SwiftUI
...
+        observationTask = Task.detached { [unowned self] in
             await withTaskGroup(of: Void.self) { group in
                 // Observe isRequired changes.
-                group.addTask { [unowned self] in
-                    for await isRequired in await element.$isRequired {
+                group.addTask {
+                    for await isRequired in await self.element.$isRequired {
                         guard !Task.isCancelled else { return }
                         await MainActor.run {
                             self.isRequired = isRequired
```
dfeinzimer, line 59: I'm able to reproduce the crash here again. Unfortunately when I go to record what I'm doing it doesn't crash.

Screenshot 2023-12-19 at 14 11 11

> Fatal error: Attempted to read an unowned reference but object 0x60000ac3c380 was already deallocated

mhdostal: @philium I've tried to use [weak self] for this, but can't get past the compiler errors. I usually get _Reference to captured var 'self' in concurrently-executing code_, but sometimes it's _Value of optional type 'FormInputModel?' must be unwrapped to refer to member 'element' of wrapped base type 'FormInputModel'_, which I can fix by using `self?.`, but then I get the _For-in loop requires 'AsyncStream<Bool>?' to conform to 'AsyncSequence'; did you mean to unwrap optional?_ error.

dfeinzimer: Does the guard possibly need to be moved down into the closure sent to the main actor? It fixes the crash for me.

E.g.
```swift
for await isRequired in await self.element.$isRequired {
    await MainActor.run {
        guard !Task.isCancelled else { return }
        self.isRequired = isRequired
    }
}
```

philium: As I started looking into this, I couldn't help but get the feeling that the `task` modifier might be a better approach to using a model that creates and manages its own task with child tasks. I see why the model approach was used—the code needed to be re-usable because it is used by multiple views. But there's more than one way to make code re-usable. The model-based approach is problematic not only because of needing to manage the task but also because it needs to be initialized with an element. Due to the nature of `StateObject`, the model is only created once per instance of the view on-screen. If the element changes while the view is displayed, the model won't get re-created and the view will be in a bad state (since the model would have a stale element).

Instead of using a model, consider adding convenience modifiers for observing the element:

```swift
extension View {
    func onChangeOfIsEditable(
        of element: FieldFormElement,
        action: @escaping (_ newIsEditable: Bool) -> Void
    ) -> some View {
        return self
            .task(id: ObjectIdentifier(element)) {
                for await isEditable in element.$isEditable {
                    action(isEditable)
                }
            }
    }
    
    func onChangeOfIsRequired(
        of element: FieldFormElement,
        action: @escaping (_ newIsRequired: Bool) -> Void
    ) -> some View {
        return self
            .task(id: ObjectIdentifier(element)) {
                for await isRequired in element.$isRequired {
                    action(isRequired)
                }
            }
    }
    
    func onChangeOfValue(
        of element: FieldFormElement,
        action: @escaping (_ newValue: Any?, _ newFormattedValue: String) -> Void
    ) -> some View {
        return self
            .task(id: ObjectIdentifier(element)) {
                for await value in element.$value {
                    action(value, element.formattedValue)
                }
            }
    }
}
```

If you really want to, you could combine those into a single modifier, although I'm not sure that really is that much of an improvement—the first closure not having a label at the point of use looks really odd:

```swift
func observe(
    _ element: FieldFormElement,
    isRequiredChanged: @escaping (_ newIsRequired: Bool) -> Void,
    isEditableChanged: @escaping (_ newIsEditable: Bool) -> Void,
    valueChanged: @escaping (_ newValue: Any?, _ newFormattedValue: String) -> Void
) -> some View {
    return self
        .onChangeOfIsEditable(of: element, action: isEditableChanged)
        .onChangeOfIsRequired(of: element, action: isRequiredChanged)
        .onChangeOfValue(of: element, action: valueChanged)
}
```

In the views that use the model, replace the state object model property with four state properties. Perhaps even just three, depending on how `formattedValue` is used. It might be possible to do all that is needed related to `formattedValue` in the action closure for `onChangeOfValue` so that it doesn't need to be stored as state. Then use the modifiers (or single modifier) to update those state properties and use them in the view's body where needed. No more model. No more manual task management.

What do you think?

mhdostal: I think that sounds good; It's a bit more code, but it probably will work better for the reasons you mention.  Thanks!

dfeinzimer (approved):
Works good

Sources/ArcGISToolkit/Components/FormView/FormInputs/ComboBoxInput.swift
```diff
@@ -75,9 +78,10 @@ struct ComboBoxInput: View {
         self.noValueLabel = input.noValueLabel
         self.noValueOption = input.noValueOption
         
-        _inputModel = StateObject(
-            wrappedValue: FormInputModel(fieldFormElement: element)
-        )
+        value = element.value
+        formattedValue = element.formattedValue
+        isRequired = element.isRequired
+        isEditable = element.isEditable
```
philium, line 84: Setting state properties in the initializer has the same problem as setting state objects in the initializer—if the values passed to the view's initializer change while the view is on screen, the state properties won't be updated, leaving the view in an inconsistent state. Instead, the state properties will get updated automatically as soon as observation starts since the first value of the streams is the current value. That also means that setting `selectedValue` in `onAppear` is unnecessary. Without that, the formatted value is only used from inside the action closure of `onChangeOfValue(of:action:)`, so the `formattedValue` state property could go away.

Sources/ArcGISToolkit/Components/FormView/FormInputs/ComboBoxInput.swift
```diff
@@ -24,9 +24,12 @@ struct ComboBoxInput: View {
     /// The model for the ancestral form view.
     @EnvironmentObject var model: FormViewModel
     
-    /// The model for the input.
-    @StateObject var inputModel: FormInputModel
-    
+    /// State properties for element events.
```
philium, line 27: You don't want this to be a doc comment because it applies to more than just the following line. When a comment applies to more than one line, I recommend putting a blank line after it so that it is clear that it doesn't go with just the following line. Also, state properties should be private:

```swift
// State properties for element events.

@State private var isRequired: Bool = false
@State private var isEditable: Bool = false
@State private var value: Any?
@State private var formattedValue: String = ""
```

Issue #533, opened 2023-12-04
`Bookmarks` - Support automatic refresh


dfeinzimer:
Currently if the set of bookmarks or the bookmarks in the geomodel passed into the component change those changes won't be reflected.

If the component is hosted in popover or sheet this isn't an issue because the view will reinitialize on every presentation.

If the component is displayed statically on screen a workaround would be to manually trigger a reinitialization using [id(_:)](https://developer.apple.com/documentation/swiftui/view/id(_:)).

Issue #406, opened 2023-08-17
Issues found during DocC migration review


dfeinzimer:
This is a tracking issue for suggestions made during review of #392.

## All Components, Examples & Tutorials

- [x] Sort all imports (precedent [here](https://devtopia.esri.com/runtime/swift/pull/4411#discussion_r1029570) and [here](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/392#discussion_r1295072966)) #689 
- [x] Preambles with empty lines (see [here](https://github.com/Esri/Dejavu/pull/14#discussion_r1308889994)) #537

## Multiple Examples & Tutorials

- [x] https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/392#discussion_r1295059513 #681 

## `Authenticator`

- [x] https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/392#discussion_r1295003207
- [x] https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/392#discussion_r1295011502
- [x] https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/392#discussion_r1295015735
- [x] https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/392#discussion_r1295041825
- [x] https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/commit/1e13232c5420a8e41928453cba586f55d93bb3f7#r124597719

## `BasemapGallery`

- [x] https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/392#discussion_r1294872695
- [x] https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/392#discussion_r1295029122 #687
- [x] https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/392#discussion_r1295038501

## `Bookmarks`

- [x] https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/392#discussion_r1295043896 #549 
- [x] https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/392#discussion_r1295084387 #549 

## `FloatingPanel`

  - [x] https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/392#discussion_r1295054040 #682
[...]

Issue #401, opened 2023-08-15
Mac Catalyst UI issues


mhdostal:
### Bookmarks

- [ ] These should appear as simple list items, instead of button-looking items. #549 

Mac Catalyst, macOS Ventura v13.3.1

![image](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/assets/3998072/854848a4-9376-498f-bdd8-02df19a521e3)

### Floor Filter

- [ ] Button styles

![image](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/assets/3998072/a8772e4f-89ad-4469-b479-1804e3b9f0c1)

Also, it doesn't appear as if the `ProgressView` is showing when the map is loading.

### Utility Network Trace Tool

- [ ] Trace Results combo box is full-width; it should have some padding.

![image](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/assets/3998072/47f993d9-366f-48d8-ac80-d5530bb4c01a)

Issue #246, opened 2023-01-11
`Bookmarks` - Add binding `selection` init parameter


dfeinzimer:
Update:
See discussion below

---

Original: 
As an alternate to using a viewpoint, the bookmarks component can potentially take advantage of the `setBookmark(_:)` method on a map or scene view proxy.

philium:
Alternatively, it could be simplified by taking a binding to the selected bookmark as an initializer parameter, giving the caller flexibility to do whatever they want when the selected bookmark is updated (`onChange(of selectedBookmark) { … }`). I think that would be better than taking a binding to a viewpoint.

dfeinzimer:
Closing; [`onSelectionChanged(perform:)`](https://developers.arcgis.com/swift/toolkit-api-reference/documentation/arcgistoolkit/bookmarks/onselectionchanged(perform:)) is provided

philium:
@dfeinzimer Was there any consideration given to taking a binding to the selection? While there is the modifier that you pointed out, taking a binding to the selected bookmark would be simpler and more idiomatic in SwiftUI. Many SwiftUI-provided modifiers take a selection binding instead of providing a separate modifier to inform when the selection has changed.

dfeinzimer:
Yes - I still think we should do it and I thought we had an issue for it but I'm not seeing it so I'll reopen and re-characterize this issue instead.

dfeinzimer:
Some additional API considerations:

The current initializers are:

`init(isPresented: Binding<Bool>, bookmarks: [Bookmark], viewpoint: Binding<Viewpoint?>?)`
&
`init(isPresented: Binding<Bool>, bookmarks: [Bookmark], viewpoint: Binding<Viewpoint?>?)`

Would we want to add 2 new initializers:
`init(isPresented: Binding<Bool>, bookmarks: [Bookmark], selection: Binding<Bookmark?>?)`
&
`init(isPresented: Binding<Bool>, geoModel: GeoModel, selection: Binding<Bookmark?>?)`

brining us to a total of 4? 4 seems like it would be a lot to me.

philium:
The selection parameter should be non-optional because those initializers should only be used with a selection binding. In my opinion, having four initializers isn't a big deal. Several SwiftUI views have much more than four initializers.

PR #549, opened 2023-12-19
`Bookmarks` - Improvements for 200.4

dfeinzimer:
- Closes 
  - #246
  - #533
- Related 
  - #401
  - #406

- Adds
  - `init(isPresented:bookmarks:selection:)`
  - `init(isPresented:geoModel:selection:)`
- Deprecates
  - `onSelectionChanged(perform:)`

|   | Before | After (with selection made) |
|---|---|---|
| iPhone | iPhone-before | iPhone-after |
| Catalyst | Catalyst-before | Catalyst-after |


Sources/ArcGISToolkit/Components/Bookmarks/Bookmarks.swift
```diff
@@ -78,67 +78,104 @@ public struct Bookmarks: View {
...
 
 extension Bookmarks {
-    /// The list of bookmarks sorted alphabetically.
-    var sortedBookmarks: [Bookmark] {
-        bookmarks.sorted { $0.name <  $1.name }
-    }
-    
     /// Sets an action to perform when the bookmark selection changes.
     /// - Parameter action: The action to perform when the bookmark selection has changed.
+    @available(*, deprecated, message: "Monitor changes on a bound Bookmark or Viewpoint instead.")
```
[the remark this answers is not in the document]
philium: If you mark the test as deprecated, then the deprecation warning should go away.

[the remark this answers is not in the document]
dfeinzimer: Oh cool, I didn't realize a test can be deprecated, thanks.

mhdostal (approved):

Sources/ArcGISToolkit/Components/Bookmarks/BookmarksHeader.swift
```diff
@@ -47,15 +47,15 @@ struct BookmarksHeader: View {
                 maxWidth: .infinity,
                 alignment: .leading
             )
-            if horizontalSizeClass == .compact || verticalSizeClass == .compact {
```
dfeinzimer, line 50: The "Done" button was missing in Catalyst. I think it's okay to always show it.

Sources/ArcGISToolkit/Documentation.docc/Resources/Bookmarks/BookmarksStep1.swift
```diff
@@ -3,7 +3,6 @@ import ArcGISToolkit
 import SwiftUI
 
 struct BookmarksExampleView: View {
-    /// The `Map` with predefined bookmarks.
```
dfeinzimer, line 6: We're not including comments in other component tutorials.

Sources/ArcGISToolkit/Components/Bookmarks/Bookmarks.swift
```diff
@@ -173,46 +222,35 @@ extension Bookmarks {
...
+                        // Make the entire row tappable.
+                        .frame(maxWidth: .infinity, alignment: .leading)
+                        .contentShape(Rectangle())
                 }
+                .buttonStyle(.plain)
+#if targetEnvironment(macCatalyst)
+                .listRowBackground(bookmark == selection?.wrappedValue ? nil : Color.clear)
+#endif
             }
+            .frame(idealWidth: 320, idealHeight: 428)
```
dfeinzimer, line 245: Samples makes [regular use](https://github.com/search?q=repo%3AEsri%2Farcgis-maps-sdk-swift-samples+.frame%28idealWidth%3A+320%2C+idealHeight%3A+428%29&type=code) of this ratio.

dfeinzimer:
@philium This is ready for review. Please review when you have some time available. Thanks!

Sources/ArcGISToolkit/Documentation.docc/Resources/Bookmarks/BookmarksStep2.swift
```diff
@@ -3,27 +3,24 @@ import ArcGISToolkit
...
+    @State private var map = Map(url: URL(string: "https://www.arcgis.com/home/item.html?id=16f1b8ba37b44dc3884afc8d5f454dd2")!)!
     
     var body: some View {
         MapViewReader { mapViewProxy in
             MapView(map: map)
                 .toolbar {
                     ToolbarItem(placement: .primaryAction) {
                         Button {
-                            showingBookmarks.toggle()
+                            bookmarksIsPresented.toggle()
```
philium, line 16:
```suggestion
                            bookmarksIsPresented = true
```

Examples/Examples/BookmarksExampleView.swift
```diff
@@ -17,40 +17,36 @@ import ArcGISToolkit
...
-    /// shown/hidden, whether that be in a group of options or a standalone button.
-    @State var showingBookmarks = false
+    /// The selected bookmark.
+    @State private var selection: Bookmark?
     
     var body: some View {
         MapViewReader { mapViewProxy in
             MapView(map: map)
                 .toolbar {
                     ToolbarItem(placement: .primaryAction) {
```
philium, line 33: Having this as the primary action is what makes it bold, which I'm assuming `buttonStyle(.plain)` undoes. Better to just specify the desired placement (`topBarTrailing`).

Sources/ArcGISToolkit/Documentation.docc/Resources/Bookmarks/BookmarksStep3.swift
```diff
@@ -3,32 +3,31 @@ import ArcGISToolkit
...
+    @State private var selection: Bookmark?
     
     var body: some View {
         MapViewReader { mapViewProxy in
             MapView(map: map)
                 .toolbar {
                     ToolbarItem(placement: .primaryAction) {
                         Button {
-                            showingBookmarks.toggle()
+                            bookmarksIsPresented.toggle()
```
philium, line 18:
```suggestion
                            bookmarksIsPresented = true
```

Sources/ArcGISToolkit/Documentation.docc/Resources/Bookmarks/BookmarksStep4.swift
```diff
@@ -3,34 +3,33 @@ import ArcGISToolkit
...
+    @State private var selection: Bookmark?
     
     var body: some View {
         MapViewReader { mapViewProxy in
             MapView(map: map)
                 .toolbar {
                     ToolbarItem(placement: .primaryAction) {
                         Button {
-                            showingBookmarks.toggle()
+                            bookmarksIsPresented.toggle()
```
philium, line 18:
```suggestion
                            bookmarksIsPresented = true
```

Test Runner/Test Runner/TestViews/BookmarksTestViews/BookmarksTestCase1View.swift
```diff
@@ -22,21 +22,21 @@ struct BookmarksTestCase1View: View {
     @State private var map = Map(url: URL(string: "https://www.arcgis.com/home/item.html?id=16f1b8ba37b44dc3884afc8d5f454dd2")!)!
     
     /// The last selected bookmark.
-    @State private var selectedBookmark: Bookmark?
+    @State private var selection: Bookmark?
```
philium, line 25: I actually think that `selectedBookmark` is a better name for this.

Sources/ArcGISToolkit/Components/Bookmarks/BookmarksHeader.swift
```diff
@@ -47,15 +47,15 @@ struct BookmarksHeader: View {
...
-                Spacer()
-                Button {
-                    isPresented.toggle()
-                } label: {
-                    Text.done
-                        .fontWeight(.semibold)
-                }
+            Spacer()
+            Button {
```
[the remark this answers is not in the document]
dfeinzimer: It is nice to get system styling where we can but I'm not sure I see where I'd fit a `NavigationStack` into the component's hierarchy, especially since it doesn't present another view. Any suggestion?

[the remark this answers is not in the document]
philium: Well, it would have to be at the root of the `Bookmarks` view and the navigation bar would essentially replace the `BookmarksHeader` view. But the styling would be difference; not sure whether that would be a deal breaker.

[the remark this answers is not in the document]
dfeinzimer: I've spent some more time experimenting with incorporating a `NavigationStack`. My feeling is that it's going to take some wrangling to get it integrated nicely. My preference is to leave the header as is but we can wait for Mark to weigh in.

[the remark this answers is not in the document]
rolson: I think If we wanted to chase that further, it should be done separately and not hold up this PR.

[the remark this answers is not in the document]
philium: Agreed. I didn't mean to suggest that it needed to be addressed in this PR. I was just pointing out a potential improvement.

Sources/ArcGISToolkit/Components/Bookmarks/Bookmarks.swift
```diff
@@ -173,46 +222,35 @@ extension Bookmarks {
...
-                    }
+    /// Makes a view that shows a list of bookmarks.
+    /// - Parameter bookmarks: The bookmarks to be shown.
+    @ViewBuilder private func makeList(bookmarks: [Bookmark]) -> some View {
+        if bookmarks.isEmpty {
+            noBookmarks
+        } else {
+            List(bookmarks.sorted { $0.name <  $1.name }, id: \.self, selection: selection) { bookmark in
+                Button {
+                    selectBookmark(bookmark)
```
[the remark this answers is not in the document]
dfeinzimer: I think we _should_ do that down the line when the newly deprecated initializers are finally removed. In the meantime we still need to support them, meaning we need to deal with both cases where there is or isn't a selection provided. 

If we did it now we'd either need a good way to fill in the gap for the deprecated initializers that don't take a selection or maintain two versions of the list, one using the selection parameter and the other not.

I can provide inline doc that calls this out.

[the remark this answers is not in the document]
philium: Ah! Yes, I didn't think about the case where a deprecated initializer is used. Never mind (for now).

Sources/ArcGISToolkit/Components/Bookmarks/Bookmarks.swift
```diff
@@ -77,74 +62,74 @@ public struct Bookmarks: View {
...
     ///   - bookmarks: An array of bookmarks. Use this when displaying bookmarks defined at runtime.
-    ///   - viewpoint: A viewpoint binding that will be updated when a bookmark is selected.
-    ///   Alternately, you can use the `onSelectionChanged(perform:)` modifier to handle
-    ///   bookmark selection.
+    ///   - selection: A selected Bookmark.
+    ///   - geoViewProxy: The proxy to provide access to geo view operations.
+    ///
+    /// When a `GeoViewProxy` is provided, the map or scene will automatically pan and zoom to the
+    /// selected bookmark.
     public init(
```
philium, line 70: If you move all of these into an extension, then the compiler would synthesize a memberwise initializer that the other initializers could delegate to:

image

For example:

```swift
public extension Bookmarks {
    /// ...
    init(
        isPresented: Binding<Bool>,
        bookmarks: [Bookmark],
        selection: Binding<Bookmark?>,
        geoViewProxy: GeoViewProxy? = nil
    ) {
        self.init(
            bookmarkSource: .array(bookmarks),
            geoViewProxy: geoViewProxy,
            isPresented: isPresented,
            selection: selection
        )
    }
    
    ...
}
```

I suggest doing that.

Sources/ArcGISToolkit/Components/Bookmarks/Bookmarks.swift
```diff
@@ -173,41 +170,41 @@ extension Bookmarks {
...
+            .listStyle(.plain)
         }
-        .padding([.horizontal])
     }
     
-    /// A view that is shown while a `GeoModel` is loading.
-    private var loading: some View {
-        ProgressView()
+    /// Makes a view that is shown while a `GeoModel` is loading.
+    private func makeLoadingView(with geoModel: GeoModel) -> some View {
```
philium, line 202: I can't tell if this is something that was actually changed in this PR or not, so I'll just comment on it and leave it up to you whether you want to do anything about it. It isn't a big deal since this is a private API, but according the the Swift API design guidelines, "The first argument to initializer and [factory methods](https://en.wikipedia.org/wiki/Factory_method_pattern) calls should not form a phrase starting with the base name." In other words, `geoModel` should not have a custom argument label (`with`).

Sources/ArcGISToolkit/Components/Bookmarks/Bookmarks.swift
```diff
@@ -230,3 +227,60 @@ extension Bookmarks {
         .padding()
     }
 }
+
+/// Deprecated members
+public extension Bookmarks {
```
philium, line 232: Extensions cannot be documented, so there is no difference between this comment and an equivalent comment with only two slashes. Also, no period. Alternatively, you could do what I suggested, putting it inline with the declaration itself. I like that approach because extensions are often used to group together declarations for a particular purpose, usually related to a protocol and the protocol name goes after the type name in the extension declaration. So it fits with the mental model of looking at the extension declaration itself to see why it is there.

Sources/ArcGISToolkit/Components/Bookmarks/Bookmarks.swift
```diff
@@ -230,3 +227,60 @@ extension Bookmarks {
...
+    init(
+        isPresented: Binding<Bool>,
+        bookmarks: [Bookmark],
+        viewpoint: Binding<Viewpoint?>? = nil
+    ) {
+        self.init(
+            bookmarkSource: .array(bookmarks),
+            geoViewProxy: nil,
+            isPresented: isPresented,
+            selection: nil,
```
philium, line 249: The `selection` parameter has a default value of `nil`.

```suggestion
```

Sources/ArcGISToolkit/Components/Bookmarks/Bookmarks.swift
```diff
@@ -230,3 +227,60 @@ extension Bookmarks {
...
+    init(
+        isPresented: Binding<Bool>,
+        geoModel: GeoModel,
+        viewpoint: Binding<Viewpoint?>? = nil
+    ) {
+        self.init(
+            bookmarkSource: .geoModel(geoModel),
+            geoViewProxy: nil,
+            isPresented: isPresented,
+            selection: nil,
```
philium, line 270:
```suggestion
```

philium (commented):
Just a few more comments.

Sources/ArcGISToolkit/Components/Bookmarks/Bookmarks.swift
dfeinzimer: Ah okay, I've moved it inline as suggested.

I wasn't sure if you were actually suggesting placing it there originally.

philium (approved):
Thank you for making those changes!

mhdostal (approved):

PR #552, opened 2024-01-03
Update world scale AR

des12437:
Updates world scale AR brach with main.


yo1995 (approved):
Verified building correctly, the branch is rebased onto `main`.

PR #553, opened 2024-01-03
State variable updates for events

mhdostal:
This updates how @State variables are used in the input types, based on [this](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/548#pullrequestreview-1791819063) discussion.

I've also removed some of that @State properties that weren't needed.

### DIFF
--- a/Sources/ArcGISToolkit/Components/FormView/FormInputs/DateTimeInput.swift
+++ b/Sources/ArcGISToolkit/Components/FormView/FormInputs/DateTimeInput.swift
@@ -73,6 +73,7 @@ struct DateTimeInput: View {
             requiredValueMissing = isRequired && date == nil
             do {
                 try element.updateValue(date)
+                formattedValue = element.formattedValue
             } catch {
                 print(error.localizedDescription)
             }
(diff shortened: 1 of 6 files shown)


Sources/ArcGISToolkit/Components/FormView/FormInputs/DateTimeInput.swift
```diff
@@ -190,15 +181,6 @@ struct DateTimeInput: View {
...
     }
     
-    /// The human-readable date and time selection.
-    var formattedDate: String? {
-        if input.includeTime {
-            return date?.formatted(.dateTime.day().month().year().hour().minute())
-        } else {
-            return date?.formatted(.dateTime.day().month().year())
-        }
-    }
```
dfeinzimer, line 200: I'm seeing some new failures in our UI tests, specifically 2.3, 2.4 and 2.5. Should the formatted value from core respect `DateTimePickerFormInput.includeTime` for us now?

Before:
Screenshot 2024-01-03 at 14 45 55

After:
Screenshot 2024-01-03 at 14 49 24

mhdostal: I would think so, yes.

dfeinzimer (approved):

PR #556, opened 2024-01-08
AR: Geotracking World Scale

des12437:
Implements world scale AR with `ARGeoTrackingConfiguration`.


Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -0,0 +1,298 @@
...
+                            self.currentLocation = location
+                            if let heading = currentHeading {
+                                await updateSceneView(for: location, heading: heading)
+                            }
+                        }
+                    }
+                    group.addTask {
+                        for await heading in locationDataSource.headings {
+                            self.currentHeading = heading
+                            if let location = self.currentLocation {
```
yo1995, line 149:
```suggestion
                            if let location = currentLocation {
```

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -0,0 +1,298 @@
...
+    private func updateSceneView(for location: Location, heading: Double) {
+        // Do not use cached location more than 10 seconds old.
+        guard abs(lastLocationTimestamp?.timeIntervalSinceNow ?? 0) < 10 else { return }
+        
+        // Make sure that horizontal and vertical accuracy are valid.
+        guard location.horizontalAccuracy > validAccuracyThreshold,
+              location.verticalAccuracy > validAccuracyThreshold else { return }
+        
+        // Make sure either the initial camera is not set, or we need to update the camera.
+        guard (!initialCameraIsSet || shouldUpdateCamera(for: location)) else { return }
```
yo1995, line 185:
```suggestion
        guard !initialCameraIsSet || shouldUpdateCamera(for: location) else { return }
```

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -0,0 +1,298 @@
...
+    /// The camera controller that will be set on the scene view.
+    @State private var cameraController: TransformationMatrixCameraController
+    /// The current interface orientation.
+    @State private var interfaceOrientation: InterfaceOrientation?
+    /// The location datasource that is used to access the device location.
+    @State private var locationDataSource: LocationDataSource
+    /// A Boolean value indicating if the camera was initially set.
+    @State private var initialCameraIsSet = false
+    /// A Boolean value that indicates whether to hide the coaching overlay view.
+    private var coachingOverlayIsHidden: Bool = false
```
yo1995, line 32:
```suggestion
    private var coachingOverlayIsHidden = false
```

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -0,0 +1,298 @@
...
+    /// A Boolean value that indicates whether the coaching overlay view is active.
+    @State private var coachingOverlayIsActive: Bool = true
```
yo1995, line 34:
```suggestion
    @State private var coachingOverlayIsActive = true
```

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -0,0 +1,298 @@
...
+    /// The current camera of the scene view.
+    @State private var currentCamera: Camera?
+    /// A Boolean value that indicates whether the calibration view is hidden.
+    private var calibrationViewIsHidden: Bool = false
```
yo1995, line 38:
```suggestion
    private var calibrationViewIsHidden = false
```

Examples/Examples/WorldScaleGeoTrackingExampleView.swift
```diff
@@ -0,0 +1,101 @@
...
+        scene.addOperationalLayer(parcelsLayer)
+        return scene
+    }()
+    
+    /// Basemap opacity.
+    @State private var opacity: Float = 1
+    /// Graphics overlay to show a graphic around your initial location.
+    @State private var graphicsOverlay = GraphicsOverlay()
+    /// The location datasource that is used to access the device location.
+    @State private var locationDatasSource = SystemLocationDataSource()
```
yo1995, line 44: Should we pass this LDS to the `WorldScaleGeoTrackingSceneView` so that it does not create a second LDS in its default init parameters?

Examples/Examples/WorldScaleGeoTrackingExampleView.swift
```diff
@@ -0,0 +1,101 @@
...
+    
+    static var parcelsLayer: FeatureLayer {
```
yo1995, line 46:
```suggestion
    private static var parcelsLayer: FeatureLayer {
```

Examples/Examples/WorldScaleGeoTrackingExampleView.swift
```diff
@@ -0,0 +1,101 @@
...
+                    .onSingleTapGesture { screen, _ in
+                        print("Identifying...")
+                        Task.detached {
+                            let results = try await proxy.identifyLayers(screenPoint: screen, tolerance: 20)
+                            print("\(results.count) identify result(s).")
+                        }
+                    }
+            }
+            // A slider to adjust the basemap opacity.
+            Slider(value: $opacity, in: 0...1.0)
```
yo1995, line 66:
```suggestion
            Slider(value: $opacity, in: 0...1)
```

Examples/Examples/WorldScaleGeoTrackingExampleView.swift
```diff
@@ -0,0 +1,101 @@
...
+                .padding(.horizontal)
+        }
+        .onChange(of: opacity) { opacity in
+            guard let basemap = scene.basemap else { return }
+            for layer in basemap.baseLayers {
+                layer.opacity = opacity
+            }
```
yo1995, line 73:
```suggestion
            basemap.baseLayers.forEach { $0.opacity = opacity }
```
up to you

Examples/Examples/WorldScaleGeoTrackingExampleView.swift
```diff
@@ -0,0 +1,101 @@
...
+        featureLayer.renderer = SimpleRenderer(symbol: SimpleLineSymbol(color: .cyan, width: 3))
+        return featureLayer
+    }
+    
+    var body: some View {
+        VStack {
+            WorldScaleGeoTrackingSceneView { proxy in
+                SceneView(scene: scene, graphicsOverlays: [graphicsOverlay])
+                    .onSingleTapGesture { screen, _ in
+                        print("Identifying...")
```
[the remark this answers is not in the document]
des12437: `TableTopExampleView` and `FlyoverExampleView` contain print statements after the identify operation as well.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -0,0 +1,298 @@
...
+        
+        // Add some of the vertical accuracy to the z value of the position, that way if the
+        // GPS location is not accurate, we won't end up below the earth's surface.
+        let altitude = (location.position.z ?? 0) + location.verticalAccuracy
+        
+        cameraController.originCamera = Camera(
+            latitude: location.position.y,
+            longitude: location.position.x,
+            altitude: altitude,
+            heading: calibrationHeading ?? heading,
```
yo1995, line 195: We may adjust the heading logic here when the calibration view is implemented.

- Currently the `updateSceneView` method is called whenever a location or heading update comes from the location data source
- If the heading is manually adjusted by the user, it is no longer "connected" to the data source, and stay as the custom value forever
- If the heading is not adjusted, it will be connected to the heading updates from the LDS

I think for the `initial` mode, having the heading fixed after receiving the first update, and let the user to adjust make sense. However, for `continuous` mode, we should discuss what is the desired behavior.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -0,0 +1,298 @@
...
+        
+        // We have to do this or the error gets bigger and bigger.
+        cameraController.transformationMatrix = .identity
+        arViewProxy.session.run(configuration, options: .resetTracking)
+        
+        // If initial camera is not set, then we set it the flag here to true
+        // and set the status text to empty.
+        if !initialCameraIsSet {
+            coachingOverlayIsActive = false
+            withAnimation {
```
yo1995, line 208: In my tests I'm not able to see the animated fade-in transition for the scene view - basemap loading takes some time. Can you see the animation on iPad? If not I think we may remove the animation.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -0,0 +1,298 @@
...
+    func shouldUpdateCamera(for location: Location) -> Bool {
+        // Do not update unless the horizontal accuracy is less than a threshold.
+        guard let currentCamera,
+              let spatialReference = currentCamera.location.spatialReference,
+              let currentPosition = GeometryEngine.project(location.position, into: spatialReference)
+        else { return false }
+        
+        // Measure the distance between the location datasource's reported location
+        // and the camera's current location.
+        guard let result = GeometryEngine.geodeticDistance(
```
yo1995, line 225: I'm a bit wary of the geodetic calculation's performance. The `updateSceneView` method is called frequently. The most time consuming part in it is the construction of a new `Camera` object. While the time consumers in `shouldUpdateCamera` are the 2 `GeometryEngine` class methods.

We can probably run a performance test for both methods in release config, and see which takes longer. 🤔 Just a thought, I guess the time difference here is negligible compared to the frame rendering.

des12437: Good point. The logic to update the heading can be removed from `updateSceneView` so that the method is called less frequently since the geodesic calculation does not use the heading value.

rolson (commented):
I skimmed this PR and it looks ok. But I need to spend some more time looking at it and testing it. That won't happen until probably next week unfortunately. If you need somebody to review it before then, you can ask for somebody else to review it. Otherwise I'll look at it then. Thanks

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -0,0 +1,293 @@
...
+                            currentLocation = location
+                            await updateSceneView(for: location)
+                        }
+                    }
+                    group.addTask {
+                        for await heading in locationDataSource.headings {
+                            currentHeading = heading
+                            updateHeading(heading)
+                        }
+                    }
```
yo1995, line 149: I noticed that with continuous heading updates, the scene view doesn't yaw smoothly when the device is slowly turning. We'll revisit this problem in the following PR where the continuous/initial modes are added.

yo1995 (dismissed):
LGTM. Let's work on the other details in the following PRs.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
des12437: This issue should be fixed now by animating the heading changes.

rolson:
adding @zkline101 as he expressed interest in looking at it as it relates to some of the other work he is doing. I can still review it next week if needed.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -0,0 +1,300 @@
...
+                                lastLocationTimestamp = location.timestamp
+                                currentLocation = location
+                                await updateSceneView(for: location)
+                            }
+                        }
+                        group.addTask {
+                            for await heading in locationDataSource.headings {
+                                currentHeading = heading
+                                
+                                guard let currentCamera else { return }
```
yo1995, line 147: This line causes the task to halt - if you put a break point in this task, you'll see it is only run once.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -0,0 +1,300 @@
...
+                                let camera = cameraController.originCamera.rotatedTo(
+                                    heading: calibrationHeading ?? heading,
+                                    pitch: currentCamera.pitch,
+                                    roll: currentCamera.roll
+                                )
+                                sceneViewProxy.setViewpointCamera(camera)
```
yo1995, line 153: After making the heading updates come through, in my testing the `setViewpointCamera` method is interfering with the frame update. We can discuss and test different ideas of updating the camera smoothly.

I think the questions we need to figure out are…

1. In the past we only use this method to adjust camera when **not** using `ARKit`. That is not the case for our new design right? If we always use ARKit now, then it already does the tracking for us.
2. When using ARKit, it obtains the location and heading from the device's GPS. If we instead using a different LDS such as NMEALDS, then it may make sense to align the location and heading between ARKit and the custom LDS.
3. For initial mode, once the initial location and heading are set, and the session starts, later heading adjustments rely on the calibration view.

So IMO we can leave this out for now, and figure it out once we have a clear separation between `initial` and `continuous` modes.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -0,0 +1,300 @@
...
+                
+                if !coachingOverlayIsHidden {
+                    ARCoachingOverlay(goal: .geoTracking)
+                        .sessionProvider(arViewProxy)
+                        .active(coachingOverlayIsActive)
+                        .allowsHitTesting(false)
+                }
+            }
+            .observingInterfaceOrientation($interfaceOrientation)
+            .onAppear {
```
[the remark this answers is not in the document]
des12437: This change was made to expand the scope of the `sceneViewProxy`.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
des12437: Okay, I will revert the last commit that uses `setViewpointCamera` and it can be added later if necessary.

yo1995 (approved):

Examples/ExamplesApp/Info.plist
```diff
@@ -59,6 +59,8 @@
 	<string>1</string>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
+	<key>NSCameraUsageDescription</key>
+	<string>This app uses augmented reality to overlay imagery over your real-world environment. Camera access is required for this functionality.</string>
```
zkline101, line 63:
```suggestion
	<string>This app uses your camera for augmented reality that overlays imagery over your real-world environment.</string>
```

Examples/Examples/WorldScaleGeoTrackingExampleView.swift
```diff
@@ -0,0 +1,92 @@
...
+        surface.navigationConstraint = .unconstrained
+        let scene = Scene(basemapStyle: .arcGISImagery)
+        scene.baseSurface = surface
+        scene.addOperationalLayer(parcelsLayer)
+        return scene
+    }()
+    
+    /// Basemap opacity.
+    @State private var opacity: Float = 1
+    /// Graphics overlay to show a graphic around your initial location.
```
zkline101, line 40:
```suggestion
    /// The graphics overlay which shows a graphic around your initial location.
```

Examples/Examples/WorldScaleGeoTrackingExampleView.swift
```diff
@@ -0,0 +1,92 @@
...
+        surface.addElevationSource(elevationSource)
+        surface.backgroundGrid.isVisible = false
+        surface.navigationConstraint = .unconstrained
+        let scene = Scene(basemapStyle: .arcGISImagery)
+        scene.baseSurface = surface
+        scene.addOperationalLayer(parcelsLayer)
+        return scene
+    }()
+    
+    /// Basemap opacity.
```
zkline101, line 38:
```suggestion
    /// The basemap opacity.
```

Examples/Examples/WorldScaleGeoTrackingExampleView.swift
```diff
@@ -0,0 +1,92 @@
...
+    @State private var opacity: Float = 1
+    /// Graphics overlay to show a graphic around your initial location.
+    @State private var graphicsOverlay = GraphicsOverlay()
+    /// The location datasource that is used to access the device location.
+    @State private var locationDataSource = SystemLocationDataSource()
+    /// A feature layer with San Bernardino parcels data.
+    private static var parcelsLayer: FeatureLayer {
```
zkline101, line 45: This might be better suited on under a feature layer extension or just declare everything in the one spot it is being used.

Examples/Examples/WorldScaleGeoTrackingExampleView.swift
```diff
@@ -0,0 +1,92 @@
...
+    var body: some View {
+        VStack {
+            WorldScaleGeoTrackingSceneView(locationDataSource: locationDataSource) { proxy in
+                SceneView(scene: scene, graphicsOverlays: [graphicsOverlay])
+                    .onSingleTapGesture { screen, _ in
+                        print("Identifying...")
+                        Task {
+                            let results = try await proxy.identifyLayers(screenPoint: screen, tolerance: 20)
+                            print("\(results.count) identify result(s).")
+                        }
```
zkline101, line 61: This will be better in a task view modifier that gets kicked off every time the screen is tapped. That is because if the user taps the screen very fast multiple times the task modifier will cancel the task before it. Here a bunch of tasks will get kicked off regardless if another one started.

zkline101 (commented):
Testing it now but here are some intial comments, nothing major looks good.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -0,0 +1,295 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ARKit
+import SwiftUI
+import ArcGIS
+
+/// A scene view that provides an augmented reality world scale experience using geotracking.
+public struct WorldScaleGeoTrackingSceneView: View {
```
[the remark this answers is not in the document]
des12437: `TableTopSceneView` and `FlyoverSceneView` have the same structure so I would rather leave the data on the view.

Examples/Examples/WorldScaleGeoTrackingExampleView.swift
des12437: I'm not sure how a task modifier can be used with the onSingleTapGesture view modifier. Can you clarify this feedback? Thanks.

zkline101: You could possibly do it like this (I have not tested it):

```swift
@State private var tappedScreenPoint: CGPoint?

SceneView(scene: scene, graphicsOverlays: [graphicsOverlay])
    .onSingleTapGesture { screenPoint, _ in
       tappedScreenPoint = screenPoint
    }
    .task(id: tappedScreenPoint) {
        guard let tappedScreenPoint else { return }
        
        let results = try? await proxy.identifyLayers(screenPoint: tappedScreenPoint, tolerance: 20)
        print("\(results?.count ?? 0) identify result(s).")
    } as! SceneView
```

But I'm not sure if we should advertise that force casting to a SceneView to our users.

des12437: I tested this code and it results in a crash when casting the task modifier to a scene view. The same logic is in `TableTopExampleView` and `FlyoverExampleView`, so maybe this logic can be refactored later and applied to all examples.

zkline101: Darn

zkline101 (dismissed):

Examples/Examples/WorldScaleGeoTrackingExampleView.swift
```diff
@@ -90,3 +83,13 @@ struct WorldScaleGeoTrackingExampleView: View {
         }
     }
 }
+
+private extension Layer {
+    // A feature layer with San Bernardino parcels data.
```
yo1995, line 88:
```suggestion
    /// A feature layer with San Bernardino parcels data.
```

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -0,0 +1,300 @@
...
+                        group.addTask {
+                            for await heading in locationDataSource.headings {
+                                currentHeading = heading
+                                
+                                guard let currentCamera else { return }
+                                let camera = cameraController.originCamera.rotatedTo(
+                                    heading: calibrationHeading ?? heading,
+                                    pitch: currentCamera.pitch,
+                                    roll: currentCamera.roll
+                                )
```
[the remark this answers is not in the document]
yo1995: Sorry if I misled you here - as discussed with Mark, we don't need to set the heading when using ARKit - it constantly align the AR view with device heading with `gravityAndHeading` alignment.

Manual heading update would only be in these situations:

- Not using ARKit, which isn't supported in our new design
- Manually setting scene view heading in `local` mode. When the initial location is set by ARKit and then  stopped, we use the buttons to adjust heading in origin camera. We probably don't need the device heading update in this case either.

So I think the code can be omitted from this PR. Otherwise with current implementation the example's scene view will rotate twice the angle of the device rotation, causing the scene view to not align with the device camera.

```swift
group.addTask {
    for await heading in locationDataSource.headings {
        currentHeading = heading
        updateHeading(heading)
    }
}
```

yo1995 (commented):
LGTM with 2 minor things

yo1995 (approved):

zkline101 (approved):

PR #557, opened 2024-01-10
`FormView` - Adopt new validation API

dfeinzimer:
Apollo 356


UI Test Runner/UITests/FormViewTests.swift
```diff
@@ -149,8 +149,8 @@ final class FormViewTests: XCTestCase {
         )
         
         XCTAssertTrue(
-            characterIndicator.isHittable,
-            "The character count isn't hittable."
+            characterIndicator.exists,
+            "The character count doesn't exist."
```
dfeinzimer, line 153: This check just seemed to be flaky.

Sources/ArcGISToolkit/Components/FormView/FormInputs/InputFooter.swift
```diff
@@ -17,23 +17,204 @@ import SwiftUI
...
+            return elementErrors?.first(where: { $0 != .featureFormFieldIsRequiredError })
+        }
+    }
+}
+
+private extension RangeDomain {
+    /// String representations of the minimum and maximum value of the range domain.
+    var displayableMinAndMax: (min: String, max: String)? {
+        if let min = minValue as? Double, let max = maxValue as? Double {
+            return (min.formatted(.number.precision(.fractionLength(1...3))), max.formatted(.number.precision(.fractionLength(1...3))))
```
dfeinzimer, line 211: Does this range for precision seem okay?

Sources/ArcGISToolkit/Components/FormView/FormInputs/TextInput.swift
```diff
@@ -100,9 +95,22 @@ struct TextInput: View {
...
+                    try element.updateValue(Int32(text))
+                case .int64:
+                    try element.updateValue(Int64(text))
+                case .float32:
+                    try element.updateValue(Float64(text))
+                case .float64:
+                    try element.updateValue(Float64(text))
+                default:
+                    try element.updateValue(text)
+                }
```
dfeinzimer, line 111: Need to verify that casting type is a requirement again.

Sources/ArcGISToolkit/Components/FormView/FormInputs/InputFooter.swift
```diff
@@ -17,23 +17,234 @@ import SwiftUI
...
+            "Minimum \(lengthRange!.lowerBound) characters",
+            bundle: .toolkitModule,
+            comment: "Text indicating a field's minimum number of allowed characters."
+        )
+    }
+    
+    /// The error to display for the input. If the element is not focused and has a required error this will be
+    /// the primary error. Otherwise the primary error is the first error in the element's set of errors.
+    var primaryError: ArcGIS.FeatureFormError? {
+        let elementErrors = model.validationErrors[element.fieldName] as? [ArcGIS.FeatureFormError]
```
mhdostal, line 205: Since `primaryError` is computed and not a `@State` or `@Published`, will the UI update if it changes? `element.validationErrors` is never accessed, we're just relying on `featureForm.validationErrors`, which is somewhat inefficient, as that goes through all elements.

I'm wondering if we can make the errors a `@State` property and localize them to the individual elements. Or make `primaryError` a `@State` object which contains the error to display, which is then passed into the `InputFooter`.

So after `model.evaluateExpressions()` is called in each of the input types, the `element.validationErrors` are retrieved and used to create `primaryError` (or another `@State` property).

I want to make sure that the error display is updated in response to validation errors and we're not always getting all validation errors from the feature form.

Also, I had envisioned `featureForm.evaluationErrors` to be only called in response to a `submit` action.

The one question I have is how does that work for other elements whose values are updated in response to expression evaluation, but then have validation errors? Are we showing those errors (the elements will never get focus, as they are non-editable)?

mhdostal (commented):
Overall this looks really good. I like how you simplified the input footer and removed a bunch of code; I think it's a lot more readable.

Just a few comments and the one bigger question about validation errors.

Sources/ArcGISToolkit/Components/FormView/FormInputs/InputFooter.swift
```diff
@@ -17,23 +17,234 @@ import SwiftUI
...
-                Text.required
-            } else {
-                Text(element.description)
+        HStack(alignment: .top) {
+            Group {
+                if isShowingError {
+                    errorMessage
+                } else if !element.description.isEmpty {
+                    Text(element.description)
+                } else if element.fieldType == .text && model.focusedElement == element {
```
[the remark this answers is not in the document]
dfeinzimer: I think I added support while you were reviewing. 🙂 Pull latest and this should be handled.

Sources/ArcGISToolkit/Components/FormView/FormInputs/InputFooter.swift
dfeinzimer: > What are you trying to accomplish?

The goal is to formate Double based RangeDomains with trailing zeros even if there was no fractional portion (e.g. minimum of 3 shows as 3.0).

We have a UI test that [expects this](https://devtopia.esri.com/runtime/common-toolkit/blob/main/designs/Forms/FormsTestDesign.md#test-case-14-numeric-text-input:~:text=the%20text%20field-,Expectation%3A%20the%20helper%20text%20now%20says%20%22Enter%20value%20from%202.0%20to%205.0%22,-clear%20the%20current).

I think I've found the right solution which is to use [PartialRangeFrom](https://developer.apple.com/documentation/swift/partialrangefrom) starting at one `1...` which from my experiment seems like it will append anything from a `.0` up to the level of accuracy stored in the Double.

```swift
min.formatted(.number.precision(.fractionLength(1...))
```

Sources/ArcGISToolkit/Components/FormView/FormInputs/InputFooter.swift
dfeinzimer: > Since primaryError is computed and not a `@State` or `@Published`, will the UI update if it changes? 

Yes, in my testing the UI changes correctly as the validation errors change. Since `InputFooter` was redrawing whenever the `FormViewModel` published changes (which was on every character input), `featureForm.validationErrors` was also being hit and re-accessed each time.

> element.validationErrors is never accessed, we're just relying on featureForm.validationErrors, which is somewhat inefficient, as that goes through all elements. ... I want to make sure that the error display is updated in response to validation errors and we're not always getting all validation errors from the feature form. ... Also, I had envisioned featureForm.evaluationErrors to be only called in response to a submit action.

👍  I've changed this branch to access `element.validationErrors` instead.

> The one question I have is how does that work for other elements whose values are updated in response to expression evaluation, but then have validation errors? Are we showing those errors (the elements will never get focus, as they are non-editable)?

Following Kaushik's validation algorithm, non-editable fields will never show validation errors. Every other field will show validation errors either once it has or had focus or the force validation flag is set.

> I'm wondering if we can make the errors a `@State` property and localize them to the individual elements. ... So after model.evaluateExpressions() is called in each of the input types, the element.validationErrors are retrieved and used to create primaryError (or another @State property).

I've implemented this in another branch ([df/NewValidationAPI-alt1](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/tree/df/NewValidationAPI-alt1)). The  diff between the two branches is [here](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/compare/df/NewValidationAPI...df/NewValidationAPI-alt1). Code-wise this implementation seems more brittle and there's a lot of repetitive conditional casting to (`element.validationErrors as? [FeatureFormError] ?? []`). Efficiency-wise `element.validationErrors` is hit a much less often but I'm not sure if that outweighs the downside of having all of the repetitive code in each input. Thoughts?

mhdostal: Ok, thanks for looking into that. It's kind of messy having to do the casting and the duplicate code. So keep what you have.

mhdostal (approved):

PR #558, opened 2024-01-17
`FormView` - ValidationErrorVisibility

dfeinzimer:
Apollo 377
Common Toolkit 232


Examples/Examples/FormViewExampleView.swift
```diff
@@ -133,7 +134,12 @@ private extension URL {
 }
 
 /// The model class for the form example view
-@MainActor class Model: ObservableObject {
+@MainActor
+class Model: ObservableObject {
+    /// A Boolean value indicating whether or not to display all validation errors in the form, irrespective of
+    /// whether a given field has received user-focus.
+    @Published var validationIsForced = false
```
philium, line 141: If this is not intended to be set from outside of the model, then it should be `private(set)`. However, this is set inside of `submitChanges()` before any other work is done. That method is called by the view and that is the only place that this property is used by the model. This could be a state property on the view, the view could update the value when the Submit button is tapped/pressed, and then the model wouldn't even need to know about it. I see zero reason why this should be a property of the model.

Examples/Examples/FormViewExampleView.swift
```diff
@@ -160,6 +166,10 @@ private extension URL {
             return
         }
         
+        validationIsForced = true
+        
+        guard featureForm?.validationErrors.isEmpty ?? true else { return }
```
philium, line 171: The feature form is already unwrapped on line `161` to get the feature. Instead of unwrapping it a second time, store the unwrapped feature form the first time it is unwrapped:

```swift
func submitChanges() async {
    guard let featureForm = featureForm,
          let table = featureForm.feature.table as? ServiceFeatureTable,
          table.isEditable,
          let database = table.serviceGeodatabase else {
        print("A precondition to submit the changes wasn't met.")
        return
    }
    
    ...
    
    self.featureForm = nil
}
```

Examples/Examples/FeatureFormExampleView.swift
```diff
@@ -153,15 +160,17 @@ class Model: ObservableObject {
...
+        guard let featureForm = featureForm,
+              let table = featureForm.feature.table as? ServiceFeatureTable,
               table.isEditable,
               let database = table.serviceGeodatabase else {
             print("A precondition to submit the changes wasn't met.")
             return
         }
         
-        try? await table.update(feature)
+        guard featureForm.validationErrors.isEmpty else { return }
```
mhdostal, line 171: For the future, we should consider showing an alert specifying how many validation errors there are, similar to FieldMaps:

image

mhdostal (approved):

Examples/Examples/FeatureFormExampleView.swift
dfeinzimer: Created Apollo 388

PR #571, opened 2024-01-23
AR: Add calibration view

des12437:
Closes `swift/4516`

Screenshots:

iPhone:

Calibrate button:

Calibration view:

iPad:

Calibrate button:

Calibration view:

### DIFF
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
@@ -150,7 +150,9 @@ private struct JoystickSliderView: View {
             if !editingChanged {
                 timer?.invalidate()
                 timer = nil
-                value = 0.0
+                withAnimation {
+                    value = 0.0
+                }
             }
         }
         .onChange(of: value) { value in
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
@@ -119,7 +119,8 @@ extension WorldScaleGeoTrackingSceneView {
                         setBasemapOpacity(0)
                     }
                 }
-                .frame(minWidth: 250, maxWidth: UIScreen.main.bounds.width/3)
+                .frame(idealWidth: UIScreen.main.bounds.width/3)
+                .fractionPresentationDetents(0.25)
                 .padding()
             }
             .esriBorder()
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
@@ -120,9 +120,13 @@ extension WorldScaleGeoTrackingSceneView {
                     Button {
                         viewModel.isCalibrating = false
                     } label: {
-                        Image(systemName: "xmark.circle")
+                        Image(systemName: "xmark.circle.fill")
+                            .font(.system(size: 20))
+                            .symbolRenderingMode(.hierarchical)
+                            .foregroundStyle(.secondary)
                             .imageScale(.large)
                     }
+                    .buttonStyle(.plain)
                     .frame(alignment: .topTrailing)
                     .padding()
                 }
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
@@ -23,8 +23,6 @@ extension WorldScaleGeoTrackingSceneView {
         @Binding var heading: Double
         /// The camera controller elevation.
         @Binding var elevation: Double
-        /// The calibrated elevation delta.
-        @Binding var elevationDelta: Double
         /// A Boolean value that indicates if the user is calibrating.
         @Binding var isCalibrating: Bool
         
@@ -92,14 +90,14 @@ extension WorldScaleGeoTrackingSceneView {
                             Spacer()
                         }
                     } onIncrement: {
-                        elevationDelta += 1
+                        elevation += 1
                     } onDecrement: {
-                        elevationDelta -= 1
+                        elevation -= 1
                     }
                 }
                 JoystickSliderView()
                     .onSliderDeltaValueChanged { delta in
-                        elevationDelta = delta
+                        elevation += delta
                     }
             }
         }
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
@@ -44,7 +44,7 @@ extension WorldScaleGeoTrackingSceneView {
             .padding()
             .background(.regularMaterial)
             .clipShape(RoundedRectangle(cornerRadius: 15))
-            .frame(maxWidth: 350)
+            .frame(maxWidth: 414)
             .padding()
         }
         
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
@@ -76,7 +76,7 @@ extension WorldScaleGeoTrackingSceneView {
                     }
                 }
                 VStack {
-                    Text("Elevation: \(viewModel.calibrationElevation?.rounded(.towardZero) ?? viewModel.cameraController.originCamera.location.z?.rounded(.towardZero) ?? 0, format: .number)")
+                    Text("Elevation: \(viewModel.calibrationElevation?.rounded(.towardZero) ?? viewModel.cameraController.originCamera.location.z?.rounded(.towardZero) ?? 0, format: .number) m")
                     HStack {
                         Button {
                             updateElevation(by: -1)
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
@@ -107,7 +107,6 @@ extension WorldScaleGeoTrackingSceneView {
         @ViewBuilder
         var dismissButton: some View {
             Button {
-                viewModel.setBasemapOpacity(0)
                 withAnimation {
                     viewModel.isCalibrating = false
                 }
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
@@ -129,7 +129,7 @@ extension WorldScaleGeoTrackingSceneView {
                 .padding(.trailing, -20)
                 .padding(.top, -30)
             }
-            .frame(maxWidth: 300)
+            .frame(maxWidth: 350)
             .padding()
             .padding(.top, 10)
             .background(.regularMaterial)
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
@@ -120,14 +120,18 @@ extension WorldScaleGeoTrackingSceneView {
                     Button {
                         viewModel.isCalibrating = false
                     } label: {
-                        Text("Done")
+                        Image(systemName: "xmark.circle")
+                            .imageScale(.large)
                     }
                     .frame(alignment: .topTrailing)
-                    
+                    .padding()
                 }
+                .padding([.trailing], -20)
+                .padding([.top,], -30)
             }
             .frame(maxWidth: .infinity)
             .padding()
+            .padding([.top], 10)
             .background(.regularMaterial)
         }
         
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
@@ -1,0 +1,190 @@
+// Copyright 2024 Esri
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ARKit
+import SwiftUI
+import ArcGIS
+
+extension WorldScaleGeoTrackingSceneView {
+    /// A view that allows the user to calibrate the heading of the scene view camera controller.
+    struct CalibrationView: View {
+        let scene: ArcGIS.Scene
+        /// The camera controller that will be set on the scene view.
+        @State private var cameraController: TransformationMatrixCameraController
+        /// A Boolean value that indicates if the AR experince is being calibrated.
+        @State private var isCalibrating = false
+        /// The calibrated camera heading.
+        @State private var calibrationHeading: Double?
+        /// The calibrated camera elevation.
+        @State private var calibrationElevation: Double?
+        /// The slider value for the camera heading.
+        @State private var headingSliderValue = 0.0
+        /// The slider value for the camera elevation.
+        @State private var elevationSliderValue = 0.0
+        /// The elevation timer for the "joystick" behavior.
+        @State private var headingTimer: Timer?
+        /// The heading timer for the "joystick" behavior.
+        @State private var elevationTimer: Timer?
+        /// The heading delta amount based on the heading slider value.
+        private var joystickHeadingDelta: Double {
+            Double(signOf: headingSliderValue, magnitudeOf: headingSliderValue * headingSliderValue / 25)
+        }
+        /// The elevation delta amount based on the elevation slider value.
+        private var joystickElevationDelta: Double {
+            Double(signOf: elevationSliderValue, magnitudeOf: elevationSliderValue * elevationSliderValue / 100)
+        }
+        
+        init(
+            scene: ArcGIS.Scene,
+            cameraController: TransformationMatrixCameraController,
+            isCalibrating: Bool,
+            calibrationHeading: Double?,
+            calibrationElevation: Double?
+        ) {
+            self.scene = scene
+            self.cameraController = cameraController
+            self.isCalibrating = isCalibrating
+            self.calibrationHeading = calibrationHeading
+            self.calibrationElevation = calibrationElevation
+        }
+        
+        var body: some View {
+            Button {
+                isCalibrating = true
+                setBasemapOpacity(0.5)
+            } label: {
+                Text("Calibrate")
+            }
+            .popover(isPresented: $isCalibrating) {
+                VStack {
+                    Text("Heading: \(calibrationHeading?.rounded(.towardZero) ?? cameraController.originCamera.heading.rounded(.towardZero), format: .number)")
+                    
+                    HStack {
+                        Button {
+                            let heading = cameraController.originCamera.heading - 1
+                            updateHeading(heading)
+                        } label: {
+                            Image(systemName: "minus")
+                        }
+                        Slider(value: $headingSliderValue, in: -10...10) { editingChanged in
+                            if !editingChanged {
+                                headingTimer?.invalidate()
+                                headingTimer = nil
+                                headingSliderValue = 0.0
+                            }
+                        }
+                        .onChange(of: headingSliderValue) { heading in
+                            
+                            guard headingTimer == nil else { return }
+                            // Create a timer which rotates the camera when fired.
+                            let timer = Timer(timeInterval: 0.1, repeats: true) { [self] (_) in
+                                rotateHeading(joystickHeadingDelta)
+                            }
+                            headingTimer = timer
+                            // Add the timer to the main run loop.
+                            RunLoop.main.add(timer, forMode: .default)
+                        }
+                        
+                        Button {
+                            let heading = cameraController.originCamera.heading + 1
+                            updateHeading(heading)
+                        } label: {
+                            Image(systemName: "plus")
+                        }
+                    }
+                    VStack {
+                        Text("Elevation: \(calibrationElevation?.rounded(.towardZero) ?? cameraController.originCamera.location.z?.rounded(.towardZero) ?? 0, format: .number)")
+                        HStack {
+                            Button {
+                                updateElevation(-1)
+                            } label: {
+                                Image(systemName: "minus")
+                            }
+                            Slider(value: $elevationSliderValue, in: -20...20) { editingChanged in
+                                if !editingChanged {
+                                    elevationTimer?.invalidate()
+                                    elevationTimer = nil
+                                    elevationSliderValue = 0.0
+                                }
+                            }
+                            .onChange(of: elevationSliderValue) { elevation in
+                                
+                                guard elevationTimer == nil else { return }
+                                // Create a timer which rotates the camera when fired.
+                                let timer = Timer(timeInterval: 0.1, repeats: true) { [self] (_) in
+                                    updateElevation(joystickElevationDelta)
+                                }
+                                elevationTimer = timer
+                                // Add the timer to the main run loop.
+                                RunLoop.main.add(timer, forMode: .default)
+                            }
+                            Button {
+                                updateElevation(1)
+                            } label: {
+                                Image(systemName: "plus")
+                            }
+                        }
+                    }
+                }
+                .onDisappear {
+                    guard !isCalibrating else { return }
+                    withAnimation(.easeInOut) {
+                        setBasemapOpacity(0)
+                    }
+                }
+                .frame(minWidth: 200, maxHeight: /*@START_MENU_TOKEN@*/.infinity/*@END_MENU_TOKEN@*/)
+                .padding()
+            }
+        }
+        
+        /// Sets the basemap base layers with the given opacity.
+        /// - Parameter opacity: The opacity of the layer.
+        private func setBasemapOpacity(_ opacity: Float) {
+            guard let basemap = scene.basemap else { return }
+            basemap.baseLayers.forEach { $0.opacity = opacity }
+        }
+        
+        /// Updates the heading of the scene view camera controller.
+        /// - Parameter heading: The camera heading.
+        private func updateHeading(_ heading: Double) {
+            cameraController.originCamera = cameraController.originCamera.rotatedTo(
+                heading: heading,
+                pitch: cameraController.originCamera.pitch,
+                roll: cameraController.originCamera.roll
+            )
+            calibrationHeading = heading
+        }
+        
+        /// Rotates the heading of the scene view camera controller by the heading delta in degrees.
+        /// - Parameter headingDelta: The heading delta in degrees.
+        private func rotateHeading(_ headingDelta: Double) {
+            let newHeading = cameraController.originCamera.heading + headingDelta
+            cameraController.originCamera = cameraController.originCamera.rotatedTo(
+                heading: newHeading,
+                pitch: cameraController.originCamera.pitch,
+                roll: cameraController.originCamera.roll
+            )
+            calibrationHeading = newHeading
+        }
+        
+        /// Elevates the scene view camera controller by the elevation delta.
+        /// - Parameter elevationDelta: The elevation delta.
+        private func updateElevation(_ elevationDelta: Double) {
+            cameraController.originCamera = cameraController.originCamera.elevated(by: elevationDelta)
+            if let elevation = cameraController.originCamera.location.z {
+                calibrationElevation = elevation
+            }
+        }
+    }
+}
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
@@ -44,9 +44,10 @@ extension WorldScaleGeoTrackingSceneView {
                 HStack {
                     Button {
                         let heading = viewModel.cameraController.originCamera.heading - 1
-                        updateHeading(heading)
+                        updateHeading(to: heading)
                     } label: {
                         Image(systemName: "minus")
+                            .imageScale(.large)
                     }
                     Slider(value: $headingSliderValue, in: -10...10) { editingChanged in
                         if !editingChanged {
@@ -58,8 +59,8 @@ extension WorldScaleGeoTrackingSceneView {
                     .onChange(of: headingSliderValue) { heading in
                         guard headingTimer == nil else { return }
                         // Create a timer which rotates the camera when fired.
-                        let timer = Timer(timeInterval: 0.1, repeats: true) { [self] (_) in
-                            rotateHeading(joystickHeadingDelta)
+                        let timer = Timer(timeInterval: 0.1, repeats: true) { _ in
+                            rotateHeading(by: joystickHeadingDelta)
                         }
                         headingTimer = timer
                         // Add the timer to the main run loop.
@@ -68,18 +69,20 @@ extension WorldScaleGeoTrackingSceneView {
                     
                     Button {
                         let heading = viewModel.cameraController.originCamera.heading + 1
-                        updateHeading(heading)
+                        updateHeading(to: heading)
                     } label: {
                         Image(systemName: "plus")
+                            .imageScale(.large)
                     }
                 }
                 VStack {
                     Text("Elevation: \(viewModel.calibrationElevation?.rounded(.towardZero) ?? viewModel.cameraController.originCamera.location.z?.rounded(.towardZero) ?? 0, format: .number)")
                     HStack {
                         Button {
-                            updateElevation(-1)
+                            updateElevation(by: -1)
                         } label: {
                             Image(systemName: "minus")
+                                .imageScale(.large)
                         }
                         Slider(value: $elevationSliderValue, in: -20...20) { editingChanged in
                             if !editingChanged {
@@ -91,17 +94,18 @@ extension WorldScaleGeoTrackingSceneView {
                         .onChange(of: elevationSliderValue) { elevation in
                             guard elevationTimer == nil else { return }
                             // Create a timer which rotates the camera when fired.
-                            let timer = Timer(timeInterval: 0.1, repeats: true) { [self] (_) in
-                                updateElevation(joystickElevationDelta)
+                            let timer = Timer(timeInterval: 0.1, repeats: true) { _ in
+                                updateElevation(by: joystickElevationDelta)
                             }
                             elevationTimer = timer
                             // Add the timer to the main run loop.
                             RunLoop.main.add(timer, forMode: .default)
                         }
                         Button {
-                            updateElevation(1)
+                            updateElevation(by: 1)
                         } label: {
                             Image(systemName: "plus")
+                                .imageScale(.large)
                         }
                     }
                 }
@@ -129,7 +133,7 @@ extension WorldScaleGeoTrackingSceneView {
         
         /// Updates the heading of the scene view camera controller.
         /// - Parameter heading: The camera heading.
-        private func updateHeading(_ heading: Double) {
+        private func updateHeading(to heading: Double) {
             viewModel.cameraController.originCamera = viewModel.cameraController.originCamera.rotatedTo(
                 heading: heading,
                 pitch: viewModel.cameraController.originCamera.pitch,
@@ -140,7 +144,7 @@ extension WorldScaleGeoTrackingSceneView {
         
         /// Rotates the heading of the scene view camera controller by the heading delta in degrees.
         /// - Parameter headingDelta: The heading delta in degrees.
-        private func rotateHeading(_ headingDelta: Double) {
+        private func rotateHeading(by headingDelta: Double) {
             let newHeading = viewModel.cameraController.originCamera.heading + headingDelta
             viewModel.cameraController.originCamera = viewModel.cameraController.originCamera.rotatedTo(
                 heading: newHeading,
@@ -152,7 +156,7 @@ extension WorldScaleGeoTrackingSceneView {
         
         /// Elevates the scene view camera controller by the elevation delta.
         /// - Parameter elevationDelta: The elevation delta.
-        private func updateElevation(_ elevationDelta: Double) {
+        private func updateElevation(by elevationDelta: Double) {
             viewModel.cameraController.originCamera = viewModel.cameraController.originCamera.elevated(by: elevationDelta)
             if let elevation = viewModel.cameraController.originCamera.location.z {
                 viewModel.calibrationElevation = elevation
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
@@ -108,9 +108,7 @@ extension WorldScaleGeoTrackingSceneView {
             }
             .onDisappear {
                 guard !viewModel.isCalibrating else { return }
-                withAnimation(.easeInOut) {
-                    setBasemapOpacity(0)
-                }
+                viewModel.setBasemapOpacity(0)
             }
             .overlay(alignment: .topTrailing) {
                 HStack {
@@ -127,13 +125,6 @@ extension WorldScaleGeoTrackingSceneView {
             .frame(maxWidth: .infinity)
             .padding()
             .background(.regularMaterial)
-        }
-        
-        /// Sets the basemap base layers with the given opacity.
-        /// - Parameter opacity: The opacity of the layer.
-        private func setBasemapOpacity(_ opacity: Float) {
-            guard let basemap = viewModel.scene.basemap else { return }
-            basemap.baseLayers.forEach { $0.opacity = opacity }
         }
         
         /// Updates the heading of the scene view camera controller.
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
@@ -25,6 +25,10 @@ extension WorldScaleGeoTrackingSceneView {
         @Binding var elevation: Double
         /// A Boolean value that indicates if the user is calibrating.
         @Binding var isCalibrating: Bool
+        /// The initial camera controller elevation.
+        @Binding var initialElevation: Double
+        /// The elevation delta value after calibrating.
+        @State private var elevationDelta = 0.0
         
         var body: some View {
             VStack {
@@ -58,6 +62,7 @@ extension WorldScaleGeoTrackingSceneView {
                                 .font(.body.smallCaps())
                                 .foregroundStyle(.secondary)
                             Spacer()
+                            Text(heading.isLess(than: 0) || heading.isZero ? "" : "+") +
                             Text(heading, format: .number.precision(.fractionLength(0)))
                             + Text("°")
                             Spacer()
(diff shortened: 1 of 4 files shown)


Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -0,0 +1,170 @@
...
+            .popover(isPresented: $viewModel.isCalibrating) {
+                VStack {
+                    Text("Heading: \(viewModel.calibrationHeading?.rounded(.towardZero) ?? viewModel.cameraController.originCamera.heading.rounded(.towardZero), format: .number)")
+                    
+                    HStack {
+                        Button {
+                            let heading = viewModel.cameraController.originCamera.heading - 1
+                            updateHeading(heading)
+                        } label: {
+                            Image(systemName: "minus")
```
yo1995, line 56: Maybe consider using `.imageScale(.large)` here. When I tried to tap the button, it is a bit small to me.

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -0,0 +1,170 @@
...
+                            if !editingChanged {
+                                headingTimer?.invalidate()
+                                headingTimer = nil
+                                headingSliderValue = 0.0
+                            }
+                        }
+                        .onChange(of: headingSliderValue) { heading in
+                            guard headingTimer == nil else { return }
+                            // Create a timer which rotates the camera when fired.
+                            let timer = Timer(timeInterval: 0.1, repeats: true) { [self] (_) in
```
yo1995, line 68: No need to strongly capture `self`, as the view is a struct, not a class.
```suggestion
                            let timer = Timer(timeInterval: 0.1, repeats: true) { _ in
```

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -0,0 +1,170 @@
...
+        var body: some View {
+            Button {
+                viewModel.isCalibrating = true
+                setBasemapOpacity(0.5)
+            } label: {
+                Text("Calibrate")
+            }
+            .popover(isPresented: $viewModel.isCalibrating) {
+                VStack {
+                    Text("Heading: \(viewModel.calibrationHeading?.rounded(.towardZero) ?? viewModel.cameraController.originCamera.heading.rounded(.towardZero), format: .number)")
```
yo1995, line 49: I think in SwiftUI we typically used [`numberFormatStyle`](https://developer.apple.com/documentation/foundation/measurement/formatstyle/3870188-numberformatstyle), though I'm not sure which is better - rounding vs formatting? 🤔 @philium would like to know your thoughts.

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -0,0 +1,170 @@
...
+                            updateHeading(heading)
+                        } label: {
+                            Image(systemName: "plus")
+                        }
+                    }
+                    VStack {
+                        Text("Elevation: \(viewModel.calibrationElevation?.rounded(.towardZero) ?? viewModel.cameraController.originCamera.location.z?.rounded(.towardZero) ?? 0, format: .number)")
+                        HStack {
+                            Button {
+                                updateElevation(-1)
```
yo1995, line 87: It may be worth retaining the argument label here. When I read this method it gives me an impression that it is updating the elevation value to `-1`

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -0,0 +1,170 @@
...
+                                updateElevation(1)
+                            } label: {
+                                Image(systemName: "plus")
+                            }
+                        }
+                    }
+                }
+                .onDisappear {
+                    guard !viewModel.isCalibrating else { return }
+                    withAnimation(.easeInOut) {
```
yo1995, line 118: I'm not seeing this animation - as the opacity is set within CoreArcGIS, I doubt it's possible to interpolate the opacity changes.

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -0,0 +1,170 @@
...
+                        setBasemapOpacity(0)
+                    }
+                }
+                .frame(idealWidth: UIScreen.main.bounds.width/3)
```
yo1995, line 122: Will be an linter error if linter is enabled.
```suggestion
                .frame(idealWidth: UIScreen.main.bounds.width / 3)
```

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -0,0 +1,170 @@
...
+                .fractionPresentationDetents(0.25)
+                .padding()
+            }
+            .esriBorder()
```
yo1995, line 126: Currently this button is at lower left with no margin. As Ryan said, a button that shows a material overlay would be a good alternative.
IMG_3499

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -288,3 +280,24 @@ public struct WorldScaleGeoTrackingSceneView: View {
         }
     }
 }
+
+extension WorldScaleGeoTrackingSceneView {
+    @MainActor
+    class ViewModel: ObservableObject {
+        /// The scene.
+        @Published var scene: ArcGIS.Scene
```
yo1995, line 288:
```suggestion
        let scene: ArcGIS.Scene
```

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -288,3 +280,24 @@ public struct WorldScaleGeoTrackingSceneView: View {
...
+        /// The camera controller that will be set on the scene view.
+        @Published var cameraController: TransformationMatrixCameraController
```
yo1995, line 290:
```suggestion
        let cameraController: TransformationMatrixCameraController
```

Sources/ArcGISToolkit/Extensions/SwiftUI/View.swift
```diff
@@ -101,6 +101,17 @@ extension View {
         modifier(SelectedModifier(isSelected: isSelected))
     }
     
+    /// Returns a new view with fraction presentation detents, if presentation
```
yo1995, line 104: See Ryan's comment on Slack.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -162,6 +159,11 @@ public struct WorldScaleGeoTrackingSceneView: View {
                 .padding(8)
                 .background(.regularMaterial, ignoresSafeAreaEdges: .horizontal)
         }
+        .onChange(of: viewModel.scene.basemap?.loadStatus) { loadStatus in
+            withAnimation {
+                viewModel.scene.basemap?.baseLayers.forEach( { $0.opacity = 0 })
```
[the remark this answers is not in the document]
des12437: This is used to set the basemap opacity to zero when the view first appears since the basemap base layers should only be visible while calibrating the elevation and heading. The baselayers are not available when the view initializes because the scene basemap has not loaded.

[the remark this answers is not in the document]
yo1995: I see 🫡

- Instead of setting the opacity on each status change, maybe only change it when status is `loaded`
- Animation seems not to be effective for setting the opacity of base layer, can be removed
- Since this method also sets the opacity like the other callsites in the calibration view, you can extract them into a method on the view model

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
des12437: I added argument labels `by` and `to` for clarity.

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -0,0 +1,170 @@
...
+                        } label: {
+                            Image(systemName: "plus")
+                                .imageScale(.large)
+                        }
+                    }
+                }
+            }
+            .onDisappear {
+                guard !viewModel.isCalibrating else { return }
+                viewModel.setBasemapOpacity(0)
```
yo1995, line 115: I'm seeing an error about dynamic member lookup on this line, but the app still compiles 🤔 Do you see an error here?

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -0,0 +1,170 @@
...
+                    Button {
+                        viewModel.isCalibrating = false
+                    } label: {
+                        Image(systemName: "xmark.circle")
+                            .imageScale(.large)
+                    }
+                    .frame(alignment: .topTrailing)
+                    .padding()
+                }
+                .padding([.trailing], -20)
```
yo1995, line 129: No need to use array with 1 option.
```suggestion
                .padding(.trailing, -20)
```

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -0,0 +1,170 @@
...
+                .padding([.top,], -30)
+            }
+            .frame(maxWidth: .infinity)
```
yo1995, line 132: Instead of having a infinite width view, maybe consider ideal width or a fixed frame, because we know the range of the sliders are fixed. Also maybe add some `cornerRadius` and `shadow` to make it look better.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -137,21 +140,27 @@ public struct WorldScaleGeoTrackingSceneView: View {
...
+                    updateSceneView(for: location)
                 }
             } catch {}
         }
-        .toolbar {
-            ToolbarItem(placement: .bottomBar) {
-                if !calibrationViewIsHidden {
-                    calibrationView
+        .overlay(alignment: calibrationViewAlignment) {
+            if !calibrationViewIsHidden {
```
yo1995, line 151: When the session isn't started, tapping on the "Calibrate" button doesn't show the view. Maybe add a disabled modifier when it is not tappable.

yo1995 (commented):
Good overall, still need some UI tweaks.

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
des12437: I do not see this error on my machine.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
des12437: Done.

rolson:
can you add a screen shot to the description please (iPhone and iPad)

rolson:
Thanks. Feedback:

- There should be padding around the view.
- I'm not sure the shadow needs to be there. For inspiration, start a podcast on your phone, then lock your phone. You will see the podcast controller on the bottom of the lockscreen. Look and feel should match as a good starting point for the calibration view.
- Should be a maxWidth specified on the frame so that it doesn't get too wide on iPad.

rolson:
In the "find my" app, tap on a person to get the half screen sheet of where they are and the circle x button looks different. It's dual color and has secondary styling. The code for that might look something like this:

```swift
        Button {
            // Dismiss
        } label: {
            Image(systemName: "xmark.circle.fill")
                .symbolRenderingMode(.hierarchical)
                .foregroundStyle(.secondary)
                .imageScale(.large)
        }
        .buttonStyle(.plain)
```

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -0,0 +1,178 @@
...
+                viewModel.setBasemapOpacity(0)
+            }
+            .overlay(alignment: .topTrailing) {
+                HStack {
+                    Spacer()
+                    Button {
+                        viewModel.isCalibrating = false
+                    } label: {
+                        Image(systemName: "xmark.circle.fill")
+                            .font(.system(size: 20))
```
des12437, line 124:
```suggestion
```
The font size is specified here to make the image larger, similar to the size of the x icon in the "find my" app. I can remove this line if the default size looks good.

rolson: We shouldn't use font sizes directly. We should always use dynamic fonts. Also the imageScale that I showed should change the size of the image.

des12437: Good point. The image size is too small using the  `.large` image scale. However, using the `.title3` font improves the icon size.

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
yo1995: I realized that the 2 `updateHeading(to:)` calls are replaceable by `rotateHeading(by:)`? For example, the add button `updateHeading(to: heading)` can be replaced by `rotateHeading(by: 1)`

des12437: Good catch.

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -0,0 +1,161 @@
...
+                }
+                .padding(.trailing, -20)
+                .padding(.top, -30)
+            }
+            .frame(maxWidth: 350)
+            .padding()
+            .padding(.top, 10)
+            .background(.regularMaterial)
+            .clipShape(RoundedRectangle(cornerRadius: 15))
+            .padding()
```
yo1995, line 136: Would it be possible to simplify the layout modifiers here? It would be better not to have multiple paddings unless they are strictly needed.

This might be a good place to use SwiftUI Preview to adjust.

Below is what I came up with. It may still need some adjustment though.

```swift
            // .padding(.top, 5)  // if you want to have some more room for the x button, after the VStacks structure is adjusted
            .overlay(alignment: .topTrailing) {
                Button {
                    viewModel.setBasemapOpacity(0)
                    withAnimation {
                        viewModel.isCalibrating = false
                    }
                } label: {
                    Image(systemName: "xmark.circle.fill")
                        .font(.title3)
                        .symbolRenderingMode(.hierarchical)
                        .foregroundStyle(.secondary)
                        .imageScale(.large)
                }
                .buttonStyle(.plain)
            }
            .frame(maxWidth: 350)
            .padding()
            .background(.regularMaterial)
            .clipShape(RoundedRectangle(cornerRadius: 15))
```

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -137,21 +140,29 @@ public struct WorldScaleGeoTrackingSceneView: View {
...
+            if configuration is ARWorldTrackingConfiguration,
+               !calibrationViewIsHidden {
+                if !viewModel.isCalibrating {
+                    Button {
+                        viewModel.isCalibrating = true
+                        viewModel.setBasemapOpacity(0.5)
+                    } label: {
+                        Text("Calibrate")
+                    }
+                    .frame(maxWidth: .infinity)
```
yo1995, line 160: I like the previous small button, just my thoughts. We can later adjust the AR view safe area to avoid the bottom blank area.

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -0,0 +1,161 @@
...
+        private var joystickHeadingDelta: Double {
+            Double(signOf: headingSliderValue, magnitudeOf: headingSliderValue * headingSliderValue / 25)
+        }
+        /// The elevation delta amount based on the elevation slider value.
+        private var joystickElevationDelta: Double {
+            Double(signOf: elevationSliderValue, magnitudeOf: elevationSliderValue * elevationSliderValue / 100)
+        }
+        
+        var body: some View {
+            VStack {
```
yo1995, line 41: Currently it is 2 nested `VStack` rather than 2 parallel VStacks. Do you want to make it like this instead?

```
VStack {
    VStack { … }
    VStack { … }
}

Comment (des12437) [reply Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift:41]: Done.

File: Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift (hunk 17 above)
Comment (des12437) [reply Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift]: I applied your suggestion and added some padding `.padding([.top, .trailing], -5)` for the x button.

File: Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift [hunk 20]
```diff
@@ -0,0 +1,161 @@
...
+                heading: newHeading,
+                pitch: originCamera.pitch,
+                roll: originCamera.roll
+            )
+            viewModel.calibrationHeading = newHeading
+        }
+        
+        /// Elevates the scene view camera controller by the elevation delta.
+        /// - Parameter elevationDelta: The elevation delta.
+        private func updateElevation(by elevationDelta: Double) {
```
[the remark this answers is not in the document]
des12437: I will leave as-is. The slider value increasing raises the elevation, and decreasing the slider value lowers the elevation which seems correct.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -137,21 +140,32 @@ public struct WorldScaleGeoTrackingSceneView: View {
...
+                    currentLocation = location
+                    updateSceneView(for: location)
                 }
             } catch {}
         }
-        .toolbar {
-            ToolbarItem(placement: .bottomBar) {
-                if !calibrationViewIsHidden {
-                    calibrationView
+        .overlay(alignment: calibrationViewAlignment) {
```
yo1995, line 150: Ideally, both the button and the calibration view are placed within the safe area. Currently the whole safe area situation is kind of messy, which can be fixed in a separate PR if you like.

- The AR view and coaching overlay should ignore safe area on all edges
- The status view is in safe area at top edge
- The calibration view and button are in safe area at each alignment

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
rolson: Actually, I guess if we are using the font size to define the size of the image then it should be hard-coded because we don't want that changing with the users font size settings.

Maybe a better way to do that though is to just set the frame width/height of the image.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -137,21 +140,32 @@ public struct WorldScaleGeoTrackingSceneView: View {
...
+            if configuration is ARWorldTrackingConfiguration,
```
yo1995, line 151: You can leave this condition here. I figured that it may be still worth having the calibration view for geo-tracking even if the adjustments are minor. I'll tweak the conditions in #574

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -162,6 +176,10 @@ public struct WorldScaleGeoTrackingSceneView: View {
                 .padding(8)
                 .background(.regularMaterial, ignoresSafeAreaEdges: .horizontal)
         }
+        .onChange(of: viewModel.scene.basemap?.loadStatus) { loadStatus in
+            guard loadStatus == .loaded else { return }
```
yo1995, line 180: Add some comment about what is the expected visual clue when the basemap first loads.

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -0,0 +1,158 @@
...
+    struct CalibrationView: View {
+        @ObservedObject var viewModel: ViewModel
+        /// The slider value for the camera heading.
+        @State private var headingSliderValue = 0.0
+        /// The slider value for the camera elevation.
+        @State private var elevationSliderValue = 0.0
+        /// The elevation timer for the "joystick" behavior.
+        @State private var headingTimer: Timer?
+        /// The heading timer for the "joystick" behavior.
+        @State private var elevationTimer: Timer?
```
rolson, line 30: instead of embedding all the joystick logic in this view, it could be separated out into a JoystickSlider view. Then that would clean up this code immensely. Also reduce duplication potentially.

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -0,0 +1,158 @@
...
+                        .imageScale(.large)
+                }
+                .buttonStyle(.plain)
+                .padding([.top, .trailing], -5)
+            }
+            .frame(maxWidth: 350)
+            .padding()
+            .background(.regularMaterial)
+            .clipShape(RoundedRectangle(cornerRadius: 15))
+            .padding()
```
yo1995, line 133: I noticed this extra padding. Is it used for avoiding going out of the safe area?

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
rolson: you can round in the format either inline or with an extension:

```swift
extension FloatingPointFormatStyle {
    public var noFractionalDigits: FloatingPointFormatStyle<Value> {
        precision(.fractionLength(0...0))
    }
}
```

usage:

```swift
Text(double, format: .number.noFractionalDigits)
```

or simply inline

```swift
Text(double, format: .number.precision(.fractionLength(0...0)))
```

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -137,21 +140,32 @@ public struct WorldScaleGeoTrackingSceneView: View {
...
+                        withAnimation {
+                            viewModel.isCalibrating = true
+                        }
+                    } label: {
+                        Text("Calibrate")
+                    }
+                    .padding()
+                    .background(.regularMaterial)
+                    .clipShape(RoundedRectangle(cornerRadius: 10))
+                    .padding()
```
yo1995, line 165: Same question here. If it is, we can work on the layouts in a separate issue.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -186,17 +204,32 @@ public struct WorldScaleGeoTrackingSceneView: View {
...
+            guard !viewModel.isCalibrating else { return }
+            viewModel.cameraController.originCamera = Camera(
+                latitude: location.position.y,
+                longitude: location.position.x,
+                altitude: viewModel.calibrationElevation ?? altitude,
+                heading: viewModel.calibrationHeading ?? 0,
+                pitch: 90,
+                roll: 0
+            )
+        }
```
yo1995, line 229: You can keep the old code here to first try using calibration values, and fall back to default values. I'll change these in #574

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
rolson, line 150: with alignment set to `.bottomLeading` this is how it looks on my phone. Needs leading padding:
IMG_6718
IMG_6717

rolson:
It would be great if we could get a preview to work because it would make it much easier to iterate on the design. But this doesn't work in Xcode for me. Fails to build.

```swift
#Preview {
    WorldScaleGeoTrackingSceneView.CalibrationView(
        viewModel: .init(scene: Scene(), cameraController: TransformationMatrixCameraController())
    )
}
```

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -0,0 +1,158 @@
...
+                        } label: {
+                            Image(systemName: "plus")
+                                .imageScale(.large)
+                        }
+                    }
+                }
+            }
+            .overlay(alignment: .topTrailing) {
+                Button {
+                    viewModel.setBasemapOpacity(0)
```
rolson, line 115: This might be over-reaching a bit. I'm not sure that we'd want to do this. We might want to be able to tell the user that calibration is happening and they can respond. But I don't know that we'd build this behavior in.

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
des12437: Done. The image size is now set with the frame width and height.

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
des12437: Done.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
des12437: Fixed with padding around the calibration button and view.

des12437: The safe area issues can be addressed in a separate PR.

des12437:
This PR is ready for re-review.

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
philium: The fraction length doesn't need to be a range. If you always want it to be a certain length, just use that length:

```swift
format: .number.precision(.fractionLength(0))
```

rolson (commented):
Running this on a device and I still see the calibration button and view partially off the screen on the left, like the screen shots I've already pasted in here. Do you see this same issue running on device? I'm running on iPhone 12 Pro.

des12437:
> Running this on a device and I still see the calibration button and view partially off the screen on the left, like the screen shots I've already pasted in here. Do you see this same issue running on device? I'm running on iPhone 12 Pro.

I do not see this on an iPad device or on an iPhone 12 Pro simulator. Since the calibrate button has padding the view should not be off screen.

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -0,0 +1,193 @@
...
+        var headingSlider: some View {
+            VStack {
+                HStack {
+                    Stepper() {
+                        HStack {
+                            Text("Heading")
+                                .font(.body.smallCaps())
+                                .foregroundStyle(.secondary)
+                            Spacer()
+                            Text((viewModel.calibrationHeading ?? viewModel.cameraController.originCamera.heading), format: .number.precision(.fractionLength(0)))
```
rolson, line 65: Add in degree symbol

```suggestion
                            Text((viewModel.calibrationHeading ?? viewModel.cameraController.originCamera.heading), format: .number.precision(.fractionLength(0)))
                            + Text("°")
```

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -0,0 +1,193 @@
...
+        var elevationSlider: some View {
+            VStack {
+                HStack {
+                    Stepper() {
+                        HStack {
+                            Text("Elevation (m)")
+                                .font(.body.smallCaps())
+                                .foregroundStyle(.secondary)
+                            Spacer()
+                            Text((viewModel.calibrationElevation ?? viewModel.cameraController.originCamera.location.z ?? 0), format: .number.precision(.fractionLength(0)))
```
rolson, line 91: The meters designation should go on the value, not the label.

```suggestion
                            Text("Elevation")
                                .font(.body.smallCaps())
                                .foregroundStyle(.secondary)
                            Spacer()
                            Text((viewModel.calibrationElevation ?? viewModel.cameraController.originCamera.location.z ?? 0), format: .number.precision(.fractionLength(0)))
                            + Text("m")
```

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -0,0 +1,193 @@
...
+        Slider(value: $value, in: -10...10) { editingChanged in
+            if !editingChanged {
+                timer?.invalidate()
+                timer = nil
+                value = 0.0
+            }
+        }
+        .onChange(of: value) { value in
+            guard timer == nil else { return }
+            // Create a timer which rotates the camera when fired.
```
rolson, line 172: comment no longer applies, should be corrected

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -0,0 +1,193 @@
...
+    /// The slider value.
+    @State private var value = 0.0
+    /// The timer for the "joystick" behavior.
+    @State private var timer: Timer?
+    /// The delta amount based on the slider value.
+    private var joystickDelta: Double {
+        Double(signOf: value, magnitudeOf: value * value / 25)
+    }
+    /// User defined action to be performed when the slider value changes.
+    var sliderValueChangedAction: ((Double) -> Void)? = nil
```
rolson, line 160: is this the slider value changing, or the slider delta changing? It seems to be the delta in this case correct? If so then name and comment should reflect that.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -288,3 +291,21 @@ public struct WorldScaleGeoTrackingSceneView: View {
         }
     }
 }
+
+extension WorldScaleGeoTrackingSceneView {
+    @MainActor
+    class ViewModel: ObservableObject {
```
rolson, line 297: It's unclear if this view model is for the calibration view or the WorldScaleGeoTrackingSceneView. 

I think that's because it's used by both. To me it seems like the CalibrationView doesn't need this view model at all, it simply needs bindings to heading and elevation. From there the WorldScaleGeoTrackingSceneView can pass bindings to state to it, and the calibration view can update the state. Then the WorldScaleGeoTrackingSceneView can implement onChange for those values and update the camera controller. So I'm not sure this view model is needed.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -137,21 +138,33 @@ public struct WorldScaleGeoTrackingSceneView: View {
...
+                    currentLocation = location
+                    updateSceneView(for: location)
                 }
             } catch {}
         }
-        .toolbar {
-            ToolbarItem(placement: .bottomBar) {
-                if !calibrationViewIsHidden {
-                    calibrationView
+        .overlay(alignment: calibrationViewAlignment) {
```
rolson, line 148: I think the calibration view should always be in the same spot, regardless of where the button is.

```suggestion
        .overlay(alignment: calibrationViewAlignment) {
            if configuration is ARWorldTrackingConfiguration,
               !calibrationViewIsHidden {
                if !viewModel.isCalibrating {
                    VStack {
                        Button {
                            withAnimation {
                                viewModel.isCalibrating = true
                            }
                        } label: {
                            Text("Calibrate calibrate")
                        }
                        .padding()
                        .background(.regularMaterial)
                        .clipShape(RoundedRectangle(cornerRadius: 10))
                        .disabled(!initialCameraIsSet)
                    }
                    .padding()
                }
            }
        }
        .overlay(alignment: .bottom) {
            if viewModel.isCalibrating {
                CalibrationView(viewModel: viewModel)
            }
        }
```

rolson: Actually the VStack for the button is unnecessary:

```swift
                        Button {
                            withAnimation {
                                viewModel.isCalibrating = true
                            }
                        } label: {
                            Text("Calibrate calibrate")
                        }
                        .padding()
                        .background(.regularMaterial)
                        .clipShape(RoundedRectangle(cornerRadius: 10))
                        .disabled(!initialCameraIsSet)
                        .padding()
```

rolson:
> I do not see this on an iPad device or on an iPhone 12 Pro simulator. Since the calibrate button has padding the view should not be off screen.

The coaching view is causing this for some reason. If I comment out the coaching view, it displays in the right location.

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
des12437: Fixed.

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
des12437: Yes the slider delta value is changing, the names have been updated.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
des12437: Removed the view model as it was not needed.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
des12437: Done.

des12437:
> > I do not see this on an iPad device or on an iPhone 12 Pro simulator. Since the calibrate button has padding the view should not be off screen.
> 
> The coaching view is causing this for some reason. If I comment out the coaching view, it displays in the right location.

I'm trying to reproduce this issue.

rolson:
> > > I do not see this on an iPad device or on an iPhone 12 Pro simulator. Since the calibrate button has padding the view should not be off screen.
> > 
> > 
> > The coaching view is causing this for some reason. If I comment out the coaching view, it displays in the right location.
> 
> I'm trying to reproduce this issue.

I have a fix [here](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/575/files). It also fixes the safe area insets

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -22,6 +22,14 @@ public struct WorldScaleGeoTrackingSceneView: View {
     @State private var arViewProxy = ARSwiftUIViewProxy()
     /// The camera controller that will be set on the scene view.
     @State private var cameraController: TransformationMatrixCameraController
+    /// The camera controller heading.
+    @State var heading: Double = 0
+    /// The camera controller elevation.
+    @State var elevation: Double = 0
+    /// The calibrated elevation delta.
+    @State private var elevationDelta: Double = 0
```
rolson, line 30: I wouldn't think you would need both the elevation and the elevation delta. I think you should be able to calculate the delta from the current elevation: cameraController.originCamera.location.z.

des12437: The elevation delta can be calculated, I will remove `elevationDelta`.

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -19,140 +19,150 @@ import ArcGIS
...
-                                .imageScale(.large)
+                .padding(.bottom)
+                headingSlider
+                Divider()
+                elevationSlider
+            }
+            .padding()
+            .background(.regularMaterial)
+            .clipShape(RoundedRectangle(cornerRadius: 15))
+            .frame(maxWidth: 350)
```
yo1995, line 47: Shall we pick a slightly bigger frame size, so that the view always take the full width minus the padding on an iPhone portrait orientation, and be a fixed size on iPad? From the [resolution list](https://developer.apple.com/design/human-interface-guidelines/layout#Specifications) it seems 414 (iPhone 15 PM's 430 - 2 * 8) can be a good value.

In this way, the leading padding for the button and calibration view will be consistent when the button is placed on leading or trailing edge.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -287,40 +305,20 @@ public struct WorldScaleGeoTrackingSceneView: View {
...
-        func setBasemapOpacity(_ opacity: Float) {
-            guard let basemap = scene.basemap else { return }
-            basemap.baseLayers.forEach { $0.opacity = opacity }
+            .multilineTextAlignment(.center)
+            .padding(8)
+            .frame(maxWidth: .infinity, alignment: .center)
+            .background(.regularMaterial)
+            .clipShape(RoundedRectangle(cornerRadius: 10))
+            .padding(.horizontal)
+            .padding(.top)
```
yo1995, line 321:
```suggestion
            .padding([.horizontal, .top])
```

yo1995 (commented):
The new changes look better! 👍

yo1995 (dismissed):

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -0,0 +1,168 @@
...
+                    .foregroundStyle(.secondary)
+                    .frame(width: 28, height: 28)
+            }
+            .buttonStyle(.plain)
+        }
+    }
+}
+
+/// A view for a joystick style slider.
+struct JoystickSliderView: View {
```
rolson, line 124:
```suggestion
private struct JoystickSliderView: View {
```

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -0,0 +1,168 @@
...
+            }
+            self.timer = timer
+            // Add the timer to the main run loop.
+            RunLoop.main.add(timer, forMode: .default)
+        }
+    }
+    
+    /// Sets an action to perform when the slider delta value changes.
+    /// - Parameter action: The action to perform when the slider delta value has changed.
+    public func onSliderDeltaValueChanged(
```
rolson, line 161:
```suggestion
    func onSliderDeltaValueChanged(
```

rolson (commented):
Correct me if I'm wrong, but it seems like in the UI we show a delta value for the heading, but we show the actual value for the elevation. Since heading and elevation are changing as new locations are coming in and as the device moves, I think we should just show delta values for both.

And if we show delta values, then I think we need to have a + prefix for positive values so it is more obvious that the value is a delta. Would that make sense?

des12437:
> Correct me if I'm wrong, but it seems like in the UI we show a delta value for the heading, but we show the actual value for the elevation. Since heading and elevation are changing as new locations are coming in and as the device moves, I think we should just show delta values for both.

This is correct, so we should show the updated heading and elevation values or the deltas for both.
 
> And if we show delta values, then I think we need to have a + prefix for positive values so it is more obvious that the value is a delta. Would that make sense?

If we do show the deltas then I agree a + prefix is helpful. Relative heading and elevation changes are probably more useful than absolute values. I'm not sure there is a case where the absolute values are necessary.

rolson:
Yes, I think delta values make more sense.

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -58,6 +62,7 @@ extension WorldScaleGeoTrackingSceneView {
                                 .font(.body.smallCaps())
                                 .foregroundStyle(.secondary)
                             Spacer()
+                            Text(heading.isLess(than: 0) || heading.isZero ? "" : "+") +
```
rolson, line 65: isn't this the heading delta?

rolson, line 65: you can send up with a "+0°" in the text because it's not comparing the rounded number we show in the text to zero, but the actual value.

rolson: Also, the heading control allows you to go over 360 and less than -360. It should limit your calibration to -180...180.

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -0,0 +1,180 @@
...
+        /// The camera controller heading.
+        @Binding var heading: Double
+        /// The camera controller elevation.
+        @Binding var elevation: Double
+        /// A Boolean value that indicates if the user is calibrating.
+        @Binding var isCalibrating: Bool
+        /// The initial camera controller elevation.
+        @Binding var initialElevation: Double
+        /// The elevation delta value after calibrating.
+        @State private var elevationDelta = 0.0
```
rolson, line 31: Are all of these values needed? It seems that the calibration view only needs to give a delta, I don't think it would actually need to know about the actual heading or elevation. Only the delta values.

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
des12437: > isn't this the heading delta?

Yes. The heading is relative to zero so a separate value is not needed like for the elevation delta which is relative to a starting elevation offset. The property could be renamed.

rolson:
The behavior now with the clamping of -180...180 is not correct. If you get to -180 and the tap the minus button on the stepper, the value goes to -1, but -1 is not -181. Ideally, the value just stops at -180 and 180. Tapping the stepper and using the slider shouldn't do anything at that point.

rolson:
> The behavior now with the clamping of -180...180 is not correct. If you get to -180 and the tap the minus button on the stepper, the value goes to -1, but -1 is not -181 (for example, -1 is 359). Ideally, the value just stops at -180 and 180. Tapping the stepper and using the slider shouldn't do anything at that point.

@des12437 this is fixed in the PR here, you can merge it into your branch please.

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
des12437: These values are necessary for the stepper control logic since the elevation delta value does not change.

rolson (approved):

yo1995 (approved):

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
philium: Consider using a measurement format style, which is best practice for formatting measurements:

```swift
+ Text(
    Measurement<UnitAngle>(value: heading, unit: .degrees),
    format: .measurement(width: .narrow, numberFormatStyle: .number.precision(.fractionLength(0)))
)
```

If you want to, you could even store the heading as a measurement.

Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
philium: Consider using a measurement format style, which is best practice for formatting measurements:

```swift
Text(
    Measurement<UnitLength>(value: elevationDelta, unit: .meters),
    format: .measurement(width: .narrow, numberFormatStyle: .number.precision(.fractionLength(0)))
)
```

If you want to, you could even store the elevation delta as a measurement.

PR #572, opened 2024-01-24
AR: Adjust coaching overlay display conditions

yo1995:
## Description

This PR adjusts the coaching overlay show/dismiss conditions for the world scale AR view. See comments in the code.

The table top AR view also uses the coaching overlay. However, it is manually controlled by inactivating the overlay when the plane anchor is added. Thus, changes in this PR won't affect the behavior of the overlay in the table top AR view.

https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/67228cfadc9b6f0af0c8e52ee3a597f91b04118e/Sources/ArcGISToolkit/Components/Augmented%20Reality/TableTopSceneView.swift#L171-L173

See more details in `swift/issues/4909`

### DIFF
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/ARCoachingOverlay.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/ARCoachingOverlay.swift
@@ -16,7 +16,14 @@ import ARKit
 import SwiftUI
 
 /// A SwiftUI version of an ARCoachingOverlayView view.
-struct ARCoachingOverlay: UIViewRepresentable {
+struct ARCoachingOverlay {
+    /// The closure to call when the coaching overlay view activates.
+    private(set) var onCoachingOverlayActivateAction: ((ARCoachingOverlayView) -> Void)?
+    /// The closure to call when the coaching overlay view deactivates.
+    private(set) var onCoachingOverlayDeactivateAction: ((ARCoachingOverlayView) -> Void)?
+    /// The closure to call when the user taps the coaching overlay view's Start Over button.
+    private(set) var onCoachingOverlayRequestSessionResetAction: ((ARCoachingOverlayView) -> Void)?
+    
     /// The data source for an AR session.
     var sessionProvider: ARSessionProviding?
     /// The goal for the coaching overlay.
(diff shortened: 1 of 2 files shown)


Sources/ArcGISToolkit/Components/Augmented Reality/ARCoachingOverlay.swift
```diff
@@ -45,7 +45,7 @@ struct ARCoachingOverlay: UIViewRepresentable {
     func makeUIView(context: Context) -> ARCoachingOverlayView {
         let view = ARCoachingOverlayView()
         view.delegate = context.coordinator
-        view.activatesAutomatically = false
+        view.activatesAutomatically = true
```
yo1995, line 48: The default value for `activatesAutomatically` is `true`. However, I feel it would be beneficial to explicitly set it to `true` here so it is easier to reason its behavior.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -48,6 +46,8 @@ public struct WorldScaleGeoTrackingSceneView: View {
     @State private var currentLocation: Location?
     /// The valid accuracy threshold for a location in meters.
     private var validAccuracyThreshold = 0.0
+    /// The app's tracking requirements for the coaching overlay to show/dismiss.
+    @State private var goal: ARCoachingOverlayView.Goal
```
yo1995, line 50: Instead of using a boolean, we use the goal to let the coaching overlay decide for itself when to show or dismiss. For instance, if the geo-tracking accuracy drops to low or undetermined, the overlay may show again to prompt the user to re-localize the AR view.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -73,10 +73,12 @@ public struct WorldScaleGeoTrackingSceneView: View {
         
         if ARGeoTrackingConfiguration.isSupported {
             configuration = ARGeoTrackingConfiguration()
+            _goal = .init(initialValue: .geoTracking)
         } else {
             configuration = ARWorldTrackingConfiguration()
-            configuration.worldAlignment = .gravityAndHeading
+            _goal = .init(initialValue: .tracking)
         }
+        configuration.worldAlignment = .gravityAndHeading
```
yo1995, line 81: According to my test, the default `worldAlignment` for `ARGeoTrackingConfiguration` is `gravityAndHeading`, and for `ARWorldTrackingConfiguration` is `gravity`.

Nonetheless, I  moved this line here so it is clearer for both configurations, we set it to `gravityAndHeading`

des12437 (commented):
Looks good. There should be a conditional check added to `updateSceneView()` to not set the initial origin camera until the coaching overlay is dismissed. Otherwise the accuracy will be low because the `ARCoachingOverlayView.Goal` is not met yet.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
yo1995: The comment above is incorrect. For both world tracking and geo tracking configurations, we can use the same `.geoTracking` goal, and the overlay is smart enough to display different prompt.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
yo1995: After some more thoughts, I'll revert this change

des12437 (approved):

PR #573, opened 2024-01-25
handle negotiate challenge

njarecha:


dfeinzimer (approved):
Thanks @njarecha!

Sources/ArcGISToolkit/Components/Authentication/NetworkChallengeContinuation.swift
```diff
@@ -72,6 +72,8 @@ extension NetworkChallengeContinuation.Kind {
             self = .login
         case .clientCertificate:
             self = .certificate
+        case .negotiate:
+            fatalError("Negotiate challenge continuation is not implemented.")
```
rolson, line 76: To me this sounds like it needs to happen at some point. But if that's not the case I think we might want a more clear message. Is this accurate?
```suggestion
            fatalError("We should not end up here as negotiate challenges are rejected when they come in.")
```

rolson: The other option is to make the init failable.

njarecha: Updated message looks good.

rolson (approved):

dfeinzimer (approved):

njarecha:
Thanks @dfeinzimer and @rolson!

PR #577, opened 2024-02-06
Add custom joystick style slider

rolson:
The current slider doesn't work for me. There are 2 problems I see with it:

- it looks exactly like the system slider (because it is), but the behavior is different. If the behavior is different, it should probably look different.
-  left side of the slider is blue, which seems to indicate that the value is half, but the value is never half. The concept of "half" doesn't exist for what we are doing.

This PR introduces a custom joystick slider that looks very close to the system slider, but has slightly different colors applied.

Also this PR fixes a bug where you can see "-0" in the text in certain situations.

This is what the custom slider looks like as you are dragging it:

IMG_6841

This is what it looks like at rest:

IMG_6840

### DIFF
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
@@ -31,7 +31,7 @@ extension WorldScaleGeoTrackingSceneView {
         @State private var elevationDelta = 0.0
         /// A number format style for signed values with their fractional component removed.
         private let numberFormat = FloatingPointFormatStyle<Double>.number
-            .precision(.fractionLength(0))
+            .precision(.fractionLength(1))
             .sign(strategy: .always(includingZero: false))
         
         var body: some View {
@@ -79,10 +79,6 @@ extension WorldScaleGeoTrackingSceneView {
                     .onChanged { delta in
                         heading = (heading + delta).clamped(to: -180...180)
                     }
-                    .onEnded {
-                        // Round the value now that it stopped changing.
-                        heading = heading.rounded()
-                    }
             }
         }
         
@@ -108,10 +104,6 @@ extension WorldScaleGeoTrackingSceneView {
                 Joyslider()
                     .onChanged { delta in
                         elevation += delta
-                    }
-                    .onEnded {
-                        // Round the value now that it stopped changing.
-                        elevation = elevation.rounded()
                     }
             }
             .onChange(of: elevation) { elevation in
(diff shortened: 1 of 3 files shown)


Sources/ArcGISToolkit/Components/Augmented Reality/CalibrationView.swift
```diff
@@ -30,6 +30,10 @@ extension WorldScaleGeoTrackingSceneView {
         /// The elevation delta value after calibrating.
         @State private var elevationDelta = 0.0
```
des12437, line 32:
```suggestion
        /// A number format style for signed values with their fractional component removed.
```

des12437 (dismissed):
Slider color and behavior looks good.

des12437 (dismissed):

rolson:
Update: I also updated this to show tenths because you actually notice these quite well and the rounding was causing a jump at the end of the joystick edit.

rolson:
@des12437 @yo1995 this is ready for review. I will merge this PR and then change [this other PR](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/579) to target v.next and merge that.

des12437 (dismissed):

Sources/ArcGISToolkit/Components/Augmented Reality/Joyslider.swift
```diff
@@ -0,0 +1,159 @@
...
+                    .frame(width: Self.size, height: Self.size)
+                    .foregroundStyle(.background)
+                    .shadow(color: .secondary.opacity(0.5), radius: 5, x: 0, y: 2)
+            }
+    }
+}
+
+/// Track view for the joyslider.
+private struct Track: View {
+    /// A Boolean value indicating if the user is dragging the thumb.
```
yo1995, line 111:
```suggestion
    /// The x offset of the thumb in points.
```

yo1995 (commented):
LGTM, some minor changes

Sources/ArcGISToolkit/Components/Augmented Reality/Joyslider.swift
```diff
@@ -0,0 +1,159 @@
...
+                    .fill(fill)
+                    .frame(width: abs(offset))
+                    .offset(x: offset / 2)
+            }
+        }
+        .frame(height: 4)
+    }
+    
+    var fill: some ShapeStyle {
+        if offset >= 0 {
```
yo1995, line 131: This can also be done with a modifier such as `.scaleEffect(x: offset >= 0 ? 1: -1, y: 1)` that flips the capsule horizontally. I'm fine with the current implementation.

yo1995 (approved):

PR #581, opened 2024-02-08
AR: Separate world vs geo tracking logic

yo1995:
## Description

This PR separates the world scale AR into 2 separate child view workflows: `WorldTrackingSceneView` that uses world-tracking, and `GeoTrackingSceneView` that uses geo-tracking. They are subsequently managed by a wrapper view `WorldScaleSceneView` to handle the fall back logics.

All the non developer facing views are moved into a subfolder `Augmented Reality/Utilities` to make it clearer which views are for public use.

Please see the comment below on some deficiencies.


Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/WorldTrackingSceneView.swift
```diff
@@ -16,74 +16,94 @@ import ARKit
...
+    @State private var interfaceOrientation: InterfaceOrientation?
+    /// The timestamp of the last received location.
+    @State private var lastLocationTimestamp: Date?
+    
+    /// Projected point from the spatial reference of the location data source's to the scene view's.
+    private var currentPosition: Point? {
+        guard let currentLocation,
+              let currentCamera,
+              let spatialReference = currentCamera.location.spatialReference,
+              let position = GeometryEngine.project(currentLocation.position, into: spatialReference)
```
yo1995, line 58: I'm not sure if we should also use the elevation (z value) from the projected point, but it doesn't make a big difference in my tests.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/WorldTrackingSceneView.swift
```diff
@@ -192,26 +180,23 @@ public struct WorldScaleGeoTrackingSceneView: View {
...
         
-        // Do not use cached location more than 10 seconds old.
-        guard abs(lastLocationTimestamp?.timeIntervalSinceNow ?? 0) < 10 else { return }
+        // Update if the location is more than 10 seconds old.
+        guard abs(lastLocationTimestamp?.timeIntervalSinceNow ?? 0) < timeThreshold else { return }
         
         // Make sure that horizontal and vertical accuracy are valid.
-        guard location.horizontalAccuracy > validAccuracyThreshold,
-              location.verticalAccuracy > validAccuracyThreshold else { return }
+        guard location.horizontalAccuracy > .zero,
```
yo1995, line 194: I changed them to a fixed `zero` because the doc says negative value means invalid.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
```diff
@@ -0,0 +1,262 @@
...
+            }
+        }
+        .overlay(alignment: .bottom) {
+            if isCalibrating {
+                CalibrationView(viewModel: calibrationViewModel, isPresented: $isCalibrating)
+                    .padding(.bottom)
+            }
+        }
+        .overlay(alignment: .top) {
+            accuracyView
```
yo1995, line 169: Depending on what we want to display in the accuracy view, a few "onSomethingChanged" modifiers may be needed in the child views. They can be done in a separate PR.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
```diff
@@ -0,0 +1,262 @@
...
+    public func calibrationViewAlignment(_ alignment: Alignment) -> Self {
+        var view = self
+        view.calibrationViewAlignment = alignment
+        return view
+    }
+    
+    /// Checks if GPS is providing the most accurate location and heading.
+    /// - Parameter locationManager: The location manager to determine the accuracy authorization.
+    /// - Returns: A Boolean value that indicates whether tracking is accurate.
+    private func checkTrackingCapabilities(_ locationManager: CLLocationManager) -> Bool {
```
yo1995, line 210: We may not need this method. I added it for discussion - we probably want a way to ensure the hardware is providing accurate updates before the AR experience starts.

Examples/Examples/WorldScaleExampleView.swift
```diff
@@ -20,7 +20,7 @@ import CoreLocation
 /// An example that utilizes the `WorldScaleGeoTrackingSceneView` to show an augmented reality view
```
des12437, line 20:
```suggestion
/// An example that utilizes the `WorldScaleSceneView` to show an augmented reality view
```

Examples/Examples/WorldScaleExampleView.swift
```diff
@@ -52,7 +52,7 @@ struct WorldScaleGeoTrackingExampleView: View {
                 }
         }
         .calibrationViewHidden(false)
```
des12437, line 54:
```suggestion
```
This is set to `false` by default.

Examples/Examples/WorldScaleExampleView.swift
```diff
@@ -52,7 +52,7 @@ struct WorldScaleGeoTrackingExampleView: View {
...
-        .calibrationViewAlignment(.bottomLeading)
+        .calibrationViewAlignment(.bottom)
```
des12437, line 55:
```suggestion
```
The alignment is set to `.bottom` by default.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/GeoTrackingSceneView.swift
```diff
@@ -0,0 +1,204 @@
+// Copyright 2023 Esri
```
des12437, line 1:
```suggestion
// Copyright 2024 Esri
```

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/GeoTrackingSceneView.swift
```diff
@@ -0,0 +1,204 @@
...
+/// A scene view that provides an augmented reality world scale experience using geo-tracking.
+public struct GeoTrackingSceneView: View {
+    /// A Boolean value indicating if the camera was initially set.
+    @Binding var initialCameraIsSet: Bool
+    /// The view model for the calibration view.
+    @ObservedObject private var calibrationViewModel: WorldScaleSceneView.CalibrationViewModel
+    /// The geo-tracking configuration for the AR session.
+    private let configuration = ARGeoTrackingConfiguration()
+    /// A Boolean value that indicates if the user is calibrating.
+    private let isCalibrating: Bool
```
des12437, line 28: Should this property be renamed to `isCalibrationViewEnabled` for example since it will not change?

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
```diff
@@ -0,0 +1,274 @@
+// Copyright 2023 Esri
```
des12437, line 1:
```suggestion
// Copyright 2024 Esri
```

Examples/Examples/WorldScaleExampleView.swift
yo1995: Maybe we should include it in the example so someone will be able to find the modifier easier? I'll revert my change. 🤔

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/GeoTrackingSceneView.swift
des12437: `calibrationViewIsEnabled` reads like an assertion.`calibrationIsEnabled` may be a better name suggestion, with documentation to specify that this property controls the calibration view visibility.

Examples/Examples/WorldScaleExampleView.swift
```diff
@@ -51,7 +51,6 @@ struct WorldScaleGeoTrackingExampleView: View {
                     }
                 }
         }
-        .calibrationViewHidden(false)
         .calibrationViewAlignment(.bottomLeading)
         .task {
```
des12437, line 55: The locationDataSource should be started in this task block otherwise the circle graphic around the initial location does not show.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/GeoTrackingSceneView.swift
```diff
@@ -0,0 +1,204 @@
...
+    /// The closure that builds the scene view.
+    private let sceneViewBuilder: (SceneViewProxy) -> SceneView
+    /// The proxy for the ARSwiftUIView.
+    @State private var arViewProxy = ARSwiftUIViewProxy()
+    /// The camera controller that will be set on the scene view.
+    @State private var cameraController: TransformationMatrixCameraController
+    /// A Boolean value that indicates whether the coaching overlay view is active.
+    @State private var coachingOverlayIsActive = false
+    /// The current camera of the scene view.
+    @State private var currentCamera: Camera?
```
des12437, line 40:
```suggestion
```
This property is updated to upon camera changes but is never read from. Can this be removed?

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/GeoTrackingSceneView.swift
rolson: SwiftUI tends to use the terms "is presented". With that in mind, `calibrationViewIsPresented` or `calibrationIsPresented` works.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/GeoTrackingSceneView.swift
yo1995: Yes. I intended to use it to get the scene view's spatial reference so to project the vertical coordinate, but now I think it may be redundant.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/GeoTrackingSceneView.swift
```diff
@@ -0,0 +1,199 @@
...
+import ARKit
+import SwiftUI
+import ArcGIS
+
+/// A scene view that provides an augmented reality world scale experience using geo-tracking.
+public struct GeoTrackingSceneView: View {
+    /// A Boolean value indicating if the camera was initially set.
+    @Binding var initialCameraIsSet: Bool
+    /// The view model for the calibration view.
+    @ObservedObject private var calibrationViewModel: WorldScaleSceneView.CalibrationViewModel
```
rolson, line 24: Now that the nested type is utilized outside the context of it's parent, we should un-nest the type and just call it WorldScaleCalibrationViewModel.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/GeoTrackingSceneView.swift
```diff
@@ -0,0 +1,199 @@
...
+                            orientation: interfaceOrientation,
+                            initialTransformation: .identity
+                        )
+                        sceneViewProxy.setFieldOfView(
+                            for: frame,
+                            orientation: interfaceOrientation
+                        )
+                    }
+                    .onDidChangeGeoTrackingStatus { session, status in
+                        switch status.state {
```
rolson, line 96: I wonder if everything in this closure should go in a @MainActor method called something like "handleGeoTrackingStatusChange". Because I see that `initialCameraIsSet` must be set on the mainthread, but I am also seeing that it's being read on line 104, and that probably should be on the same thread as well.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/GeoTrackingSceneView.swift
```diff
@@ -0,0 +1,199 @@
...
+                        }
+                    }
+                
+                if initialCameraIsSet {
+                    sceneViewBuilder(sceneViewProxy)
+                        .cameraController(cameraController)
+                        .attributionBarHidden(true)
+                        .spaceEffect(.transparent)
+                        .atmosphereEffect(.off)
+                        .interactiveNavigationDisabled(true)
```
rolson, line 127: I assume this is duplicated down below in the world tracking version. We should try to take as much as we can and consolidate to avoid code duplication. For example, these 5 modifiers can be rolled into a single modifier.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/GeoTrackingSceneView.swift
```diff
@@ -0,0 +1,199 @@
...
+            for await location in locationDataSource.locations {
+                currentLocation = location
+            }
+        }
+        .task {
+            for await heading in locationDataSource.headings {
+                currentHeading = heading
+            }
+        }
+        .onReceive(calibrationViewModel.headingCorrections) { correction in
```
rolson, line 164: If this logic is also duplicated in the WorldTracking version, then we should see if there is a way to break the logic out into a view modifier that can be reused. It might need to take a calibrationViewModel and a camera controller.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/GeoTrackingSceneView.swift
```diff
@@ -0,0 +1,199 @@
...
+    
+    /// Updates the scene view's camera controller with location and heading.
+    /// - Parameters:
+    ///   - location: The location for the camera.
+    ///   - heading: The heading for the camera.
+    ///   - altitude: The altitude for the camera.
+    private func updateCameraController(location: Location, heading: Double, altitude: Double) {
+        // Add some of the vertical accuracy to the z value of the position, that way if the
+        // GPS location is not accurate, it won't end up below the earth's surface.
+        let adjustedAltitude = altitude + location.verticalAccuracy
```
rolson, line 185: do we still want to do this? If so, should we only add half the accuracy?

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/WorldTrackingSceneView.swift
rolson: Maybe something to ask Koushik.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/WorldTrackingSceneView.swift
rolson: So zero would be perfectly precise then? If so, wouldn't we want to update the location in that case?

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
```diff
@@ -0,0 +1,274 @@
...
+                Text("V. Accuracy: \(currentLocation.verticalAccuracy.formatted(.number.precision(.fractionLength(2))))")
+            }
+            .multilineTextAlignment(.center)
+            .padding(8)
+            .frame(maxWidth: .infinity, alignment: .center)
+            .background(.regularMaterial)
+            .clipShape(RoundedRectangle(cornerRadius: 10))
+            .padding([.horizontal, .top])
+        }
+    }
```
rolson, line 188: I don't think the location datasource is used other than for this, so it would be nice to clean this up and put it at the app level if it's helpful for debugging.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/GeoTrackingSceneView.swift
yo1995: I feel it's more like a timing issue. Will leave it to more testing. 🤔

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/GeoTrackingSceneView.swift
```diff
@@ -0,0 +1,204 @@
...
+                            }
+                        case .localized:
+                            // Update the camera controller every time geo-tracking is localized,
+                            // to ensure the best experience.
+                            if !initialCameraIsSet, let currentLocation, let currentHeading {
+                                // Set the initial heading of scene view camera based on location
+                                // and heading. Geo-tracking requires 90 degrees rotation.
+                                updateCameraController(
+                                    location: currentLocation,
+                                    heading: currentHeading + 90,
```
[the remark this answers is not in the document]
yo1995: I think this is incorrect, but I'm still trying to figure out what is the right way to get the heading aligned 😥 .

When I use the replay footage, the heading difference is a fixed non-zero value regardless of heading changes, which makes me confused.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/WorldTrackingSceneView.swift
yo1995: Actually I'm seeing a difference between the 3D point's z value from the System LDS, and the point after I project it into the scene's SR. I'll reach out and ask.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/WorldTrackingSceneView.swift
yo1995: > The accuracy indicates the radius of the location display circle. A negative value indicates that the latitude and longitude are invalid. [doc](https://developer.apple.com/documentation/corelocation/cllocation/1423599-horizontalaccuracy#discussion)

Greater than or equal to `zero` means that when they are negative, we can return early because the location is invalid.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/WorldTrackingSceneView.swift
yo1995: After talking to Koushik, I'm pretty sure we can use the z value from GPS directly as both Apple and us are representing it in MSL. I'll remove this logic

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
yo1995: OK, we can take another look when we tackle the "status view" issue, #4515.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/WorldTrackingSceneView.swift
rolson: > Greater than or equal to zero 

but the check is > not >=

yo1995: Fixed in 2ba0fe3

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/GeoTrackingSceneView.swift
yo1995: aa77704077e3ba7b69aa881a12357e3dbe1b984f , will need a better place to put it and a better name…

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/GeoTrackingSceneView.swift
yo1995: I'll leave it unchanged for now. In my tests, once the geo-tracking is localized, it doesn't need adjustments and is pretty accurate, so maybe eventually we don't need this for geo-tracking.

The problem is I cannot find the right way to set the initial heading correctly. 😢 Will discuss in Slack tomorrow.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/WorldTrackingSceneView.swift
```diff
@@ -192,27 +167,24 @@ public struct WorldScaleGeoTrackingSceneView: View {
...
     /// - Parameter location: The location data source location.
-    @MainActor
-    private func updateSceneView(for location: Location) {
+    private func updateWorldTrackingSceneView(for location: Location) {
         // Do not update the scene view when the coaching overlay is in place.
         guard !coachingOverlayIsActive else { return }
         
-        // Do not use cached location more than 10 seconds old.
-        guard abs(lastLocationTimestamp?.timeIntervalSinceNow ?? 0) < 10 else { return }
+        // Update if the location is more than 10 seconds old.
```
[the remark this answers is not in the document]
yo1995: Maybe I can revert to the old comment.

I mean "refresh the stale location if it is older than 10 seconds". It may be clearer than the current comment.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/WorldTrackingSceneView.swift
```diff
@@ -16,74 +16,85 @@ import ARKit
...
-/// A scene view that provides an augmented reality world scale experience using geotracking.
-public struct WorldScaleGeoTrackingSceneView: View {
-    /// The proxy for the ARSwiftUIView.
-    @State private var arViewProxy = ARSwiftUIViewProxy()
-    /// The camera controller that will be set on the scene view.
-    @State private var cameraController: TransformationMatrixCameraController
+/// A scene view that provides an augmented reality world scale experience using world-tracking.
+struct WorldTrackingSceneView: View {
+    /// A Boolean value indicating if the camera was initially set.
+    @Binding var initialCameraIsSet: Bool
```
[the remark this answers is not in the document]
yo1995: Do you suggest me to initialize this parameter in the initializer?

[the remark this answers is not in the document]
des12437: No, this line can be left as-is.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/GeoTrackingSceneView.swift
```diff
@@ -0,0 +1,194 @@
...
+            // to ensure the best experience.
+            if !initialCameraIsSet, let currentLocation, let currentHeading {
+                // Set the initial heading of scene view camera based on location
+                // and heading. Geo-tracking requires 90 degrees rotation.
+                updateCameraController(
+                    location: currentLocation,
+                    heading: currentHeading + 90,
+                    altitude: currentLocation.position.z ?? 0
+                )
+                initialCameraIsSet = true
```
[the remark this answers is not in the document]
rolson: let's create an issue for followup

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
```diff
@@ -0,0 +1,275 @@
...
+            comment: """
+                 A label for a button to show the calibration view.
+                 """
+        )
+    }
+}
+
+extension SceneView {
+    /// A modifier to combine various modifiers that apply to both world and geo tracking scene view.
+    func worldScaleViewModifier(cameraController: TransformationMatrixCameraController) -> SceneView {
```
rolson, line 267: Can we call this `worldScaleSetup`

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/GeoTrackingSceneView.swift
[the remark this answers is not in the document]
yo1995: issue 4991

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/GeoTrackingSceneView.swift
yo1995: issue 4992

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/GeoTrackingSceneView.swift
yo1995: issue 4991

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/GeoTrackingSceneView.swift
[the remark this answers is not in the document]
yo1995: I'll ask on the developer forum. Probably it is related to the [`eulerAngles`](https://developer.apple.com/documentation/arkit/arcamera/2866109-eulerangles) of the AR camera, but I tried a few calculations to no avail.

rolson (approved):

des12437 (approved):

PR #583, opened 2024-02-09
Remove SwiftUI preview

des12437:
Removes SwiftUI preview which causes build error when running a build script.


yo1995 (approved):
Would be good to find out if this is a known issue, or file a FB to Apple.

rolson (approved):
Thanks for fixing!

PR #584, opened 2024-02-13
AR: Add raycast for world scale screen point to scene point

yo1995:
## Description

This PR adds the equivalent of https://github.com/Esri/arcgis-runtime-toolkit-ios/blob/main/Sources/ArcGISToolkit/AR/ArcGISARView.swift#L220-L238 to translate a screen point to a scene point. It may be useful for adding real world anchor/pin by tapping on the screen.

## To Discuss

- All "HitTest" phrasing are reworded - https://github.com/Esri/arcgis-runtime-toolkit-ios/pull/181 Apple deprecated the older hit test methods in iOS 14 and moved to all raycasting.
- See other comments in the code.


Sources/ArcGISToolkit/Components/Augmented Reality/ARSwiftUIView.swift
```diff
@@ -153,3 +154,40 @@ class ARSwiftUIViewProxy: NSObject, ARSessionProviding {
...
+extension ARSwiftUIViewProxy {
+    /// Performs a raycast to get the transformation matrix representing the corresponding 
+    /// real-world point for `screenPoint`.
+    /// - Parameter screenPoint: The screen point to determine the real world transformation matrix from.
+    /// - Returns: A `TransformationMatrix` representing the real-world point corresponding to `screenPoint`.
+    func raycast(from screenPoint: CGPoint) -> TransformationMatrix? {
+        // Use the `raycastQuery` method on ARSCNView to get the location of `screenPoint`.
+        guard let query = raycastQuery(
+            from: screenPoint,
+            allowing: .existingPlaneGeometry,
```
yo1995, line 167: https://developer.apple.com/documentation/arkit/arraycastquery/target/estimatedplane

According to this doc, maybe we should use `estimatedPlane` for real world, and `existingPlaneGeometry` for table top.

Sources/ArcGISToolkit/Components/Augmented Reality/ARSwiftUIView.swift
```diff
@@ -153,3 +154,40 @@ class ARSwiftUIViewProxy: NSObject, ARSessionProviding {
         )
     }
 }
+
+extension ARSwiftUIViewProxy {
+    /// Performs a raycast to get the transformation matrix representing the corresponding 
+    /// real-world point for `screenPoint`.
+    /// - Parameter screenPoint: The screen point to determine the real world transformation matrix from.
+    /// - Returns: A `TransformationMatrix` representing the real-world point corresponding to `screenPoint`.
+    func raycast(from screenPoint: CGPoint) -> TransformationMatrix? {
```
yo1995, line 163: Since this method is used by both tabletop and world scale, I moved it here as an internal extension.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -323,3 +323,31 @@ private extension WorldScaleGeoTrackingSceneView {
...
+
+public extension WorldScaleGeoTrackingSceneView {
+    /// Determines the scene point for the given screen point.
+    /// - Parameter screenPoint: The point in screen's coordinate space.
+    /// - Returns: The scene point corresponding to screen point.
+    func arScreenToLocation(screenPoint: CGPoint) -> Point? {
+        // Use the `raycast` method to get the matrix of `screenPoint`.
+        guard let localOffsetMatrix = arViewProxy.raycast(from: screenPoint) else { return nil }
+        
+        let translationFactor = cameraController.translationFactor
```
yo1995, line 335: For world scale AR, the translation factor should always be 1. I'll change it in the next commit.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleGeoTrackingSceneView.swift
```diff
@@ -323,3 +323,17 @@ private extension WorldScaleGeoTrackingSceneView {
         )
     }
 }
+
+public extension WorldScaleGeoTrackingSceneView {
```
yo1995, line 327: This is not the right place for the extension. Should be on `ARSwiftUIViewProxy`

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
```diff
@@ -134,22 +135,41 @@ class ARSwiftUIViewProxy: NSObject, ARSessionProviding {
...
+extension ARSwiftUIViewProxy {
+    /// Performs a raycast to get the transformation matrix representing the corresponding 
+    /// real-world point for `screenPoint`.
+    /// - Parameter screenPoint: The screen point to determine the real world transformation matrix from.
+    /// - Returns: A `TransformationMatrix` representing the real-world point corresponding to `screenPoint`.
+    func raycast(from screenPoint: CGPoint) -> TransformationMatrix? {
+        // Use the `raycastQuery` method on ARSCNView to get the location of `screenPoint`.
+        guard let query = arView.raycastQuery(
+            from: screenPoint,
+            allowing: .existingPlaneGeometry,
```
yo1995, line 149: I'm a little bit hesitant to make a change here. According to the doc, an [existingPlaneGeometry](https://developer.apple.com/documentation/arkit/arraycastquery/target/existingplanegeometry) is for "a plane to have a definitive size and shape", such as a table top. An [estimatedPlane](https://developer.apple.com/documentation/arkit/arraycastquery/target/estimatedplane) is for nonplanar surfaces, such as a real-world surface. 

I was tempted to change the method to take the allowed target plane as a parameter, so that table top AR can pass in `existingPlaneGeometry` and world scale AR can use `estimatedPlane`. Do you think that would be a good idea?

It may be hard to quantify how much effect it has on the raycasting accuracy in real world.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
```diff
@@ -262,6 +262,31 @@ private extension WorldScaleSceneView {
     }
 }
 
+public extension WorldScaleSceneView {
+    /// Determines the scene point for the given screen point.
+    /// - Parameter screenPoint: The point in screen's coordinate space.
+    /// - Returns: The scene point corresponding to screen point.
+    private func arScreenToLocation(screenPoint: CGPoint) -> Point? {
```
yo1995, line 269: According to the doc, this method is only used for the identifying workflow: https://developers.arcgis.com/ios/scenes-3d/display-scenes-in-augmented-reality/#identify-real-world-and-in-scene-objects

So I didn't expose it to the user, but rather provided a custom tap modifier.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
```diff
@@ -262,6 +262,31 @@ private extension WorldScaleSceneView {
...
+    
+    /// Sets the closure to call when the scene's nodes are updated.
+    func onARScreenSingleTapGesture(
+        perform action: @escaping (_ screenPoint: CGPoint, _ scenePoint: Point?) -> Void
+    ) -> Self {
+        self
+            .onSingleTapGesture { tapPoint in
+                let scenePoint = arScreenToLocation(screenPoint: tapPoint)
+                action(tapPoint, scenePoint)
+            } as! WorldScaleSceneView
```
yo1995, line 286: The table top view doesn't force cast it to the type itself and returned `some View`. Do you think it has value to force cast here, or returning an opaque type is fine?

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
```diff
@@ -134,22 +135,41 @@ class ARSwiftUIViewProxy: NSObject, ARSessionProviding {
...
         arView.session
     }
-    
-    /// Creates a raycast query that originates from a point on the view, aligned with the center of the camera's field of view.
-    /// - Parameters:
-    ///   - point: The point on the view to extend the raycast from.
-    ///   - target: The type of surface the raycast can interact with.
-    ///   - alignment: The target's alignment with respect to gravity.
-    /// - Returns: An `ARRaycastQuery`.
-    func raycastQuery(
```
yo1995, line 144: This method was only used by the `raycast(from:)` method below, so I condensed it away.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
des12437: > I was tempted to change the method to take the allowed target plane as a parameter, so that table top AR can pass in existingPlaneGeometry and world scale AR can use estimatedPlane. Do you think that would be a good idea?

Yes I agree that the allowed target plane should be a parameter. I tested both target types and .estimatedPlane is better for a real-world surface.

Sources/ArcGISToolkit/Extensions/SwiftUI/View.swift
```diff
@@ -152,3 +152,28 @@ extension View {
...
 }
+
+extension View {
+    /// Sets a closure to perform when a single tap occurs on the view.
+    ///
+    /// - Note: This is to retrofit the tap gesture to iOS 15.0.
+    /// - Parameters:
+    ///   - action: The closure to perform upon single tap.
+    ///   - screenPoint: The location of the tap in the view's coordinate space.
+    func onSingleTapGesture(perform action: @escaping (_ screenPoint: CGPoint) -> Void) -> some View {
```
rolson, line 163: I have no problem if we just want to make the single tap functionality only available on iOS 16 and up. We are dropping support for iOS 15 after this April release.

yo1995: Since this extension is moved from the tabletop implementation, I'm OK with keeping the iOS 15 compatibility code, as it is not too verbose.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
```diff
@@ -262,6 +262,31 @@ private extension WorldScaleSceneView {
...
+        guard let localOffsetMatrix = arViewProxy.raycast(from: screenPoint, allowing: .estimatedPlane) else { return nil }
+        let originTransformationMatrix = cameraController.originCamera.transformationMatrix
+        let scenePointMatrix = originTransformationMatrix.adding(localOffsetMatrix)
+        // Create a camera from transformationMatrix and return its location.
+        return Camera(transformationMatrix: scenePointMatrix).location
+    }
+    
+    /// Sets a closure to perform when a single tap occurs on the view.
+    func onSingleTapGesture(
+        perform action: @escaping (_ screenPoint: CGPoint, _ scenePoint: Point?) -> Void
```
rolson, line 280: in what cases will this `scenePoint` be `nil`?

yo1995: When `TransformationMatrix` is `nil`, which can be caused by a failed [raycast](https://developer.apple.com/documentation/arkit/arscnview/3194595-raycastquery). Or in other words, if raycast query result is `nil`, the scene point is `nil`.

des12437 (dismissed):

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
rolson: I see that. I was wondering in what cases the recast query result is nil. Thank you

yo1995: I added some explanation to the docstrings.

rolson (approved):

PR #593, opened 2024-02-27
`FeatureFormView` - Improve redraw

dfeinzimer:
Apollo 513

### DIFF
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/FormInputs/InputFooter.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/FormInputs/InputFooter.swift
@@ -205,7 +205,7 @@ extension InputFooter {
     var isShowingError: Bool {
         element.isEditable 
         && primaryError != nil
-        && (model.previouslyFocusedFields.contains(element) || validationErrorVisibility == .visible)
+        && (model.previouslyFocusedElements.contains(element) || validationErrorVisibility == .visible)
     }
     
     /// The allowable number of characters in the input.
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/FormInputs/InputFooter.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/FormInputs/InputFooter.swift
@@ -26,6 +26,11 @@ struct InputFooter: View {
     /// The form element the footer belongs to.
     let element: FieldFormElement
     
+    /// An ID regenerated each time the element's value changes.
+    ///
+    /// Allows the footer to be recomputed to reflect changes in validation errors or input length.
+    @State private var id = UUID()
+    
     var body: some View {
         HStack(alignment: .top) {
             Group {
@@ -56,6 +61,12 @@ struct InputFooter: View {
         }
         .font(.footnote)
         .foregroundColor(isShowingError ? .red : .secondary)
+        .id(id)
+        .task {
+            for await _ in element.$value {
+                id = UUID()
+            }
+        }
     }
 }
 
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/FormViewModel.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/FormViewModel.swift
@@ -39,9 +39,6 @@ import SwiftUI
     /// The list of visible form elements.
     @Published var visibleElements = [FormElement]()
     
-    /// The list of expression evaluation errors.
-    @Published var expressionEvaluationErrors = [FormExpressionEvaluationError]()
-    
     /// The set of all fields which previously held focus.
     @Published var previouslyFocusedFields = [FormElement]()
     
@@ -79,8 +76,7 @@ import SwiftUI
     
     /// Performs an initial evaluation of all form expressions.
     func initialEvaluation() async {
-        let evaluationErrors = try? await featureForm.evaluateExpressions()
-        expressionEvaluationErrors = evaluationErrors ?? []
+        _ = try? await featureForm.evaluateExpressions()
         initializeIsVisibleTasks()
     }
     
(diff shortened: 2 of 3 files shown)


Sources/ArcGISToolkit/Components/FeatureFormView/FormViewModel.swift
```diff
@@ -90,11 +87,11 @@ import SwiftUI
...
         evaluateTask?.cancel()
-        isEvaluating = true
         evaluateTask = Task {
-            let evaluationErrors = try? await featureForm.evaluateExpressions()
-            expressionEvaluationErrors = evaluationErrors ?? []
-            isEvaluating = false
+            if let evaluationErrors = try? await featureForm.evaluateExpressions(),
+               !evaluationErrors.isEmpty {
+                expressionEvaluationErrors = evaluationErrors
+            }
```
mhdostal, line 94: What about the case where there previously WERE evaluation errors (`expressionEvaluationErrors` is not empty), but the next call to `evaluateExpressions()` returns an empty array. In that case, the non-empty list of `expressionEvaluationErrors` would not be cleared (because we're only setting it if `evaluationErrors` is not empty).

You would want to update `expressionEvaluationErrors` only if it was not equal to the new list of `evaluationErrors`. I don't know if arrays or `Errors` are `Equatable`...

Sources/ArcGISToolkit/Components/FeatureFormView/FormViewModel.swift
```diff
@@ -42,9 +42,6 @@ import SwiftUI
     /// The list of expression evaluation errors.
     @Published var expressionEvaluationErrors = [FormExpressionEvaluationError]()
     
-    /// A Boolean value indicating whether evaluation is running.
-    @Published var isEvaluating = true
```
mhdostal, line 46: This was originally going to be used for displaying a progress indicator for long-running expressions. However, it's currently not being used, so I'm OK with removing it.

mhdostal (commented):
You mention in the issue:

> FormViewModel.evaluateExpressions() is called upon any user interaction with a field.

Is that an issue? We'd only want `evaluateExpressions()` to be called after the user changes the value. Does it happen in other cases, like when just getting focus?

dfeinzimer:
> Is that an issue? We'd only want `evaluateExpressions()` to be called after the user changes the value.

Sorry that's what I meant by "interaction" - interaction by changing the value.

> Does it happen in other cases, like when just getting focus?

The only other time expressions are evaluated is when the form view loads in `FormViewModel.initialEvaluation()`

Sources/ArcGISToolkit/Components/FeatureFormView/FormViewModel.swift
dfeinzimer: > In that case, the non-empty list of expressionEvaluationErrors would not be cleared.

That's a good point.

> I don't know if arrays or Errors are Equatable...

Arrays should be if their content is but in this case neither `FormExpressionEvaluationError` nor its underlying `Error` is.

dfeinzimer: I see `expressionEvaluationErrors` is not currently being read either. We could consider removing its `@Published` attribute or removing the property all together. Off the top of my head I'm not thinking of where we'd need to use it.

dfeinzimer: Kotlin isn't currently the errors and they don't seem to have a use case presently defined so I'll remove the property for now.

Ref Apollo 271, 280

dfeinzimer:
Taking this out of drafts now. There was some additional handling I needed to add into the footer to account for changes that was previously being handled by the entire form redrawing.

All tests are passing again (using `4160`).

There is a crash in core I'm seeing when we call evaluate expressions that happens fairly ofter (but not all the time) when I run the tests:

    terminating due to uncaught exception of type boost::wrapexcept<boost::lock_error>: boost: mutex lock failed in pthread_mutex_lock: Invalid argument [generic:22]

I'm not sure yet if it predates these changes.

dfeinzimer:
> I'm not sure yet if it predates these changes.

I've verified that it's unrelated to these changes and opened Apollo 519.

Sources/ArcGISToolkit/Components/FeatureFormView/FormInputs/InputFooter.swift
```diff
@@ -26,6 +26,11 @@ struct InputFooter: View {
     /// The form element the footer belongs to.
     let element: FieldFormElement
     
+    /// An ID regenerated each time the element's value changes.
+    ///
+    /// Allows the footer to be recomputed to reflect changes in validation errors or input length.
+    @State private var id = UUID()
```
mhdostal, line 32: This is convenient, but I'm wondering if instead of a special-case `@State var` we should be looking at the `.onChanged` event for `element.validationErrors` and triggering off of that. Not sure about how to handle `input length`... thoughts?

dfeinzimer: > I'm wondering if instead of a special-case `@State var` we should be looking at the `.onChanged` event for `element.validationErrors` and triggering off of that.

We'd need to tackle the `Error` equitability issue to watch for new errors with `onChange`. 

Alternatively we could have another `task` awaiting `element.$validationErrors` but we'd also need an additional property to store received errors to make them available to `var primaryError`.  Also, that `task` would *only* help us in collecting the latest validation errors, we'd still need to handle character count changes.

The current changes (`.task { for await _ in element.$value { ...`) helps us keep both the validation errors and the character count refreshed.

dfeinzimer:
There're two other properties on `FormViewModel` causing full redraws, `focusedElement` and `previouslyFocusedFields`. Does redraw on focus change sound like something we should mitigate?

mhdostal (approved):

mhdostal:
> Does redraw on focus change sound like something we should mitigate?
Yes, but I don't think it's a high priority. Enter another, low-priority, issue and we can get to it when we have time.

It won't impact the performance as much as what you fixed in this PR. Thanks!

dfeinzimer:
> Yes, but I don't think it's a high priority. Enter another, low-priority, issue and we can get to it when we have time.

Done, Apollo 530

PR #597, opened 2024-03-01
Combo Box Clear button should only show if the field is not required

mhdostal:
Apollo # 501

Only show the Clear button in the combo box input if the field is not required AND there is an existing value.


dfeinzimer (commented):
Changes in the component look good but this breaks test 3.5. I've created a PR to update the test design in common TK.

Also it looks like this PR may include changes previously merged in #595

mhdostal:
> Changes in the component look good but this breaks test 3.5. I've created a PR to update the test design in common TK.

Fixed.
 
> Also it looks like this PR may include changes previously merged in #595

I fixed this too.

dfeinzimer (approved):

PR #598, opened 2024-03-01
Apply fixes found during Dev Summit prep

dfeinzimer:
Various small fixes around the Toolkit found during Dev Summit prep.

cc @philium

### DIFF
--- a/Examples/Examples/BasemapGalleryExampleView.swift
+++ b/Examples/Examples/BasemapGalleryExampleView.swift
@@ -30,7 +30,7 @@ struct BasemapGalleryExampleView: View {
     )
     
     /// The initial list of basemaps.
-    @State private var basemaps = initialBasemaps()
+    @State private var basemaps = makeBasemapGalleryItems()
     
     var body: some View {
         MapView(map: map, viewpoint: initialViewpoint)
@@ -63,7 +63,7 @@ struct BasemapGalleryExampleView: View {
         }
     }
     
-    private static func initialBasemaps() -> [BasemapGalleryItem] {
+    private static func makeBasemapGalleryItems() -> [BasemapGalleryItem] {
         let identifiers = [
             "46a87c20f09e4fc48fa3c38081e0cae6",
             "f33a34de3a294590ab48f246e99958c9",
--- a/Examples/Examples/FloorFilterExampleView.swift
+++ b/Examples/Examples/FloorFilterExampleView.swift
@@ -18,7 +18,7 @@ import ArcGIS
 
 struct FloorFilterExampleView: View {
     /// Determines the arrangement of the inner `FloorFilter` UI components.
-    private var floorFilterAlignment: Alignment { .bottomLeading }
+    let floorFilterAlignment = Alignment.bottomLeading
     
     /// The height of the map view's attribution bar.
     @State private var attributionBarHeight = 0.0
--- a/Examples/Examples/SearchExampleView.swift
+++ b/Examples/Examples/SearchExampleView.swift
@@ -18,7 +18,7 @@ import ArcGISToolkit
 
 struct SearchExampleView: View {
     /// Provides search behavior customization.
-    let locatorSearchSource = SmartLocatorSearchSource(
+    @State private var locatorSearchSource = SmartLocatorSearchSource(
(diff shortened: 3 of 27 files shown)


Examples/Examples/BasemapGalleryExampleView.swift
```diff
@@ -63,7 +63,7 @@ struct BasemapGalleryExampleView: View {
         }
     }
     
-    private static func initialBasemaps() -> [BasemapGalleryItem] {
+    private static func makeBasemapGalleryItems() -> [BasemapGalleryItem] {
```
dfeinzimer, line 66: This name more closely matches the type returned by the method.

Examples/Examples/FloorFilterExampleView.swift
```diff
@@ -18,7 +18,7 @@ import ArcGIS
 
 struct FloorFilterExampleView: View {
     /// Determines the arrangement of the inner `FloorFilter` UI components.
-    private var floorFilterAlignment: Alignment { .bottomLeading }
+    let floorFilterAlignment = Alignment.bottomLeading
```
dfeinzimer, line 21: No reason for this to be computed. It never changes.

Examples/Examples/SearchExampleView.swift
```diff
@@ -18,7 +18,7 @@ import ArcGISToolkit
 
 struct SearchExampleView: View {
     /// Provides search behavior customization.
-    let locatorDataSource = SmartLocatorSearchSource(
+    @State private var locatorSearchSource = SmartLocatorSearchSource(
```
dfeinzimer, line 21: Making this a state property would prevent it from being recreated _if_ this were in a view that was often recomputed. The new name also more closely matches its type.

Examples/Examples/SearchExampleView.swift
```diff
@@ -30,7 +30,7 @@ struct SearchExampleView: View {
     )
     
     /// The `GraphicsOverlay` used by the `SearchView` to display search results on the map.
-    private let searchResultsOverlay = GraphicsOverlay()
+    @State private var searchResultsOverlay = GraphicsOverlay()
```
dfeinzimer, line 33: Same reason as changing `locatorSearchSource` to a state property above.

Examples/Examples/UtilityNetworkTraceExampleView.swift
```diff
@@ -26,19 +26,19 @@ struct UtilityNetworkTraceExampleView: View {
     @State private var map = makeMap()
     
     /// The current detent of the floating panel presenting the trace tool.
-    @State var activeDetent: FloatingPanelDetent = .half
+    @State private var activeDetent: FloatingPanelDetent = .half
```
dfeinzimer, line 29: All the state properties in this file should've been marked private.

Examples/Examples/FloorFilterExampleView.swift
yo1995: Should it still be private?

Sources/ArcGISToolkit/Documentation.docc/Resources/FloorFilter/FloorFilterStep2.swift
```diff
@@ -3,7 +3,7 @@ import ArcGISToolkit
 import ArcGIS
 
 struct FloorFilterExampleView: View {
-    private var floorFilterAlignment: Alignment { .bottomLeading }
+    let floorFilterAlignment = Alignment.bottomLeading
```
yo1995, line 6: Same question here and below

Examples/Examples/FloorFilterExampleView.swift
philium: > It never changes.

That is precisely why it should be computed: so no memory is allocated for it each time the example view is instantiated. Either that or it should be static. Same with similar changes in this PR.

yo1995 (approved):

Examples/Examples/FloorFilterExampleView.swift
dfeinzimer: Ah okay, reverted that commit. Thanks for catching!

PR #603, opened 2024-03-06
Only show clear button in DateIinput if date is not required

mhdostal:
Finalizes the Clear button updates based on the latest design [here](https://devtopia.esri.com/runtime/common-toolkit/blob/main/designs/Forms/InputValidationDesign.md#clear-button-usage).


dfeinzimer (approved):
Looks good

PR #615, opened 2024-03-15
AR: Refactor calibration view

des12437:
Closes `swift/4992`.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
@@ -33,7 +33,7 @@ public struct WorldScaleSceneView: View {
     /// The camera controller that will be set on the scene view.
     @State private var cameraController: TransformationMatrixCameraController
     /// The view model for the calibration view.
-    @StateObject private var calibrationViewModel = WorldScaleCalibrationViewModel()
+    @State private var calibrationViewModel: WorldScaleCalibrationViewModel
     /// A Boolean value that indicates whether the geo-tracking configuration is available.
     @State private var geoTrackingIsAvailable = true
     /// A Boolean value that indicates whether the initial camera is set for the scene view.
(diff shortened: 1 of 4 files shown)


yo1995 (approved):

rolson (commented):
I think perhaps a better way to do this is to remove the dependency on Combine altogether. I will put a PR up against your branch.

rolson (approved):

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
```diff
@@ -33,7 +33,7 @@ public struct WorldScaleSceneView: View {
     /// The camera controller that will be set on the scene view.
     @State private var cameraController: TransformationMatrixCameraController
     /// The view model for the calibration view.
-    @StateObject private var calibrationViewModel = WorldScaleCalibrationViewModel()
+    @State private var calibrationViewModel: WorldScaleCalibrationViewModel
```
yo1995, line 36: Sorry for the delayed response. The thing that held me back a few days ago was the decision on whether this should be a state or a state object. Since this type is a class, shall we use StateObject [initialization](https://developer.apple.com/documentation/swiftui/stateobject#Initialize-state-objects-using-external-data)?

des12437: This suggestions works, although the variable needs to be declared as a `@StateObject` in order to use StateObject [initialization](https://developer.apple.com/documentation/swiftui/stateobject#Initialize-state-objects-using-external-data). I can make this change.

rolson: I don't think it needs to be a `StateObject` here because it's not being observed here. It's being observed in other views. That's why I changed it to `State`. Also recall other classes (ie `Map`) are `@State`.

yo1995: I agree now. We aren't using StateObject's observation capability in `WorldScaleSceneView`, so StateObject has no benefit over State in this particular case.

yo1995 (approved):

PR #620, opened 2024-03-18
`FeatureFormView` - Repair test case 6.2

dfeinzimer:


mhdostal (approved):

dfeinzimer:
Ready for re-review. Added an additional test to verify correct fallback behavior.

mhdostal (approved):
Does the core code for radio buttons fall back to a combo box in certain situations or is that all here in the TK?

dfeinzimer:
> Does the core code for radio buttons fall back to a combo box in certain situations or is that all here in the TK?

I don't believe so but I'm checking on #Apollo to confirm. Will perform any followup work in another PR if needed.

PR #621, opened 2024-03-18
AR: Fix safe area

des12437:
This PR changes the safe area from ignoring all edges to only the bottom edge. When the top edge safe area is ignored when the `WorldScaleSceneView` is used with a navigation bar, the navigation bar is transparent and the camera feed is visible:

Screenshot 2024-03-18 at 1 31 32 PM

This safe area change fixes the navigation bar appearance and looks like this when there is no navigation bar present:
Screenshot 2024-03-18 at 1 32 49 PM

### DIFF
--- a/Examples/Examples/WorldScaleExampleView.swift
+++ b/Examples/Examples/WorldScaleExampleView.swift
@@ -63,7 +63,7 @@ struct WorldScaleExampleView: View {
         .onSingleTapGesture { _, scenePoint in
             graphicsOverlay.addGraphic(Graphic(geometry: scenePoint))
         }
-        .ignoresSafeArea(edges: .bottom)
+        .ignoresSafeArea(edges: [.horizontal, .bottom])
         .task {
             // Request when-in-use location authorization.
             // This is necessary for 2 reasons:
--- a/Examples/Examples/WorldScaleExampleView.swift
+++ b/Examples/Examples/WorldScaleExampleView.swift
@@ -63,6 +63,7 @@ struct WorldScaleExampleView: View {
         .onSingleTapGesture { _, scenePoint in
             graphicsOverlay.addGraphic(Graphic(geometry: scenePoint))
         }
+        .ignoresSafeArea(edges: .bottom)
         .task {
             // Request when-in-use location authorization.
             // This is necessary for 2 reasons:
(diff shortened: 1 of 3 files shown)


yo1995 (approved):

yo1995 (dismissed):
Sorry for suggesting deprecated API! The new API is correct.

rolson (commented):
I'm not sure it makes sense to do this here. I removed these lines:

```swift
.ignoresSafeArea(.all)
```

And put it in the example (app level) and it works as expected:

```swift
    var body: some View {
        WorldScaleSceneView(trackingMode: .worldTracking) { proxy in
            SceneView(scene: scene, graphicsOverlays: [graphicsOverlay])
                .onSingleTapGesture { screen, _ in
                    print("Identifying...")
                    Task {
                        let results = try await proxy.identifyLayers(screenPoint: screen, tolerance: 20)
                        print("\(results.count) identify result(s).")
                    }
                }
        }
        .calibrationButtonAlignment(.bottomLeading)
        .onSingleTapGesture { _, scenePoint in
            graphicsOverlay.addGraphic(Graphic(geometry: scenePoint))
        }
        .ignoresSafeArea(edges: .bottom)
        ...
```

des12437:
> I'm not sure it makes sense to do this here. I removed these lines:
> 
> ```swift
> .ignoresSafeArea(.all)
> ```

I'm in favor of removing this from the AR scene views so the safe area can be handled at the app level.

Examples/Examples/WorldScaleExampleView.swift
```diff
@@ -63,6 +63,7 @@ struct WorldScaleExampleView: View {
         .onSingleTapGesture { _, scenePoint in
             graphicsOverlay.addGraphic(Graphic(geometry: scenePoint))
         }
+        .ignoresSafeArea(edges: .bottom)
```
yo1995, line 66:
```suggestion
        .ignoresSafeArea(edges: [.horizontal, .bottom])
```
IMG_3580

In the rare case when the app is used in landscape mode, we still need horizontal safe area ignore. Otherwise it won't extend into the shorter edges of the screen.

yo1995 (approved):

PR #623, opened 2024-03-19
world-scale: animate the scene view in

rolson:
world-scale: animate the scene view in


des12437 (dismissed):

yo1995 (approved):
With the default animation speed it looks pretty much like a jump-in rather than a fade-in though.

rolson:
> With the default animation speed it looks pretty much like a jump-in rather than a fade-in though.

Good point. I slowed it down. Please re-review.

des12437 (approved):

yo1995 (approved):

PR #627, opened 2024-03-19
`FeatureFormView` - Additional revisions pre-merge

dfeinzimer:
Additional debt cleanup in preparation of merge:
- Removes an accidental line addition in `UtilityNetworkTraceExampleView`, currently causing a warning
- Fixes a few inconsistencies in where the `accessibilityIdentifier` modifier is placed
- Applies common ordering for property declarations (attributed properties in alphabetical order followed by `let`s in alphabetical order)
- Removes types from property declarations where it is easily inferred from the initial value
- Synchronizes doc for all `element`, `formattedValue`, `isRequired`, and `value` properties
- Removes the since tag on `FormViewModel` which is no longer public
- Remove the 2nd `InputHeader` initializer, resolving a bug where a `DateTimeInput`'s header wouldn't properly update when the field became required
- Removes 3 lines from the 2nd `ComboBoxInput` initializer setting the initial values of some state properties, which shouldn't be done from an initializer

### DIFF
--- a/Examples/Examples/UtilityNetworkTraceExampleView.swift
+++ b/Examples/Examples/UtilityNetworkTraceExampleView.swift
@@ -20,7 +20,6 @@ import SwiftUI
 /// a utility network and trace configurations.
 struct UtilityNetworkTraceExampleView: View {
     @Environment(\.isPortraitOrientation)
-    
     private var isPortraitOrientation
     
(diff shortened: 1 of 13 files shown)


mhdostal (approved):
The changes look great, thanks!

PR #643, opened 2024-03-25
`FeatureFormView` - Improve file organization

dfeinzimer:
Improves the organization of the component's files which are currently all splayed across two directories

| Before | After |
|---|---|
| before | after |


mhdostal (approved):
Looks much more understandable, thanks!!

PR #646, opened 2024-03-27
AR: Rename modifier to `onCalibratingChanged`

des12437:
This PR renames a modifier from `onCalibrationViewVisibilityChanged` to `onCalibratingChanged`.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
@@ -222,8 +222,8 @@ public struct WorldScaleSceneView: View {
         return view
     }
     
-    /// Sets a closure to perform when the calibration view visibility changes.
-    /// - Parameter action: The closure to perform when the calibration view visibility has changed.
+    /// Sets a closure to perform when calibration begins or ends.
+    /// - Parameter action: The closure to perform when calibration begins or ends.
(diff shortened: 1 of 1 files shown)


rolson (approved):

yo1995 (approved):

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
```diff
@@ -222,15 +222,15 @@ public struct WorldScaleSceneView: View {
...
     }
     
-    /// Sets a closure to perform when the calibration view visibility changes.
-    /// - Parameter action: The closure to perform when the calibration view visibility has changed.
-    public func onCalibrationViewVisibilityChanged(
+    /// Sets a closure to perform when calibration begins or ends.
+    /// - Parameter action: The closure to perform when calibration begins or ends.
+    public func onCalibratingChanged(
         perform action: @escaping (
             _ isPresented: Bool
```
philium, line 229: This should be renamed as well. I would suggest `newCalibrating` as the new name—it matches the `newValue` name of the `onChange` modifier's closure parameter and it is less likely to conflict with the corresponding state property (the user may have a state property named `isCalibrating`; using that as the name of the parameter makes the user have to add `self.` to deconflict). We used that same naming scheme for some of our modifiers (like `onInteractingChanged`, for example).

PR #657, opened 2024-04-01
Fix `WorldScaleSceneView` safe area bug

philium:
This fixes a bug I noticed where the camera view did not extend into the safe area.

Before | Screenshot 2024-04-01 at 11 46 13
--- | ---
After | Screenshot 2024-04-01 at 11 48 24

### DIFF
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
@@ -122,6 +122,7 @@ public struct WorldScaleSceneView: View {
                 self.error = error
             }
         }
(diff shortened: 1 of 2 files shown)


dfeinzimer (approved):

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
```diff
@@ -122,6 +122,7 @@ public struct WorldScaleSceneView: View {
                 self.error = error
             }
         }
+        .ignoresSafeArea(.container, edges: [.horizontal])
```
philium, line 125: This was supposed to include `bottom`. 🤦🏼‍♂️ I'll submit another PR.

PR #659, opened 2024-04-01
Make `ARSwiftUIView.proxy` non-optional

philium:
`ARSwiftUIView`'s initializer isn't ever called without a proxy object.


dfeinzimer (commented):
Any downside to making `proxy` internal and using a synthesized initializer instead?

mhdostal (approved):

philium:
> Any downside to making `proxy` internal and using a synthesized initializer instead?

The only downside is that the doc comment goes away. That's why I left it. But if you're fine with that, I'm fine with doing it.

dfeinzimer (approved):
My hunch is that since it's internal, using a synthesized initializer is okay but I'll leave it up to you. No strong feeling either way.

philium:
I'll leave it as is since the compiler synthesized implementation would include all of the other properties as parameters as well.

PR #661, opened 2024-04-01
`FeatureFormView` - Stabilize test case 1.4

dfeinzimer:
This case has failed twice at this point over the last two days.

Bumping to 10 seconds has worked elsewhere in these tests.


mhdostal (approved):

mhdostal:
I wonder if we should just switch to a 10 second wait for all future tests. It won't increase the execution time if the checks complete within the original 5 seconds, but might prevent future issues like this.

PR #392, opened 2023-08-10
Migrate documentation to DocC

dfeinzimer:
Closes #383 

Migrates component documentation to DocC

The basic usage sections have been transformed into DocC interactive tutorials.

PR #682, opened 2024-04-12
`FloatingPanel` Revisions

dfeinzimer:
Responds to [this](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/392#discussion_r1295054040) and [this](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/392#discussion_r1294989490) comment.

### DIFF
--- a/Sources/ArcGISToolkit/Components/FloatingPanel/FloatingPanel.swift
+++ b/Sources/ArcGISToolkit/Components/FloatingPanel/FloatingPanel.swift
@@ -15,9 +15,7 @@
 import SwiftUI
 
 /// A floating panel is a view that overlays a view and supplies view-related
-/// content. For a map view, for instance, it could display a legend, bookmarks, search results, etc..
-/// Apple Maps, Google Maps, Windows 10, and Collector have floating panel
-/// implementations, sometimes referred to as a "bottom sheet".
+/// content. A map view, for instance, could display a legend, bookmarks, search results, etc..
 ///
 /// Floating panels are non-modal and can be transient, only displaying
 /// information for a short period of time like identify results,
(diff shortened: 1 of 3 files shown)


Sources/ArcGISToolkit/Components/FloatingPanel/FloatingPanel.swift
```diff
@@ -15,9 +15,7 @@
 import SwiftUI
 
 /// A floating panel is a view that overlays a view and supplies view-related
-/// content. For a map view, for instance, it could display a legend, bookmarks, search results, etc..
-/// Apple Maps, Google Maps, Windows 10, and Collector have floating panel
-/// implementations, sometimes referred to as a "bottom sheet".
```
dfeinzimer, line 20: Overall I'd say the concept of sheets are ubiquitous enough in mobile to the point that we don't need to mention these but if we do want to keep them I did some more research:

- Apple [doesn't seem to have Apple Maps trademarked](https://www.apple.com/legal/intellectual-property/trademark/appletmlist.html) so I think we're in the clear there. If it was trademarked I think we'd need to include a symbol according to [this](https://www.apple.com/legal/intellectual-property/guidelinesfor3rdparties.html#:~:text=On%20product%2C%20product%20documentation%2C%20or%20other%20product%20communications%20that%20will%20be%20distributed%20only%20in%20the%20United%20States%2C%20use%20the%20appropriate%20trademark%20symbol%20(TM%2C%20SM%2C%20%C2%AE)%20the%20first%20time%20the%20Apple%20trademark%20appears%20in%20the%20text%20of%20the%20advertisement%2C%20brochure%2C%20or%20other%20material.).

- Google [does](https://about.google/brand-resource-center/trademark-list/#:~:text=Google%20Maps%E2%84%A2%20mapping%20service) have Maps trademarked so I think we'd need to change this to "Google Maps™ mapping service" if we decide to keep it to be compliant with [this](https://about.google/brand-resource-center/rules/#:~:text=Spell%20and%20capitalize%20the%20trademark%20exactly%20as%20they%20are%20shown%20in%20the%20Google%20Trademarks%20List.).

- Windows 10 is a little murky, I didn't find good info.

mhdostal (approved):

PR #703, opened 2024-04-25
Address Main actor warnings

dfeinzimer:
Resolves the "Main actor-isolated class has different actor isolation from nonisolated superclass 'XCTestCase'; this is an error in Swift 6" warnings.

See also Swift 5188, 5218


yo1995 (approved):
Warnings are gone after changes.

CalebRas (approved):

PR #704, opened 2024-04-26
`FeatureFormView` - Refresh attachment display on add

dfeinzimer:
Related: Apollo 408

- Provides a callback to `AttachmentImportMenu` to use when a new attachment has been added.
- `AttachmentsFeatureElementView` uses the callback to update its list of attachment models.

Know issues:
  - Attachment naming isn't correct yet
  - Thumbnails don't seem to be rendering with correct display scale - appear blurry


mhdostal:
> * Thumbnails don't seem to be rendering with correct display scale - appear blurry

This is because for the `.preview` list type, the thumbnail size is 120 x 120, but the default thumbnail size argument for `AttachmentModel.load` is 40 x 40. I'll fix this in a separate PR.

mhdostal (approved):

PR #708, opened 2024-04-29
`FeatureFormView` - Improve new attachment button

dfeinzimer:
Early feedback indicated it's difficult to hit the new attachment button

Before

After


mhdostal (approved):
This looks fine, but I'm wondering if this is good enough to fix the original issue, which had to do with no attachments and having the "bottom bar" *potentially* covering the Add button:

_for existing attachement, looks really good.   With a longer form and no existing attachments it is difficult to click on the plus. is the 'bottom bar' covering it?  I was using Editing Test Map_

PR #710, opened 2024-04-30
Apple Edits fix for geodatabase/table

mhdostal:
Added a check for `geodatabase.serviceInfo.canUseServiceGeodatabaseApplyEdits` to determine if we can use `applyEdits()` on the geodatabase or if we need to call it on the `table`.

Moved that code into a do-catch block and set an error if `applyEdits()` fails.

Updated the Tutorial as well.


dfeinzimer (approved):
Looks good!

PR #711, opened 2024-04-30
Add support for `AttachmentFormElement.isEditable`.

mhdostal:
This adds support for `AttachmentFormElement.isEditable`, which is a `@Streamed` property that can change based on an expression evaluation.  This PR:

- Removes `editControlsDisabled` modifier
- Uses `@State isEditable` property to control showing/hiding edit controls
- Adds a new modifier `onAttachmentIsEditableChange` event for watching for `isEditable` changes on `AttachmentFormElement`
- Adds a new `isShowingAttachmentFormElement` property

The `isEditable` bit can be tested by putting the following code at the top of the Group in `AttachmentFeatureElementView`:

```
        Toggle(isOn: $isEditable) {}
            .toggleStyle(.switch)
```

### DIFF
--- a/Sources/ArcGISToolkit/Common/AttachmentsFeatureElementView.swift
+++ b/Sources/ArcGISToolkit/Common/AttachmentsFeatureElementView.swift
@@ -25,6 +25,13 @@ struct AttachmentsFeatureElementView: View {
     
     @Environment(\.displayScale) var displayScale
     
+    /// A Boolean value indicating whether the input is editable.
+    @State private var isEditable = false
+    
+    /// A Boolean value indicating whether the feature Element
+    /// is an `AttachmentFormElement`.
+    private var isShowingAttachmentFormElement = false
+
     /// A Boolean value denoting if the view should be shown as regular width.
     var isRegularWidth: Bool {
         !isPortraitOrientation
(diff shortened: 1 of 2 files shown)


dfeinzimer:
Looks like this is targeting `main` currently.

mhdostal:
> Looks like this is targeting main currently.

fixed, thanks.

Sources/ArcGISToolkit/Common/AttachmentsFeatureElementView.swift
```diff
@@ -25,6 +25,13 @@ struct AttachmentsFeatureElementView: View {
     
     @Environment(\.displayScale) var displayScale
     
+    /// A Boolean value indicating whether the input is editable.
+    @State private var isEditable = false
+    
+    /// A Boolean value indicating whether the feature Element
+    /// is an `AttachmentFormElement`.
+    private var isShowingAttachmentFormElement = false
```
dfeinzimer, line 33: How would you feel about this as a computed property in an extension instead?

```swift
/// A Boolean value indicating whether the feature element
/// is an `AttachmentFormElement`.
var isShowingAttachmentFormElement: Bool {
    featureElement is AttachmentFormElement
}
```

dfeinzimer (approved):
Looks good just the one suggestion.

mhdostal:
Done. Please re-review. I fixed an opposite `if` check when reversing the attachment array. Booleans are hard. ;-)

dfeinzimer (approved):

PR #712, opened 2024-04-30
`FeatureFormView` - Improve attachment rename experience

dfeinzimer:
Related Apollo 408


dfeinzimer:
Ready for review - I've revised this to evaluate expressions after renaming and to not include the file extension in the rename dialogue.

Sources/ArcGISToolkit/Common/AttachmentRenameAlert.swift
```diff
@@ -0,0 +1,55 @@
...
+    /// The action to perform when the attachment is renamed.
+    let onRename: (AttachmentModel, String) async throws -> Void
+    
+    /// The attachment's file extension.
+    @State private var fileExtension: String?
+    
+    /// The attachment's new name.
+    @State private var newName = ""
+    
+    var body: some View {
```
mhdostal, line 31: I'm not sure what's happening, but If I change the name of an attachment, then attempt to change the name of another attachment, the $newName still shows the name of the previous attachment, not the current one.

Also, since there are no `async` methods in the `.task` below, should it be `.onAppear()`? Although if I set a breakpoint there, the second time through it never gets hit, leading me to believe it's not being recreated (or some off thing(.

dfeinzimer: Nice catch. It looks like `AttachmentRenameAlert` was being re-initialized but the task was only running once - same for `onAppear()`.

I found using `task(id: attachmentModel)` would work but then we also need to reset the state properties. I think that all makes the new view more trouble than its worth so I've removed it alltogether.

Sources/ArcGISToolkit/Common/AttachmentPreview.swift
```diff
@@ -77,8 +85,20 @@ struct AttachmentPreview: View {
             }
         }
         .alert("Rename attachment", isPresented: $renameDialogueIsShowing) {
-            if let onRename, let renamedAttachmentModel {
-                AttachmentRenameAlert(attachmentModel: renamedAttachmentModel, onRename: onRename)
+            TextField("New name", text: $newAttachmentName)
+            Button("Cancel", role: .cancel) { }
+            Button("Ok") {
```
mhdostal, line 90: Do we need to localize these? And I think "Ok" should be all caps: "OK". That string exists in the localizable.strings files.

mhdostal: It would be nice if there was a "clear" (`x`) button in the text filed, but that can come later, unless it's easy to do now.

mhdostal (changes requested):

Sources/ArcGISToolkit/Common/AttachmentPreview.swift
dfeinzimer: - Changed "Ok" to "OK".
- Yes we need to localize but I'd prefer to do that as a separate PR because I believe there're other non rename related strings we've since added that need to be done too and generating the strings file is a bit involved. I can add a bullet for this in the Apollo issue and do this next.

dfeinzimer: - Alerts seems to be very restrictive on how content is structured. A button placed into a `ZStack` with the `TextField` just ends up appearing inline with the cancel and OK buttons. Even in an `.overlay` on the `TextField` the same thing happens. If we want to pursue this we might need to do away with the alert.

mhdostal: Ok, no worries, we can leave it as-is.

mhdostal (approved):

PR #722, opened 2024-05-09
`FeatureFormView` - Improve text input performance for lengthy data

dfeinzimer:
Ref Apollo 649

### DIFF
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/TextInput.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/TextInput.swift
@@ -81,6 +81,7 @@ private extension TextInput {
             Group {
                 if element.isMultiline {
                     Text(text)
+                        .accessibilityIdentifier("\(element.label) Text Input Preview")
                         .fixedSize(horizontal: false, vertical: true)
                         .lineLimit(10)
                         .truncationMode(.tail)
@@ -98,10 +99,10 @@ private extension TextInput {
                         prompt: Text(element.hint).foregroundColor(.secondary),
                         axis: .horizontal
                     )
+                    .accessibilityIdentifier("\(element.label) Text Input")
                     .keyboardType(keyboardType)
                 }
             }
-            .accessibilityIdentifier("\(element.label) Text Input")
             .focused($isFocused)
             .frame(maxWidth: .infinity, alignment: .leading)
             .toolbar {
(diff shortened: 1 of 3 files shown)


Sources/ArcGISToolkit/Utility/RepresentedUITextView.swift
```diff
@@ -0,0 +1,49 @@
...
+    
+    class Coordinator: NSObject, UITextViewDelegate {
+        var text: Binding<String>
+        
+        init(text: Binding<String>) {
+            self.text = text
+        }
+        
+        func textViewDidChange(_ textView: UITextView) {
+            text.wrappedValue = textView.text
```
dfeinzimer, line 46: Performance improves even more if we only update the bound value via `textViewDidEndEditing` but then we lose the live footer character count and validation updates.

We could look at reworking how `InputFooter` works but I'm not sure if this'll be a common enough problem to do that at this point.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/TextInput.swift
```diff
@@ -23,9 +23,6 @@ struct TextInput: View {
     /// A Boolean value indicating whether or not the field is focused.
     @FocusState private var isFocused: Bool
     
-    /// The formatted version of the element's current value.
-    @State private var formattedValue = ""
-
```
dfeinzimer, line 28: Property was unused

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/TextInput.swift
```diff
@@ -82,24 +78,33 @@ private extension TextInput {
...
-            .disabled(element.isMultiline)
-            .sheet(isPresented: $fullScreenTextInputIsPresented) {
-                FullScreenTextInput(text: $text, element: element)
+            Group {
+                if element.isMultiline {
+                    Text(text)
+                        .accessibilityIdentifier("\(element.label) Text Input Preview")
+                        .fixedSize(horizontal: false, vertical: true)
+                        .lineLimit(10)
+                        .truncationMode(.tail)
```
dfeinzimer, line 87: This is a considerable change where we no longer show the entire text by default. Limiting the size makes a significant performance improvement when the form is drawn.

mhdostal: I think it will also improve the user experience, as there's no point in showing 28k worth of text; it just makes viewing the rest of the form more difficult.

mhdostal (approved):
Nice! The performance is much better.

PR #729, opened 2024-05-16
`FeatureFormView` - Attachment handling revisions

dfeinzimer:
Related Apollo 684


dfeinzimer:
This is ready for review once again

dfeinzimer:
Still ready for review. Updated `AttachmentFormElement.deleteAttachment(attachment)` to `AttachmentFormElement.delete(attachment)`.

mhdostal:
> Updated `AttachmentFormElement.deleteAttachment(attachment)` to `AttachmentFormElement.delete(attachment)`.

Is that what the new generated code outputs?

Sources/ArcGISToolkit/Common/AttachmentsFeatureElementView.swift
```diff
@@ -177,8 +177,7 @@ struct AttachmentsFeatureElementView: View {
     ///   - newAttachmentName: The new attachment name.
     @MainActor
     func onRename(attachmentModel: AttachmentModel, newAttachmentName: String) async throws -> Void {
```
mhdostal, line 179: This no longer needs to be `async throws`.
```suggestion
    func onRename(attachmentModel: AttachmentModel, newAttachmentName: String) -> Void {
```

Sources/ArcGISToolkit/Common/AttachmentsFeatureElementView.swift
```diff
@@ -192,10 +191,10 @@ struct AttachmentsFeatureElementView: View {
     func onDelete(attachmentModel: AttachmentModel) async throws -> Void {
```
mhdostal, line 191: This no longer needs to be `async throws`.
```suggestion
    func onDelete(attachmentModel: AttachmentModel) -> Void {
```

mhdostal:
Does the `AttachmentImportMenu.init(element: AttachmentsFormElement, onAdd: ((FeatureAttachment) async throws -> Void)? = nil)` method still need the `onAdd` argument to be `async throws`?

Same question for `onDelete` and `onRename` for the `AttachmentPreview` class?

mhdostal (changes requested):

dfeinzimer:
> Is that what the new generated code outputs?

It looks like taking the `attachments` property out of generation to make it not-cached actually caused that.. which I wouldn't have expected. Reverting for now.

mhdostal:
I think `AttachmentFormElement.delete(attachment)` is actually better. See the Swift API doc [here](https://www.swift.org/documentation/api-design-guidelines/#naming).

mhdostal (approved):
This looks good. We can figure out the naming during the API Audit.

PR #723, opened 2024-05-10
List preplanned map areas

des12437:
Related to `swift/5309`

PR #734, opened 2024-05-21
pr feedback: list-preplanned-map-areas

rolson:
This is my first round of feedback for [PR](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/723). I see several issues with the code in `List-Preplanned-Map-Areas`. 

This PR addresses several things, but the main issue I see is that status and download action are combined. That means that with the current code, I don't see a way that a download can fail, we can express that to the user, and they can try again. The status needs to be separate from the download button.

Other things it cleans up:
- don't need to specify label color parameter for foregroundStyle
- renames for clarity
- removes excess HStack
- separate out some subviews for cleaner body

image

There is more feedback that I have that revolves around ambiguous status states and errors. Also the model needs to contain the `AreaStatus` because the view should simply be a function of the model. The view is doing a bit too much there, IMO.

### DIFF
--- a/Sources/ArcGISToolkit/Components/OfflineMapAreasView/PreplannedListItemView.swift
+++ b/Sources/ArcGISToolkit/Components/OfflineMapAreasView/PreplannedListItemView.swift
@@ -91,7 +91,7 @@ public struct PreplannedListItemView: View {
             case .notLoaded:
                 areaStatus = .notLoaded
             case .failed:
-                areaStatus = .failed
+                areaStatus = packagingStatus == .processing ? .packaging : .failed
             }
         }
         .task {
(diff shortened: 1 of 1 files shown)


Sources/ArcGISToolkit/Components/OfflineMapAreasView/PreplannedListItemView.swift
```diff
@@ -38,111 +38,124 @@ public struct PreplannedListItemView: View {
...
+                if packagingStatus == .processing {
+                    // Disable downloading map area when still packaging.
+                    areaStatus = .packaging
+                } else {
+                    areaStatus = .notLoaded
                 }
+            case .notLoaded:
+                areaStatus = .notLoaded
+            case .failed:
+                areaStatus = .failed
```
des12437, line 94:
```suggestion
                areaStatus = packagingStatus == .processing ? .packaging : .failed
```
The map area will fail to load while packaging. The state needs to be set to `.processing` here so the correct status text shows when packaging.

rolson: this is fixed in the follow-up PR

PR #735, opened 2024-05-21
List Preplanned Map Areas: Simplify status, move more logic into model

rolson:


Sources/ArcGISToolkit/Components/OfflineMapAreasView/PreplannedListItemView.swift
```diff
@@ -17,142 +17,82 @@ import ArcGIS
...
-                }
-            }
+    }
+    
+    @ViewBuilder private var titleView: some View {
+        Text(model.preplannedMapArea.portalItem.title)
+            .font(.body)
+    }
+    
+
```
des12437, line 50:
```suggestion
```

Sources/ArcGISToolkit/Components/OfflineMapAreasView/PreplannedMapModel.swift
```diff
@@ -16,18 +16,86 @@ import SwiftUI
...
+    private func load() async {
+        do {
+            // Load preplanned map area to obtain packaging status.
+            status = .loading
+            try await preplannedMapArea.retryLoad()
+            // Note: Packaging status is `nil` for compatibility with
+            // legacy webmaps that have incomplete metadata.
+            // If the area loads, then you know for certain the status is complete.
+            updateStatus(for: preplannedMapArea.packagingStatus ?? .complete)
+        } catch is IllegalStateError {
```
des12437, line 44:
```suggestion
        } catch is MappingError {
```
When `retryLoad()` throws an error due to the map area packaging, the error is caught in the last catch block as a MappingError. This error should be cast as a MappingError instead.

rolson: which mapping error is it? we should be specific about it

des12437: This is the mapping error:
```
(lldb) po error
▿ MappingError
  ▿ packagingNotComplete : 1 element
    - details : "Failed to load preplanned map as the packaging status is not complete. Status: processing."
(lldb) po error.localizedDescription
"The operation couldn’t be completed. (ArcGIS.MappingError error 20.)"
```

des12437 (approved):

rolson:
cc @yo1995

PR #741, opened 2024-05-24
`FeatureFormView` - Revise example level error handling for attachments

dfeinzimer:
Related Apollo 679

### DIFF
--- a/Examples/Examples/FeatureFormExampleView.swift
+++ b/Examples/Examples/FeatureFormExampleView.swift
@@ -112,13 +112,6 @@ struct FeatureFormExampleView: View {
                         EmptyView()
                     }
                 }
-                .sheet(isPresented: model.applyEditsErrorsArePresented) {
-                    if case let .applyEditError(featureForm, errors) = model.state {
-                        ErrorTable(errors: errors) {
-                            model.state = .editing(featureForm)
-                        }
-                    }
-                }
                 .toolbar {
                     if model.formIsPresented.wrappedValue {
                         ToolbarItem(placement: .navigationBarLeading) {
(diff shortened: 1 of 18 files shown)


dfeinzimer:
@mhdostal Revised to simplify error display. This is ready for review.

Examples/Examples/FeatureFormExampleView.swift
```diff
@@ -147,70 +168,204 @@ private extension URL {
...
+        case idle
+        /// A feature form is in use.
+        case editing(FeatureForm)
+        /// The form is being checked for validation errors.
+        case validating(FeatureForm)
+        /// Edits are being committed to the local geodatabase.
+        case finishingEdits(FeatureForm)
+        /// Edits are being applied to the remote service.
+        case applyingEdits(FeatureForm)
+
```
mhdostal, line 183: Was there a reason for the empty line?
```suggestion
```

Examples/Examples/FeatureFormExampleView.swift
```diff
@@ -147,70 +168,204 @@ private extension URL {
...
         }
-        
-        try? await table.update(featureForm.feature)
-        
-        guard database.hasLocalEdits else {
-            print("No submittable changes found.")
+        if resultErrors.isEmpty {
+            state = .idle
+        } else {
+            state = .generalError(featureForm, Text("Changes were not applied."))
```
mhdostal, line 321: Maybe a comment here stating the client could display the errors to the user, if so desired.
```suggestion
            // Additionally, you could display the errors to the user using `resultErrors`.
            state = .generalError(featureForm, Text("Changes were not applied."))
```

mhdostal (commented):
Just a couple of comments.

mhdostal:
It is still pretty complicated, but the process of validating and applying changes is complicated as well, so I'm OK with it. We'll need to update the Tutorial as well.

Examples/Examples/FeatureFormExampleView.swift
dfeinzimer: Just to form a separation between the linear states and the two others that can pop up at any time.

dfeinzimer:
Tutorial updated, ready for re-review

mhdostal (approved):

PR #742, opened 2024-05-24
Offline: Add unit tests for `PreplannedMapModel`

yo1995:
See `swift/issues/5414`

This PR adds the unit tests for `PreplannedMapModel`.

Code coverage:

image

- Omitted unknown default
- Omitted the extended protocol strings for title, description, and thumbnail
- The download related states will be tested in a follow up PR

### DIFF
--- a/Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
+++ b/Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
@@ -1,0 +1,201 @@
+// Copyright 2024 Esri
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import XCTest
+import ArcGIS
+@testable import ArcGISToolkit
+
+private extension PreplannedMapAreaProtocol {
+    var packagingStatus: PreplannedMapArea.PackagingStatus? { nil }
+    var title: String { "Mock Preplanned Map Area" }
+    var description: String { "This is the description text" }
+    var thumbnail: ArcGIS.LoadableImage? { nil }
+}
+
+class PreplannedMapModelTests: XCTestCase {
+    private static var sleepNanoseconds: UInt64 { 100_000 }
+    
+    @MainActor
+    func testInitialStatus() {
+        class MockPreplannedMapArea: PreplannedMapAreaProtocol {
+            func retryLoad() async throws { }
+        }
+        
+        let mockArea = MockPreplannedMapArea()
+        let model = PreplannedMapModel(preplannedMapArea: mockArea)
+        guard case .notLoaded = model.status else {
+            XCTFail("PreplannedMapModel initial status is not \".notLoaded\".")
+            return
+        }
+        
+        XCTAssertFalse(model.canDownload)
+    }
+    
+    @MainActor
+    func testLoadingStatus() async throws {
+        class MockPreplannedMapArea: PreplannedMapAreaProtocol {
+            func retryLoad() async throws {
+                // Delay `retryLoad` method to slow down the model's private async
+                // `load` method, so the model status stays at "loading".
+                try await Task.sleep(nanoseconds: 2 * PreplannedMapModelTests.sleepNanoseconds)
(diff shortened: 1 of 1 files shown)


Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
```diff
@@ -0,0 +1,201 @@
...
+    
+    @MainActor
+    func testInitialStatus() {
+        class MockPreplannedMapArea: PreplannedMapAreaProtocol {
+            func retryLoad() async throws { }
+        }
+        
+        let mockArea = MockPreplannedMapArea()
+        let model = PreplannedMapModel(preplannedMapArea: mockArea)
+        guard case .notLoaded = model.status else {
```
yo1995, line 37: Because the `PreplannedMapModel.Status` is not `Equatable` and has associated values (error), I went for the detour to compare the enum cases. Please let me know if there is a better approach!

Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
```diff
@@ -0,0 +1,201 @@
...
+            func retryLoad() async throws {
+                // Delay `retryLoad` method to slow down the model's private async
+                // `load` method, so the model status stays at "loading".
+                try await Task.sleep(nanoseconds: 2 * PreplannedMapModelTests.sleepNanoseconds)
+            }
+        }
+        
+        let mockArea = MockPreplannedMapArea()
+        let model = PreplannedMapModel(preplannedMapArea: mockArea)
+        try await Task.sleep(nanoseconds: PreplannedMapModelTests.sleepNanoseconds)
```
yo1995, line 57: As discussed, I used `Task.sleep` to wait for the detached `Task` in the initializer. Please let me know if there is a better approach.

Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
```diff
@@ -0,0 +1,201 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import XCTest
+import ArcGIS
+@testable import ArcGISToolkit
+
+private extension PreplannedMapAreaProtocol {
```
yo1995, line 19: Added an extension to provide default values for properties unrelated to this test.

Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
```diff
@@ -0,0 +1,201 @@
...
+        XCTAssertTrue(model.canDownload)
+    }
+    
+    @MainActor
+    func testLoadFailureStatus() async throws {
+        class MockPreplannedMapArea: PreplannedMapAreaProtocol {
+            func retryLoad() async throws {
+                // Throws an error other than `MappingError.packagingNotComplete`
+                // to indicate the area fails to load in the first place.
+                throw MappingError.notLoaded(details: "")
```
yo1995, line 95: I didn't test the associated error's message.

Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
```diff
@@ -0,0 +1,212 @@
...
+                // Delay `retryLoad` method to slow down the model's private async
+                // `load` method, so the model status stays at "loading".
+                try await Task.sleep(nanoseconds: 2 * PreplannedMapModelTests.sleepNanoseconds)
+            }
+        }
+        
+        let mockArea = MockPreplannedMapArea()
+        let model = PreplannedMapModel(preplannedMapArea: mockArea)
+        try await Task.sleep(nanoseconds: PreplannedMapModelTests.sleepNanoseconds)
+        guard case .loading = model.status else {
```
des12437, line 69: This test case is flakey when ran repeatedly, where the the model.status equals .packaged and the assertion fails on line 70. The test succeeds when running the test suite or the individual test. Maybe the test can be kept as-is and then be updated if it fails in the daily tests.

yo1995: I'll add more time to the sleep.

I tried with the idea of moving the load method outside to the view instead of in a detached task. It removes most of the `Task.sleep`, but that doesn't solve the problem for the "loading" status test `func testLoadingStatus()`. In its case to test the `loading` status, I still have to use some sleep in `retryLoad` to delay its execution. 😕 

So for now I'll keep the detached task changes stashed until later discussion.

des12437 (approved):

rolson (commented):
Once the load changes are made in the other PR, and you remove the sleeps, I can re-review

rolson (approved):

des12437 (approved):

yo1995:
Thanks for the feedback.

PR #747, opened 2024-05-30
Offline: Remove detached task from model and call `load()` in the view

yo1995:
Instead of calling `Task.detached { await self.load() }` in the view model's initializer, call it in the task modifier so it is easier to unit test the method.

### DIFF
--- a/Sources/ArcGISToolkit/Components/OfflineMapAreasView/OfflineMapAreasView.swift
+++ b/Sources/ArcGISToolkit/Components/OfflineMapAreasView/OfflineMapAreasView.swift
@@ -73,9 +73,6 @@ public struct OfflineMapAreasView: View {
             if mapViewModel.hasPreplannedMapAreas {
                 List(models) { preplannedMapModel in
                     PreplannedListItemView(model: preplannedMapModel)
-                        .task {
-                            await preplannedMapModel.load()
-                        }
                 }
             } else {
                 emptyPreplannedMapAreasView
--- a/Sources/ArcGISToolkit/Components/OfflineMapAreasView/OfflineMapAreasView.swift
+++ b/Sources/ArcGISToolkit/Components/OfflineMapAreasView/OfflineMapAreasView.swift
@@ -73,6 +73,9 @@ public struct OfflineMapAreasView: View {
             if mapViewModel.hasPreplannedMapAreas {
                 List(models) { preplannedMapModel in
                     PreplannedListItemView(model: preplannedMapModel)
+                        .task {
+                            await preplannedMapModel.load()
+                        }
                 }
             } else {
                 emptyPreplannedMapAreasView
@@ -139,14 +142,6 @@ public extension OfflineMapAreasView {
             }
             if let models = try? preplannedMapModels!.get() {
                 hasPreplannedMapAreas = !models.isEmpty
-                // Kick off loading the map areas.
-                await withTaskGroup(of: Void.self) { group in
-                    for model in models {
-                        group.addTask {
-                            await model.load()
-                        }
-                    }
-                }
             }
         }
     }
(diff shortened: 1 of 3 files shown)


Sources/ArcGISToolkit/Components/OfflineMapAreasView/PreplannedListItemView.swift
```diff
@@ -100,8 +100,10 @@ public struct PreplannedListItemView: View {
 }
 
 #Preview {
-    PreplannedListItemView(
-        model: PreplannedMapModel(preplannedMapArea: MockPreplannedMapArea())
+    let model = PreplannedMapModel(preplannedMapArea: MockPreplannedMapArea())
+    Task.detached { await model.load() }
```
yo1995, line 104: This can also be written like https://stackoverflow.com/questions/77336072

des12437 (approved):

rolson (commented):
It seems like the view that needs it loaded should load it. In this case the `PreplannedListItemView`.

Sources/ArcGISToolkit/Components/OfflineMapAreasView/OfflineMapAreasView.swift
```diff
@@ -73,6 +73,9 @@ public struct OfflineMapAreasView: View {
             if mapViewModel.hasPreplannedMapAreas {
                 List(models) { preplannedMapModel in
                     PreplannedListItemView(model: preplannedMapModel)
+                        .task {
```
rolson, line 76: this should move to `PreplannedListItemView` itself

rolson (commented):
My feedback hasn't changed since last review. I think the view that uses it should load it, in this case `PreplannedListItemView`. Then the preview doesn't need the load.

Sources/ArcGISToolkit/Components/OfflineMapAreasView/PreplannedMapModel.swift
```diff
@@ -26,13 +26,10 @@ class PreplannedMapModel: ObservableObject, Identifiable {
...
     init(preplannedMapArea: PreplannedMapAreaProtocol) {
         self.preplannedMapArea = preplannedMapArea
-        
-        // Kick off a load of the map area.
-        Task.detached { await self.load() }
     }
     
     /// Loads the preplanned map area and updates the status.
-    private func load() async {
+    func load() async {
```
rolson, line 32: We could add a guard status != .loaded here

yo1995: Added conditional checks. Do you want me to write it in a switch-case (case .packaged, .downloaded: … default: …) instead?

Sources/ArcGISToolkit/Components/OfflineMapAreasView/PreplannedMapModel.swift
```diff
@@ -26,13 +26,12 @@ class PreplannedMapModel: ObservableObject, Identifiable {
...
+        if case .packaged = status { return }
+        if case .downloaded = status { return }
```
rolson, line 34: is there a reason to go with `if` instead of `guard` here? `guard` is tailor made for early exit of a function

yo1995: I don't know what is the syntax of using `guard` here. It cannot do something like `guard case .packaged != status else { return }` because guard-case or if-case behave somewhat like guard-let and if-let, which doesn't allow `!=`. Also, since the `Status` enum is not `Equatable`, we cannot simply use `==` to compare the enum cases.

Alternatively as I mentioned, we can use switch-case here such as`

```swift
switch status {
    case .packaged, .downloaded: return
    default: break
}
```

if you prefer that, I can make the change.

rolson: I would do something like this:

```swift

    func foo(status: Foo) {
        guard status.needsToBeLoaded else { return }
        /// ...
    }

enum Foo {
    case loaded
    case loading
    case packaged
    case failure(any Error)
    
    var needsToBeLoaded: Bool {
        switch self {
        case .loaded:
            false
        case .loading:
            false
        case .packaged:
            false
        case .failure(let error):
            true
        }
    }
}

```

rolson (approved):

des12437 (approved):

PR #755, opened 2024-06-17
Small tweaks for new, final, FF Attachments API.

mhdostal:
Build fixes for consuming the final FF Attachments API.

- `AnyAttachmentsFormInput` is now internal.
- The `FormAttachment.attachment` property is now non-nullable, which differs from `PopupAttachment.attachment`. However, the `FeatureForm` protocol's `attachment` property wasn't being used, so it was removed, which eliminated the discrepancy between `FormAttachment` and `PopupAttachment`.

### DIFF
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentImportMenu.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentImportMenu.swift
@@ -102,12 +102,9 @@ struct AttachmentImportMenu: View {
                 .progressViewStyle(.circular)
         }
         Menu {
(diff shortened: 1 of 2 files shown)


dfeinzimer (approved):
I was just working on a PR :)

mhdostal:
> I was just working on a PR :)

I hope you had essentially the same changes as me!!

dfeinzimer:
> > I was just working on a PR :)
> 
> I hope you had essentially the same changes as me!!

Yep, was just looking at how to handle the protocol issue.

mhdostal:
> Yep, was just looking at how to handle the protocol issue.

That one was bumming me out, as I didn't see a good way around it. But fortunately, it wasn't used, so could be removed.

PR #758, opened 2024-06-18
`FeatureFormView` - Consume `async throws` attachments property

dfeinzimer:


dfeinzimer:
Closing in favor of #765

mhdostal (approved):
Looks good.

Sources/ArcGISToolkit/Extensions/ArcGIS/AttachmentHelpers/AttachmentsFormElement.swift
```diff
@@ -27,8 +27,7 @@ extension AttachmentsFormElement : AttachmentsFeatureElement {
     /// This property will be empty if the element has not yet been evaluated.
     public var featureAttachments: [FeatureAttachment] {
         get async throws {
-            try await fetchAttachments()
-            return attachments.map { $0 as FeatureAttachment }
+            try await attachments.map { $0 as FeatureAttachment }
```
philium, line 30: This map is confusing. The cast in the transform closure is unconditional, so `attachments` itself could just be cast (`try await attachments as [FeatureAttachment]`). But since the cast is unconditional, the compiler knows that the cast will succeed, which makes me wonder if a cast is even needed.

dfeinzimer: Good catch, thanks!

PR #767, opened 2024-06-21
Remove unneeded cast

dfeinzimer:
https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/758/files#r1649264601


philium (approved):

philium:
> Remove unneeded cast

Note that this wasn't just an unneeded cast but an unneeded map, which is more significant. The map operation created an entirely new array, adding each element one by one. We are no longer doing any of that unnecessary work.

dfeinzimer:
> Note that this wasn't just an unneeded cast but an unneeded map, which is more significant. The map operation created an entirely new array, adding each element one by one.

Good point 👍

PR #770, opened 2024-06-24
fix retain cycle

rolson:


des12437 (approved):

rolson:
@yo1995 I'd like to get this merged even though this code will be changed [in a future PR](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/774). Because I'd like to have the bug fixed in case we have to revert the other PR for some reason. This doesn't fix all known issues with the code that we discovered, but it fixes the memory leak, which I think would be good to do.

PR #774, opened 2024-06-25
Add offline manager

rolson:
- [x] user notification should be fired at the app level, example app can do this so that app devs can just copy/paste the code
- [x] it's best if the developer does not need to interact with the job manager to use the component
- [x] it would be nice to have a WindowGroup/Scene modifier that is a one stop shop for the user to wire things up for the offline toolkit component(s)

Note: the user will still have to add the required job manager plist settings to the app, there is no way to workaround that requirement.

The approach taken in this PR fixes a few lifetime issues. One main one, consider this case:

- app starts up, user opens OfflineAreasView, starts downloading a job.
- app is backgrounded and killed at some point, doesn't matter how
- app restarts, either by user or by the OS, doesn't matter how
- any code dealing with a job completion will not be run unless the view OfflineAreasView is re-shown. 
- this might include removing the job from the job manager or showing a notification that the job was complete

This PR fixes the above situation by introducing an always-alive OfflineManager singleton, which is initialized through the `offlineManager(...)` Scene extension method. Now when the app is restarted and a job completes, you don't need to open the OfflineAreasView to run the completion code.


Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -264,7 +237,6 @@ protocol PreplannedMapAreaProtocol {
     var title: String { get }
     var description: String { get }
     var thumbnail: LoadableImage? { get }
-    var id: PortalItem.ID? { get }
```
rolson, line 267: This wasn't used anywhere from the protocol. Everywhere it was used there is an actual PreplannedMapArea instance.

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -172,8 +148,9 @@ class PreplannedMapModel: ObservableObject, Identifiable {
                 parameters: parameters,
                 downloadDirectory: mmpkDirectory
             )
-            JobManager.shared.jobs.append(job)
-            startAndObserveJob(job)
+            
+            OfflineManager.shared.start(job: job)
```
rolson, line 152: The models that kick off jobs now need to do that through the OfflineManager and not use the JobManager at all.

Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
```diff
@@ -0,0 +1,104 @@
...
+    /// - Parameter job: The job to start.
+    func start(job: any JobProtocol) {
+        jobManager.jobs.append(job)
+        observeJob(job)
+        job.start()
+    }
+    
+    /// Observes a job for completion.
+    private func observeJob(_ job: any JobProtocol) {
+        Task {
```
rolson, line 53: I think it's ok to have an unstructured task here. It's captures a singleton that remains alive throughout the life of the application. Making this somehow structured doesn't buy us anything that I can think of.

des12437 (approved):
Looks good. `OfflineMapAreasExampleApp.AppDelegate` could be removed if we decide that only background and not foreground notifications are needed.

rolson:
> Looks good. `OfflineMapAreasExampleApp.AppDelegate` could be removed if we decide that only background and not foreground notifications are needed.

I have another PR that I will open after this with some logging added and a bug fix, and I will remove the file in that PR.

Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
```diff
@@ -0,0 +1,104 @@
...
+    var jobCompletionAction: ((any JobProtocol) -> Void)?
+    
+    /// The job manager used by the offline manager.
+    let jobManager = JobManager.shared
+    
+    /// The jobs managed by this instance.
+    var jobs: [any JobProtocol] { jobManager.jobs }
+    
+    private init() {
+        // Observe each job's status
```
yo1995, line 34:
```suggestion
        // Observe each job's status.
```

Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
```diff
@@ -0,0 +1,104 @@
...
+        for job in jobManager.jobs {
+            observeJob(job)
+        }
+        
+        // Resume all paused jobs
```
yo1995, line 39:
```suggestion
        // Resume all paused jobs.
```

Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
```diff
@@ -0,0 +1,104 @@
...
+    var isComplete: Bool {
+        switch self {
+        case .notStarted, .started, .paused, .canceling:
+            false
+        case .succeeded, .failed:
+            true
+        }
+    }
+}
+
```
yo1995, line 77: This Boolean seems not to be used anymore?

```suggestion
```

Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
```diff
@@ -0,0 +1,104 @@
...
+    ) -> some SwiftUI.Scene {
+        // Set the background status check schedule.
+        OfflineManager.shared.jobManager.preferredBackgroundStatusCheckSchedule = preferredBackgroundStatusCheckSchedule
+        
+        // Set callback for job completion.
+        OfflineManager.shared.jobCompletionAction = jobCompletionAction
+        
+        // Support app-relaunch after background downloads.
+        return self.backgroundTask(.urlSession(ArcGISEnvironment.defaultBackgroundURLSessionIdentifier)) {
+            // Allow the `ArcGISURLSession` to handle it's background task events.
```
yo1995, line 96:
```suggestion
            // Allow the `ArcGISURLSession` to handle its background task events.
```

Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
yo1995: As long as the `try? await job.output` async method deterministically returns within finite time, `Task` block should be fine.

OfflineMapAreasExample/OfflineMapAreasExample/OfflineMapAreasExampleApp.swift
```diff
@@ -24,13 +24,36 @@ struct OfflineMapAreasExampleApp: App {
...
+        .offlineManager(preferredBackgroundStatusCheckSchedule: .regularInterval(interval: 30)) { job in
+            // Post a local notification that the job is finished.
+            Self.notifyJobCompleted(job: job)
         }
     }
 }
+
+extension OfflineMapAreasExampleApp {
+    /// Posts a local notification that the job completed with success or failure.
+    static func notifyJobCompleted(job: any JobProtocol) {
```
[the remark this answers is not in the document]
rolson: yes

yo1995 (approved):

PR #776, opened 2024-06-26
`FeatureFormView` - Use audited attachments API

dfeinzimer:
Apollo 650

### DIFF
--- a/Sources/ArcGISToolkit/Extensions/ArcGIS/AttachmentHelpers/FeatureAttachment.swift
+++ b/Sources/ArcGISToolkit/Extensions/ArcGIS/AttachmentHelpers/FeatureAttachment.swift
@@ -30,9 +30,6 @@ public enum FeatureAttachmentKind {
 }
 
 public protocol FeatureAttachment: Loadable {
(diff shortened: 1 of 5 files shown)


Sources/ArcGISToolkit/Extensions/ArcGIS/AttachmentHelpers/FeatureAttachment.swift
```diff
@@ -48,8 +45,8 @@ public protocol FeatureAttachment: Loadable {
     /// The name of the attachment.
     var name: String { get }
     
-    /// The size of the attachment in bytes.
-    var size: Int { get }
+    /// The size of the attachment.
+    var measuredSize: Measurement<UnitInformationStorage> { get }
```
dfeinzimer, line 49: This was cleaner than completely dropping `size` from the protocol.

Dropping it completely meant we'd need to add checking for `PopupAttachments`/`FormAttachments` in `AttachmentList` and `AttachmentPreview`.

mhdostal (approved):

PR #801, opened 2024-07-11
`AttachmentsFeatureElementView` - Validate microphone access

dfeinzimer:
Apollo 695

If microphone access is denied, videos will be recorded without audio, likely without the user knowing/remembering. This allows us to check the microphone access status when the user switches to video and show an alert if access is denied. 

iOS will only automatically prompt for microphone access the first time after the app is installed.

IMG_5DBE13DA049E-1

### DIFF
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentCameraController.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentCameraController.swift
@@ -25,6 +25,9 @@ struct AttachmentCameraController: UIViewControllerRepresentable {
     /// The image picker controller represented within the view.
     private let controller = UIImagePickerController()
     
+    /// The action to perform when the camera capture mode has changed.
+    var onCameraCaptureModeChanged: ((UIImagePickerController.CameraCaptureMode) -> Void)? = nil
+    
     /// Dismisses the picker controller.
     func endCapture() {
         controller.dismiss(animated: true)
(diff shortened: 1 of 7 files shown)


mhdostal (approved):

Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentCameraController.swift
```diff
@@ -75,3 +75,37 @@ final class CameraControllerCoordinator: NSObject, UIImagePickerControllerDelega
...
+    /// Observes changes to the camera capture mode.
+    var cameraCaptureModeObserver: (any NSObjectProtocol)?
+    
+    /// An action to perform when the camera capture mode changes.
+    var action: ((UIImagePickerController.CameraCaptureMode) -> Void)?
+    
+    override func viewDidAppear(_ animated: Bool) {
+        cameraCaptureModeObserver = NotificationCenter.default.addObserver(forName: NSNotification.Name("SourceFormatDidChange"), object: nil, queue: nil) { _ in
+            Task {
+                await self.action?(self.cameraCaptureMode)
```
philium, line 101: Always check for warnings. This introduced a warning because the `action` closure is not asynchronous.

dfeinzimer: Hmm, I see it now when using 15.4. I didn't see it because I was using 16.0 while working on this, and 16 doesn't show the error.

dfeinzimer: Promise I wouldn't have merged had I seen it 🙂

philium: Remember to always test on the minimum supported Xcode version. Otherwise there's the risk of introducing code that has build warnings/errors when building with a supported version of Xcode.

dfeinzimer: #804

PR #804, opened 2024-07-11
Resolve newly introduced warnings

dfeinzimer:
Re: https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/801#discussion_r1674378809

### DIFF
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentCameraController.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentCameraController.swift
@@ -98,7 +98,7 @@ class AttachmentUIImagePickerController: UIImagePickerController {
     override func viewDidAppear(_ animated: Bool) {
         cameraCaptureModeObserver = NotificationCenter.default.addObserver(forName: NSNotification.Name("SourceFormatDidChange"), object: nil, queue: nil) { _ in
             Task {
-                await self.action?(self.cameraCaptureMode)
+                self.action?(self.cameraCaptureMode)
             }
         }
     }
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentCameraController.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentCameraController.swift
@@ -98,7 +98,11 @@ class AttachmentUIImagePickerController: UIImagePickerController {
     override func viewDidAppear(_ animated: Bool) {
         cameraCaptureModeObserver = NotificationCenter.default.addObserver(forName: NSNotification.Name("SourceFormatDidChange"), object: nil, queue: nil) { _ in
             Task {
-                self.action?(self.cameraCaptureMode)
+                #if compiler(>=6.0)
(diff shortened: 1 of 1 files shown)


Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentCameraController.swift
```diff
@@ -98,7 +98,7 @@ class AttachmentUIImagePickerController: UIImagePickerController {
     override func viewDidAppear(_ animated: Bool) {
         cameraCaptureModeObserver = NotificationCenter.default.addObserver(forName: NSNotification.Name("SourceFormatDidChange"), object: nil, queue: nil) { _ in
             Task {
-                await self.action?(self.cameraCaptureMode)
+                self.action?(self.cameraCaptureMode)
```
dfeinzimer, line 101: Xcode 15.4
image

Xcode 16
image

I'm not sure I understand why 16.0 considers the expression async while 15.4 doesn't. 

Since we're not using Swift 6 yet, I assume the proper thing to do is remove `await` for now and revisit this when we move to Swift 6?

dfeinzimer:
I just noticed the following warning is back in 15.4 and 16.0 after [this commit](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/commit/c71edf74ab9d4bff65fea62f0a0a8346123b5f09#).

Screenshot 2024-07-11 at 10 33 48 AM

Any reason to not have `swiftlint:disable vertical_parameter_alignment_on_call` in there?

Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentCameraController.swift
philium: You could do that. But if you want to, you could account for both:

```swift
#if compiler(>=6.0)
// Use await.
#else
// Don't use await.
#endif
```

dfeinzimer:
> Are you using the latest version of SwiftLint? I'm using 0.55.1.

Ah, I was stuck on 0.53.0.

Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentCameraController.swift
```diff
@@ -98,7 +98,11 @@ class AttachmentUIImagePickerController: UIImagePickerController {
     override func viewDidAppear(_ animated: Bool) {
         cameraCaptureModeObserver = NotificationCenter.default.addObserver(forName: NSNotification.Name("SourceFormatDidChange"), object: nil, queue: nil) { _ in
             Task {
-                await self.action?(self.cameraCaptureMode)
+                #if compiler(>=6.0)
+                    await self.action?(self.cameraCaptureMode)
+                #else
+                    self.action?(self.cameraCaptureMode)
+                #endif
```
philium, line 105: Xcode prefers to left-align compiler directives.

```suggestion
#if compiler(>=6.0)
                await self.action?(self.cameraCaptureMode)
#else
                self.action?(self.cameraCaptureMode)
#endif
```

philium (approved):

PR #807, opened 2024-07-11
Fix build error in test

dfeinzimer:
Re: #801

### DIFF
--- a/Test Runner/Test Runner/TestViews/AttachmentCameraControllerTestView.swift
+++ b/Test Runner/Test Runner/TestViews/AttachmentCameraControllerTestView.swift
@@ -21,7 +21,7 @@ struct AttachmentCameraControllerTestView: View {
     @State private var captureMode: UIImagePickerController.CameraCaptureMode?
     
     var body: some View {
-        Text(captureMode?.description ?? "None")
+        Text(captureMode?.name ?? "None")
             .accessibilityIdentifier("Camera Capture Mode")
         AttachmentCameraController(importState: .constant(.none))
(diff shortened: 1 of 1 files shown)


Test Runner/Test Runner/TestViews/AttachmentCameraControllerTestView.swift
```diff
@@ -31,6 +31,7 @@ struct AttachmentCameraControllerTestView: View {
     
 }
 
+#if compiler(>=6.0)
```
philium, line 34: I was about to comment that you could extract the body of the property to only conditionalize what is needed:

```swift
#if compiler(>=6.0)
extension UIImagePickerController.CameraCaptureMode: @retroactive CustomStringConvertible {
    public var description: String { _description }
}
#else
extension UIImagePickerController.CameraCaptureMode: CustomStringConvertible {
    public var description: String { _description }
}
#endif

private extension UIImagePickerController.CameraCaptureMode {
    var _description: String {
        switch self {
        case .photo: "Photo"
        case .video: "Video"
        @unknown default: "N/A"
        }
    }
}
```

But then I started thinking about this more and realized that these are user-facing (used in a UI) strings. Even though this is test code, that isn't what `CustomStringConvertible.description` is designed for, nor should it be used for. A better solution would be to just remove the conformance to `CustomStringConvertible` (since it isn't needed at all) and rename the property to `title` or `name`, better reflecting what the returned value represents. (For an enum, the compiler-synthesized description is the case name itself, which is another reason why `CustomStringConvertible` doesn't fit for this.)

philium (approved):

PR #812, opened 2024-07-13
Enable targeted strict concurrency checking

philium:
Enables targeted strict concurrency in the package and associated targets as a step on the way to complete strict concurrency. If a target built with other warnings, those were addressed as well.


mhdostal (approved):
Awesome, thank you!

dfeinzimer (approved):

PR #814, opened 2024-07-16
Address remaining complete strict concurrency error

philium:
Addresses the following error:

Screenshot 2024-07-16 at 10 33 42 AM
image

Note: I do not have a device running iOS 16 to test with. If one of you does, I would appreciate it if you could verify that code pathway.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
@@ -150,7 +150,7 @@ class ARSwiftUIViewProxy: NSObject, ARSessionProviding {
     
     /// The AR session.
     @objc dynamic var session: ARSession {
-        arView.session
+        MainActor.runUnsafely { arView.session }
     }
 }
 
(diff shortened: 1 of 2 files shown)


dfeinzimer:
Unfortunately, I don't have a device running 16 anymore.

dfeinzimer (approved):
Seems good to me but we may want another reviewer with a device on 16 to test.

philium:
The good news is that very few users are likely to use the hand-rolled implementation. As long as they are running iOS 17 or the developer built the app with Xcode 15.3 or newer, they'll get the built-in implementation.

philium:
@saratchandrakarumuri Do you have a device running iOS 16?

saratchandrakarumuri:
I do have a device that we lent to other team... I can get it back this evening and give it a try

mhdostal:
> Note: I do not have a device running iOS 16 to test with. If one of you does, I would appreciate it if you could verify that code pathway.

I tested all three AR scenarios with an iPhone 8 w/iOS 16.7.4. No issues.

mhdostal (approved):

philium:
@mhdostal Thank you!

PR #815, opened 2024-07-16
Enable complete strict concurrency

philium:


Sources/ArcGISToolkit/Components/Search/SearchResult.swift
```diff
@@ -16,7 +16,7 @@ import UIKit.UIImage
 import ArcGIS
 
 /// Wraps a search result for display.
-public struct SearchResult: Sendable {
+public struct SearchResult: @unchecked Sendable {
```
philium, line 19: This is not ideal, but it can't really be helped because making `SearchSource` sendable would be _very_ complicated, if it is even possible. Ideally this (and `SearchSuggestion`) wouldn't have a reference to the search source, but that ship has sailed.

dfeinzimer (dismissed):

mhdostal (approved):

mhdostal (approved):

Issue #777, opened 2024-06-26
Xcode 16 warnings


dfeinzimer:
Triage and resolve 13 new warnings raised in Xcode 16

<img width="1012" alt="Screenshot 2024-06-25 at 6 28 58 PM" src="https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/assets/16397058/72d67d7a-e901-41a9-b464-8774947a6c61">

rolson:
We have warnings in Xcode 15.4 that need to be fixed as well:
<img width="400" alt="image" src="https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/assets/3718875/c8237fee-51ab-4ce2-8e78-ede6722aa3dd">
Some of these will go away by getting the latest ArcGIS, but others have been there for a while and need to be fixed.

PR #829, opened 2024-07-30
Resolve extended conformance warnings

dfeinzimer:
Reference #777 

If any of these conformances are ever added in the SDK we should still get a new warning about it.


philium (changes requested):
The `retroactive` attribute will be introduced in Swift 6. Instead, please use the fully qualified protocol name as described in [SE-0364](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0364-retroactive-conformance-warning.md).

philium (approved):

mhdostal (approved):

PR #832, opened 2024-08-01
Fix `DisclosureGroup` arrow padding for iOS

### LINKED ISSUE #528: Fix `DisclosureGroup` trailing padding on iOS
The disclosure groups in a Popup from the `Popup` example need more trailing padding since they look partially covered on iOS. Mac Catalyst does not have this issue since the disclosure group arrow icon is left rather than right aligned.

iPad Air:
Screenshot 2023-12-01 at 10 45 56 AM

iPhone 14 Pro Max iOS 16:

dfeinzimer:
Closes #528


des12437 (approved):
Looks good. Thank you for making this change.

PR #837, opened 2024-08-08
Fix SwiftUI preview strings generating entries in Localizable.strings

dfeinzimer:
Strings in SwiftUI previews generate entries in Localizable.strings for translation unless we use `Text(verbatim:)`.

I don't believe we need to translate preview strings, as they're essentially developer tooling and not for end-users. If we do, I can close this PR.

### DIFF
--- a/Sources/ArcGISToolkit/Utility/CredentialInputView.swift
+++ b/Sources/ArcGISToolkit/Utility/CredentialInputView.swift
@@ -47,7 +47,7 @@ extension View {
 
 struct CredentialInputSheetView_Previews: PreviewProvider {
     static var previews: some View {
-        Text(verbatim: "test")
+        Color.clear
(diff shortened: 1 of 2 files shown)


Sources/ArcGISToolkit/Utility/CredentialInputView.swift
```diff
@@ -47,27 +47,23 @@ extension View {
 
 struct CredentialInputSheetView_Previews: PreviewProvider {
     static var previews: some View {
-        Text(
-            "test",
-            bundle: .toolkitModule,
-            comment: "A label for development purposes only. This is not user visible."
-        )
```
dfeinzimer, line 54: Text is overkill for this preview - it is still completely functional without it.

mhdostal (approved):

PR #838, opened 2024-08-12
`FeatureFormExampleView` - `Model.textForState` -> `Model.State.label`

dfeinzimer:
As suggested by @rolson

> Consider making this logic an extension on the State type that has something like var label: String
> Then usage is just something like
Text(state.label)


rolson (approved):
Looks good. I feel bad for suggesting it if it modified so many tutorial files

dfeinzimer:
> I feel bad for suggesting it if it modified so many tutorial files

No worries, goes quickly

mhdostal (approved):

PR #839, opened 2024-08-14
`FeatureFormView` & `PopupView` -  Add ability to hide default headers

### LINKED ISSUE #835: Support Navigation Bar Title For Popup and FeatureForm
I have an app I am writing that goes from a readonly popup to an editable feature form. 

The UI uses a Sheet and wraps the PopupView and FeatureForm view inside a navigation stack. It would be great to be able to put the title in the navigation stack title, however, doing so would cause the title to display twice. Not doing so causes a gap to show where the navigation title is supposed to be.

<details><summary>Double Title</summary>
image
</details>

Done Criteria:
- a way to show the title for PopupView and FeatureFormView within a Navigation Bar title, without it showing in the views itself.
- For feature form, the title will need to somehow be available as Observable/Published, so that it can be updated from a `.navigationTitle(...)` view modifier, as the title can change dynamically based on the data being edited.

dfeinzimer:
Closes #835 

- Adds:
  - `PopupView.popupHeader(_:)`
    - Controls the visibility of the title and close button (visible by default).
  - `FeatureFormView.formHeader(_:)`
    - Controls the visibility of the title (visible by default).
- Deprecates:
  - `PopupView.showCloseButton(_:)`

### DIFF
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView+TitleVisibility.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView+TitleVisibility.swift
@@ -1,0 +1,34 @@
+// Copyright 2024 Esri
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import SwiftUI
+
+public extension FeatureFormView {
+    /// The visibility of the form title.
+    enum TitleVisibility: Sendable {
+        /// The form title is hidden.
+        case hidden
+        /// The form title is visible.
+        case visible
+    }
+    
+    /// Specifies the visibility of the form title.
+    /// - Parameter visibility: The preferred visibility of the form title.
+    /// - Since: 200.6
+    func formTitle(_ visibility: TitleVisibility) -> FeatureFormView {
+        var copy = self
+        copy.titleVisibility = visibility
+        return copy
+    }
+}
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView+TitleVisibility.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView+TitleVisibility.swift
@@ -16,6 +16,7 @@ import SwiftUI
 
 public extension FeatureFormView {
     /// The visibility of the form title.
+    /// - Since: 200.6
     enum TitleVisibility: Sendable {
         /// The form title is hidden.
         case hidden
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView+ValidationErrorVisibility.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView+ValidationErrorVisibility.swift
@@ -26,7 +26,7 @@ public extension FeatureFormView {
     
     /// Specifies the visibility of validation errors in the form.
     /// - Parameter visibility: The preferred visibility of validation errors in the form.
-    func validationErrors(_ visibility: ValidationErrorVisibility) -> FeatureFormView {
+    func validationErrors(_ visibility: ValidationErrorVisibility) -> Self {
         var copy = self
         copy.validationErrorVisibility = visibility
         return copy
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView+ValidationErrorVisibility.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView+ValidationErrorVisibility.swift
@@ -26,8 +26,10 @@ public extension FeatureFormView {
     
     /// Specifies the visibility of validation errors in the form.
     /// - Parameter visibility: The preferred visibility of validation errors in the form.
-    func validationErrors(_ visibility: ValidationErrorVisibility) -> some View {
-        environment(\.validationErrorVisibility, visibility)
+    func validationErrors(_ visibility: ValidationErrorVisibility) -> FeatureFormView {
+        var copy = self
+        copy.validationErrorVisibility = visibility
+        return copy
     }
 }
 
@@ -42,6 +44,6 @@ extension EnvironmentValues {
 
 /// The validation error visibility configuration of a form.
 @available(visionOS, unavailable)
-private struct FormViewValidationErrorVisibility: EnvironmentKey {
+struct FormViewValidationErrorVisibility: EnvironmentKey {
     static let defaultValue: FeatureFormView.ValidationErrorVisibility = .automatic
 }
--- a/Sources/ArcGISToolkit/Components/Popups/PopupView.swift
+++ b/Sources/ArcGISToolkit/Components/Popups/PopupView.swift
@@ -69,6 +69,13 @@ public struct PopupView: View {
     /// so that the the "close" button can close the view.
     var showCloseButton = false
     
+    /// A Boolean value indicating whether the deprecated `PopupView.showCloseButton(_:)`
+    /// modifier is applied.
+    ///
+    /// Once the modifier is removed this property can be removed and the presence or lack of a value for
+    /// `isPresented` should be the sole criteria to show or hide the close button.
+    var showCloseButtonDeprecatedModifierIsApplied = false
+    
     /// The visibility of the popup header.
     var headerVisibility: Visibility = .automatic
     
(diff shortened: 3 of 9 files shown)


Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView+ValidationErrorVisibility.swift
```diff
@@ -39,6 +41,6 @@ extension EnvironmentValues {
 }
 
 /// The validation error visibility configuration of a form.
-private struct FormViewValidationErrorVisibility: EnvironmentKey {
```
dfeinzimer, line 42: Changing this to internal allows `FeatureFormView` to access the default value.

Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView+ValidationErrorVisibility.swift
```diff
@@ -25,8 +25,10 @@ public extension FeatureFormView {
     
     /// Specifies the visibility of validation errors in the form.
     /// - Parameter visibility: The preferred visibility of validation errors in the form.
-    func validationErrors(_ visibility: ValidationErrorVisibility) -> some View {
```
dfeinzimer, line 28: This needed to change to `-> FeatureFormView` so that the two `FeatureFormView` modifiers can be placed in any order.

Sources/ArcGISToolkit/Components/Popups/PopupView.swift
```diff
@@ -203,6 +206,24 @@ extension PopupView {
...
+        /// The popup title is hidden.
+        case hidden
+        /// The popup title is visible.
+        case visible
+    }
+    
+    /// Specifies the visibility of the popup title.
+    /// - Parameter visibility: The visibility of the popup title.
+    /// - Since: 200.6
+    public func popupTitle(_ visibility: TitleVisibility) -> Self {
```
dfeinzimer, line 221: I'm not set on this modifier for one reason - the `showCloseButton` button modifier.

It seems disjoint that a developer can hide the title and simultaneously leave the close button visible.
```swift
PopupView(popup: popup!, isPresented: $showPopup)
    .popupTitle(.hidden)
    .showCloseButton(true)
```
Screenshot 2024-08-14 at 10 56 03 AM


Should we instead deprecate `showCloseButton` and add something like `.popupHeader(_:)` which would hide both?

mhdostal: The title and close button are not necessarily related, other than they have been occupying the same space. It may be that the client doesn't want the title in the popup/view but still requires the close button. So I'm OK having them both be distinct and put the onus on the client to figure out how they want it to display/behave.

Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView+TitleVisibility.swift
```diff
@@ -0,0 +1,35 @@
...
+
+public extension FeatureFormView {
+    /// The visibility of the form title.
+    /// - Since: 200.6
+    enum TitleVisibility: Sendable {
+        /// The form title is hidden.
+        case hidden
+        /// The form title is visible.
+        case visible
+    }
```
mhdostal, line 25: Why an enum instead of a `Bool`? `shouldShowTitle`, or something similar?

Sources/ArcGISToolkit/Components/Popups/PopupView.swift
rolson: I agree with David that it is a bit disjoint. I'm not sure we should've had a simple boolean for the close button. Some people might prefer a "Done" button. There is no way to customize that (well I guess the library is open source so people _could_ customize it). 

The `isPresented` works if using the toolkit's FloatingPanel. And I guess it works with `.sheet` as well, but it's actually unnecessary with `.sheet` because we could have used the DismissAction environment value:

```swift
@Environment(\.dismiss) private var dismiss
```

I think, for the most part, our components should not assume their container. 

Since title text is sometimes (often) on the "same line" as some buttons (done, cancel, menu), we might do better with a constructor like this that enables the PopupView to build a header for situations where they want to add a close button, but the title isn't displayed in a navigation toolbar:

```swift
struct PopupView<HeaderContent: View>: View {
    let popup: Popup
    let headerContent: ((String) -> HeaderContent)?

    ...
    
    /// Creates a PopupView...
    /// - Parameters:
    ///   - popup: The Popup to display
    ///   - headerContent: A closure that builds the header for a given title
    init(popup: Popup, @ViewBuilder headerContent: @escaping (String) -> HeaderContent) {
        self.popup = popup
        self.headerContent = headerContent
    }
```

This would allow somebody to add their own title, buttons, etc.

That would solve the close button thing. Then I guess there could be some modifier for `header(_ visibility:)`.

dfeinzimer: @mhdostal Any preference on the individual title and close button controls vs. Ryan's suggestion?

mhdostal: I'm OK with @rolson 's suggestion. That does solve the issue with the "close" button and allows clients to set it up how they want. It's a good point that our components should not assume their container. We could be better to trying different containers to verify that.

Sources/ArcGISToolkit/Components/Popups/PopupView.swift
```diff
@@ -67,7 +67,10 @@ public struct PopupView: View {
     /// A Boolean value specifying whether a "close" button should be shown or not. If the "close"
     /// button is shown, you should pass in the `isPresented` argument to the initializer,
     /// so that the the "close" button can close the view.
-    private var showCloseButton = false
+    var showCloseButton = false
```
dfeinzimer, line 70: `showCloseButton` is now internal so that it can be modified from `PopupView+ShowCloseButton.swift`.

Sources/ArcGISToolkit/Components/Popups/PopupView.swift
```diff
@@ -201,17 +206,3 @@ extension PopupView {
...
-    /// initializer, so that the the "close" button can close the view.
-    /// Defaults to `false`.
-    /// - Parameter newShowCloseButton: The new value.
-    /// - Returns: A new `PopupView`.
-    public func showCloseButton(_ newShowCloseButton: Bool) -> Self {
-        var copy = self
-        copy.showCloseButton = newShowCloseButton
-        return copy
-    }
-}
```
dfeinzimer, line 217: Moved to `PopupView+ShowCloseButton.swift` to be consistent with how other PopupView and FeatureFormView modifiers are written.

Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView+TitleVisibility.swift
dfeinzimer: I've updated this to use the native SwiftUI [`Visibility`](https://developer.apple.com/documentation/swiftui/visibility) type which is used in other similar modifiers.

Sources/ArcGISToolkit/Components/Popups/PopupView.swift
```diff
@@ -77,27 +80,29 @@ public struct PopupView: View {
...
     public var body: some View {
         VStack(alignment: .leading) {
-            HStack {
-                if !popup.title.isEmpty {
-                    Text(popup.title)
-                        .font(.title)
-                        .fontWeight(.bold)
-                }
-                Spacer()
-                if showCloseButton {
```
dfeinzimer, line 87: By removing `showCloseButton` from here now, users can avoid a deprecation warning and begin the new practice of controlling the close button's visibility via the binding in the initializer.

dfeinzimer:
@mhdostal and @rolson this is ready for review once again. The updated changes are:

- Adds:
  - `PopupView.popupHeader(_:)`
    - Controls the visibility of the title **and** close button (visible by default).
  - `FeatureFormView.formHeader(_:)`
    - Controls the visibility of the title (visible by default).
- Deprecates:
  - `PopupView.showCloseButton(_:)`

In regards to the [suggestion above](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/839#discussion_r1744586053), would it be better to avoid the complication of maintaining new initializers? With these changes devs can hide the header and easily include a custom one with a `VStack`:

```swift
.sheet(isPresented: $showPopup) { [popup] in
    VStack(alignment: .leading) {
        Text(popup!.title)
            .font(.title)
        PopupView(popup: popup!)
            .popupHeader(.hidden)
    }
    .padding()
}
```

dfeinzimer:
This is still ready for review. I made one more slight change to avoid breaking the behavior of `PopupView.showCloseButton(_:)`.

mhdostal (dismissed):
One question.

Sources/ArcGISToolkit/Components/Popups/PopupView.swift
```diff
@@ -77,27 +87,30 @@ public struct PopupView: View {
...
+                HStack {
+                    if !popup.title.isEmpty {
+                        Text(popup.title)
+                            .font(.title)
+                            .fontWeight(.bold)
                     }
-                    .labelStyle(.iconOnly)
+                    Spacer()
+                    if (showCloseButtonDeprecatedModifierIsApplied && showCloseButton)
+                        || (!showCloseButtonDeprecatedModifierIsApplied && isPresented != nil) {
```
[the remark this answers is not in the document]
dfeinzimer: Yes, assumed. One could technically do something odd like pass `.constant(false)` into the initializer parameter and still have the `PopupView` presented but that would be unintended use of the parameter.

rolson (approved):

mhdostal (approved):

rolson:
I utilized this in an app I'm working on and it's working well.

PR #840, opened 2024-08-14
Offline: Resolve concurrency warnings

des12437:
Resolves concurrency warnings after merging `v.next` into this branch.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
@@ -121,7 +121,19 @@ public struct PreplannedListItemView: View {
 }
 
 #Preview {
-    PreplannedListItemView(
+    struct MockPreplannedMapArea: PreplannedMapAreaProtocol {
+        var packagingStatus: PreplannedMapArea.PackagingStatus? = .complete
+        var title: String = "Mock Preplanned Map Area"
+        var description: String = "This is the description text"
+        var thumbnail: LoadableImage? = nil
+        
+        func retryLoad() async throws { }
+        func makeParameters(using offlineMapTask: OfflineMapTask) async throws -> DownloadPreplannedOfflineMapParameters {
+            DownloadPreplannedOfflineMapParameters()
+        }
+    }
+    
+    return PreplannedListItemView(
         model: PreplannedMapModel(
             offlineMapTask: OfflineMapTask(onlineMap: Map()),
             mapArea: MockPreplannedMapArea(),
@@ -130,16 +142,4 @@ public struct PreplannedListItemView: View {
         )
     )
     .padding()
-}
-
-private struct MockPreplannedMapArea: PreplannedMapAreaProtocol {
-    var packagingStatus: PreplannedMapArea.PackagingStatus? = .complete
-    var title: String = "Mock Preplanned Map Area"
-    var description: String = "This is the description text"
-    var thumbnail: LoadableImage? = nil
-    
-    func retryLoad() async throws { }
-    func makeParameters(using offlineMapTask: OfflineMapTask) async throws -> DownloadPreplannedOfflineMapParameters {
-        DownloadPreplannedOfflineMapParameters()
-    }
 }
--- a/Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
@@ -16,6 +16,7 @@ import SwiftUI
 import ArcGIS
 
 @MainActor
+@preconcurrency
 public struct PreplannedListItemView: View {
     /// The view model for the preplanned map.
     @ObservedObject var model: PreplannedMapModel
--- a/Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
@@ -15,6 +15,7 @@
 import SwiftUI
 import ArcGIS
 
+@MainActor
 public struct PreplannedListItemView: View {
     /// The view model for the preplanned map.
     @ObservedObject var model: PreplannedMapModel
--- a/Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
@@ -76,11 +76,11 @@ class PreplannedMapModel: ObservableObject, Identifiable {
             // Note: Packaging status is `nil` for compatibility with
             // legacy webmaps that have incomplete metadata.
             // If the area loads, then you know for certain the status is complete.
-            updateStatus(for: preplannedMapArea.packagingStatus ?? .complete)
+            status = preplannedMapArea.packagingStatus.map(Status.init) ?? .packaged
         } catch MappingError.packagingNotComplete {
             // Load will throw an `MappingError.packagingNotComplete` error if not complete,
             // this case is not a normal load failure.
-            updateStatus(for: preplannedMapArea.packagingStatus ?? .failed)
+            status = preplannedMapArea.packagingStatus.map(Status.init) ?? .packageFailure
         } catch {
             // Normal load failure.
             status = .loadFailure(error)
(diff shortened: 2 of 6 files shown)


Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
```diff
@@ -15,6 +15,7 @@
 import SwiftUI
 import ArcGIS
 
+@MainActor
 public struct PreplannedListItemView: View {
```
philium, line 19: This also needs the preconcurrency attribute since it is public. The MainActor annotation adds a new constraint to this view and since we don't know how everyone uses this type—they may have extended it with their own custom code, for example—we need the preconcurrency warning to prevent potentially breaking their source.

```suggestion
@MainActor
@preconcurrency
public struct PreplannedListItemView: View {
```

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -101,12 +101,14 @@ class PreplannedMapModel: ObservableObject, Identifiable {
...
-        case .complete:
-            status = .packaged
+            case .processing:
+                status = .packaging
+            case .failed:
+                status = .packageFailure
+            case .complete:
+                status = .packaged
+            @unknown default:
+                fatalError("Unknown case")
```
philium, line 111: The indentation is off here. Besides that, this is essentially type conversion (`PreplannedMapArea.PackagingStatus` -> `PreplannedMapModel.Status`), which is typically done in Swift via an initializer:

```swift
extension PreplannedMapModel.Status {
    init(packagingStatus: PreplannedMapArea.PackagingStatus) {
        self = switch packagingStatus {
        case .processing: .packaging
        case .failed: .packageFailure
        case .complete: .packaged
        @unknown default: fatalError("Unknown case")
        }
    }
}
```

Then, at the point of use, you can use it like so:

```swift
status = preplannedMapArea.packagingStatus.map(Status.init) ?? .packaged
```

That's what I suggest doing. A big reason that Swift prefers initializers for type conversion is that they can be used anywhere, whereas a method like `updateStatus(for:)` can only be used on this particular type.

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -229,13 +231,13 @@ extension PreplannedMapModel: Hashable {
         lhs === rhs
     }
     
-    public func hash(into hasher: inout Hasher) {
+    nonisolated public func hash(into hasher: inout Hasher) {
```
philium, line 234: This class is internal, so neither of these methods (`==` and `hash(into:)`) should be public. Also, while you're at it, I would suggest breaking this extension up into two extensions, one per protocol adoption:

```swift
extension PreplannedMapModel: Equatable {
    nonisolated static func == (lhs: PreplannedMapModel, rhs: PreplannedMapModel) -> Bool {
        return lhs === rhs
    }
}

extension PreplannedMapModel: Hashable {
    nonisolated func hash(into hasher: inout Hasher) {
        hasher.combine(ObjectIdentifier(self))
    }
}
```

There's also the issue that neither of those conformances are abiding by their expected semantics (they're using identity, but identity is not the same as equality), but that's really outside the scope of this work.

Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
```diff
@@ -32,7 +32,7 @@ private extension PreplannedMapAreaProtocol {
 class PreplannedMapModelTests: XCTestCase {
     @MainActor
     func testInit() {
-        class MockPreplannedMapArea: PreplannedMapAreaProtocol {
+        final class MockPreplannedMapArea: PreplannedMapAreaProtocol {
```
philium, line 35: An alternative (that I would suggest) would be to turn this into a struct:

```swift
struct MockPreplannedMapArea: PreplannedMapAreaProtocol {
    func retryLoad() async throws { }
}
```

You could even move that `retryLoad()` implementation into the `PreplannedMapAreaProtocol` extension so that you don't need it in three separate places. The assertion would then look like this:

```swift
XCTAssert(model.preplannedMapArea is MockPreplannedMapArea)
```

Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
```diff
@@ -87,7 +87,7 @@ class PreplannedMapModelTests: XCTestCase {
     
     @MainActor
     func testPackagingStatus() async throws {
-        class MockPreplannedMapArea: PreplannedMapAreaProtocol {
+        final class MockPreplannedMapArea: PreplannedMapAreaProtocol, @unchecked Sendable {
```
philium, line 90: While this works, it would be better for the sendability to be checked. You could do that by importing `os` and then using `OSAllocatedUnfairLock`:

```swift
final class MockPreplannedMapArea: PreplannedMapAreaProtocol {
    let packagingStatus = OSAllocatedUnfairLock<PreplannedMapArea.PackagingStatus?>(initialState: nil)
    
    func retryLoad() async throws {
        packagingStatus.withLock { $0 = .processing }
    }
}
```

Similar with the other mocks that need to set the packaging status.

Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
```diff
@@ -136,7 +136,7 @@ class PreplannedMapModelTests: XCTestCase {
     
     @MainActor
     func testPackagingNotCompletePackageFailureStatus() async throws {
-        class MockPreplannedMapArea: PreplannedMapAreaProtocol {
+        final class MockPreplannedMapArea: PreplannedMapAreaProtocol, @unchecked Sendable {
             var packagingStatus: PreplannedMapArea.PackagingStatus? = nil
```
philium, line 140: Here, there's a simpler solution. Just make `packagingStatus` computed:

```suggestion
        final class MockPreplannedMapArea: PreplannedMapAreaProtocol {
            var packagingStatus: PreplannedMapArea.PackagingStatus? { nil }
```

Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
des12437: I tried to apply this feedback and the status initialization fails on this line, so the test assertion fails:
```
status = preplannedMapArea.packagingStatus.map(Status.init) ?? .packaged
```
```
model.status.assertExpectedValue(.packaging) // assertion fails because status equals `.packaged`
```
With the current implementation the status is set with the initializer successfully so the test succeeds. I did not apply this feedback for now. Does the assertion need to be changed?

philium: I'm a bit confused, but I'll re-review the PR and comment further based on what I see.

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
philium, line 1: I noticed that `MockPreplannedMapArea` is only used by the preview, but it is outside of the preview's body. This means that it won't get stripped out of optimized/release builds like the preview itself is, even though it should be. FYI: You can move it into the body of the preview:

```swift
#Preview {
    struct MockPreplannedMapArea: PreplannedMapAreaProtocol {
        var packagingStatus: PreplannedMapArea.PackagingStatus? = .complete
        var title: String = "Mock Preplanned Map Area"
        var description: String = "This is the description text"
        var thumbnail: LoadableImage? = nil
        
        func retryLoad() async throws { }
        func makeParameters(using offlineMapTask: OfflineMapTask) async throws -> DownloadPreplannedOfflineMapParameters {
            DownloadPreplannedOfflineMapParameters()
        }
    }
    
    return PreplannedListItemView(
        model: PreplannedMapModel(
            offlineMapTask: OfflineMapTask(onlineMap: Map()),
            mapArea: MockPreplannedMapArea(),
            portalItemID: .init("preview")!,
            preplannedMapAreaID: .init("preview")!
        )
    )
    .padding()
}
```

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -224,18 +211,31 @@ extension PreplannedMapModel {
     }
 }
 
-extension PreplannedMapModel: Hashable {
-    nonisolated public static func == (lhs: PreplannedMapModel, rhs: PreplannedMapModel) -> Bool {
-        lhs === rhs
+extension PreplannedMapModel.Status {
```
philium, line 214: Should this be private?

Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
philium: Ah! I see. I'm sorry, I should have actually run the test before making my suggestion. The problem is that `packagingStatus` no longer satisfied the protocol requirement. Here is how to fix that:

```swift
final class MockPreplannedMapArea: PreplannedMapAreaProtocol {
    var packagingStatus: PreplannedMapArea.PackagingStatus? {
        _packagingStatus.withLock { $0 }
    }
    
    private let _packagingStatus = OSAllocatedUnfairLock<PreplannedMapArea.PackagingStatus?>(initialState: nil)
    
    func retryLoad() async throws {
        _packagingStatus.withLock { $0 = .processing }
    }
}
```

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
des12437: Yes.

philium (approved):

philium:
Thank you for making those changes!

PR #841, opened 2024-08-20
Offline: Open a preplanned map area

des12437:
Closes `swift/5358`

Adds support to open a downloaded preplanned map area with the "open" button which dismisses the view and returns a `Map` via the `onMapSelectionChanged` view modifier. If the preplanned map area's mobile map package fails to load when opened then the view's status is set to `mmpkLoadFailure`.

### DIFF
--- a/OfflineMapAreasExample/OfflineMapAreasExample/OfflineMapAreasExampleView.swift
+++ b/OfflineMapAreasExample/OfflineMapAreasExample/OfflineMapAreasExampleView.swift
@@ -19,13 +19,16 @@ import SwiftUI
 @MainActor
 struct OfflineMapAreasExampleView: View {
     /// The map of the Naperville water network.
-    @State private var map = Map(item: PortalItem.naperville())
+    @State private var onlineMap = Map(item: PortalItem.naperville())
+    
+    /// The selected map.
+    @State private var map: Map?
     
     /// A Boolean value indicating whether the offline map ares view should be presented.
     @State private var isShowingOfflineMapAreasView = false
     
     var body: some View {
-        MapView(map: map)
+        MapView(map: map ?? onlineMap)
             .toolbar {
                 ToolbarItem(placement: .bottomBar) {
                     Button("Offline Maps") {
@@ -34,7 +37,10 @@ struct OfflineMapAreasExampleView: View {
                 }
             }
             .sheet(isPresented: $isShowingOfflineMapAreasView) {
-                OfflineMapAreasView(map: map)
+                OfflineMapAreasView(map: onlineMap)
+                    .onMapSelectionChanged { newMap in
+                        map = newMap
+                    }
             }
     }
 }
--- a/OfflineMapAreasExample/OfflineMapAreasExample/OfflineMapAreasExampleView.swift
+++ b/OfflineMapAreasExample/OfflineMapAreasExample/OfflineMapAreasExampleView.swift
@@ -38,8 +38,8 @@ struct OfflineMapAreasExampleView: View {
             }
             .sheet(isPresented: $isShowingOfflineMapAreasView) {
                 OfflineMapAreasView(
-                    onlineMap: onlineMap,
-                    selectedMap: $selectedMap
+                    online: onlineMap,
+                    selection: $selectedMap
                 )
             }
     }
--- a/OfflineMapAreasExample/OfflineMapAreasExample/OfflineMapAreasExampleView.swift
+++ b/OfflineMapAreasExample/OfflineMapAreasExample/OfflineMapAreasExampleView.swift
@@ -22,13 +22,13 @@ struct OfflineMapAreasExampleView: View {
     @State private var onlineMap = Map(item: PortalItem.naperville())
     
     /// The selected map.
-    @State private var map: Map?
+    @State private var selectedMap: Map?
     
     /// A Boolean value indicating whether the offline map ares view should be presented.
     @State private var isShowingOfflineMapAreasView = false
     
     var body: some View {
-        MapView(map: map ?? onlineMap)
+        MapView(map: selectedMap ?? onlineMap)
             .toolbar {
                 ToolbarItem(placement: .bottomBar) {
                     Button("Offline Maps") {
@@ -39,7 +39,7 @@ struct OfflineMapAreasExampleView: View {
             .sheet(isPresented: $isShowingOfflineMapAreasView) {
                 OfflineMapAreasView(map: onlineMap)
                     .onMapSelectionChanged { newMap in
-                        map = newMap
+                        selectedMap = newMap
                     }
             }
     }
--- a/OfflineMapAreasExample/OfflineMapAreasExample/OfflineMapAreasExampleView.swift
+++ b/OfflineMapAreasExample/OfflineMapAreasExample/OfflineMapAreasExampleView.swift
@@ -37,10 +37,7 @@ struct OfflineMapAreasExampleView: View {
                 }
             }
             .sheet(isPresented: $isShowingOfflineMapAreasView) {
-                OfflineMapAreasView(
-                    online: onlineMap,
-                    selection: $selectedMap
-                )
+                OfflineMapAreasView(online: onlineMap, selection: $selectedMap)
             }
     }
 }
--- a/OfflineMapAreasExample/OfflineMapAreasExample/OfflineMapAreasExampleView.swift
+++ b/OfflineMapAreasExample/OfflineMapAreasExample/OfflineMapAreasExampleView.swift
@@ -37,10 +37,10 @@ struct OfflineMapAreasExampleView: View {
                 }
             }
             .sheet(isPresented: $isShowingOfflineMapAreasView) {
-                OfflineMapAreasView(map: onlineMap)
-                    .onMapSelectionChanged { newMap in
-                        selectedMap = newMap
-                    }
+                OfflineMapAreasView(
+                    onlineMap: onlineMap,
+                    selectedMap: $selectedMap
+                )
             }
     }
 }
--- a/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
@@ -29,18 +29,18 @@ public struct OfflineMapAreasView: View {
     @State private var isReloadingPreplannedMapAreas = false
     
     /// The currently selected map.
-    private var selectedMap: Binding<Map?>
+    @Binding private var selectedMap: Map?
     
-    /// Creates an `OfflineMapAreasView` with a given web map.
+    /// Creates a view with a given web map.
     /// - Parameters:
-    ///   - onlineMap: The web map.
-    ///   - selectedMap: A binding to the currently selected map.
+    ///   - online: The web map to be taken offline.
+    ///   - selection: A binding to the currently selected map.
     public init(
-        onlineMap: Map,
-        selectedMap: Binding<Map?>
+        online: Map,
+        selection: Binding<Map?>
     ) {
-        _mapViewModel = StateObject(wrappedValue: MapViewModel(map: onlineMap))
-        self.selectedMap = selectedMap
+        _mapViewModel = StateObject(wrappedValue: MapViewModel(map: online))
+        _selectedMap = selection
     }
     
     public var body: some View {
@@ -91,9 +91,8 @@ public struct OfflineMapAreasView: View {
                 List(models) { preplannedMapModel in
                     PreplannedListItemView(
                         model: preplannedMapModel
-                    )
-                    .onMapSelectionChanged { newMap in
-                        selectedMap.wrappedValue = newMap
+                    ) { newMap in
+                        selectedMap = newMap
                         dismiss()
                     }
                 }
@@ -133,13 +132,13 @@ public struct OfflineMapAreasView: View {
         
         var body: some View {
             OfflineMapAreasView(
-                onlineMap: Map(
+                online: Map(
                     item: PortalItem(
                         portal: .arcGISOnline(connection: .anonymous),
                         id: PortalItem.ID("acc027394bc84c2fb04d1ed317aac674")!
                     )
                 ),
-                selectedMap: $map
+                selection: $map
             )
         }
     }
--- a/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
@@ -28,13 +28,19 @@ public struct OfflineMapAreasView: View {
     /// A Boolean value indicating whether the preplanned map areas are being reloaded.
     @State private var isReloadingPreplannedMapAreas = false
     
-    /// The closure to perform when the map selection changes.
-    private var onMapSelectionChanged: ((Map) -> Void)?
+    /// The currently selected map.
+    private var selectedMap: Binding<Map?>
     
     /// Creates an `OfflineMapAreasView` with a given web map.
-    /// - Parameter map: The web map.
-    public init(map: Map) {
-        _mapViewModel = StateObject(wrappedValue: MapViewModel(map: map))
+    /// - Parameters:
+    ///   - onlineMap: The web map.
+    ///   - selectedMap: A binding to the currently selected map.
+    public init(
+        onlineMap: Map,
+        selectedMap: Binding<Map?>
+    ) {
+        _mapViewModel = StateObject(wrappedValue: MapViewModel(map: onlineMap))
+        self.selectedMap = selectedMap
     }
     
     public var body: some View {
@@ -87,7 +93,7 @@ public struct OfflineMapAreasView: View {
                         model: preplannedMapModel
                     )
                     .onMapSelectionChanged { newMap in
-                        onMapSelectionChanged?(newMap)
+                        selectedMap.wrappedValue = newMap
                         dismiss()
                     }
                 }
@@ -118,24 +124,24 @@ public struct OfflineMapAreasView: View {
         }
         .frame(maxWidth: .infinity)
     }
-    
-    /// Sets the closure to call when the map selection changes.
-    public func onMapSelectionChanged(
-        perform action: @escaping (_ newMap: Map) -> Void
-    ) -> Self {
-        var copy = self
-        copy.onMapSelectionChanged = action
-        return copy
-    }
 }
 
 #Preview {
-    OfflineMapAreasView(
-        map: Map(
-            item: PortalItem(
-                portal: .arcGISOnline(connection: .anonymous),
-                id: PortalItem.ID("acc027394bc84c2fb04d1ed317aac674")!
+    @MainActor
+    struct OfflineMapAreasViewPreview: View {
+        @State private var map: Map?
+        
+        var body: some View {
+            OfflineMapAreasView(
+                onlineMap: Map(
+                    item: PortalItem(
+                        portal: .arcGISOnline(connection: .anonymous),
+                        id: PortalItem.ID("acc027394bc84c2fb04d1ed317aac674")!
+                    )
+                ),
+                selectedMap: $map
             )
-        )
-    )
+        }
+    }
+    return OfflineMapAreasViewPreview()
 }
--- a/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
@@ -35,10 +35,7 @@ public struct OfflineMapAreasView: View {
     /// - Parameters:
     ///   - online: The web map to be taken offline.
     ///   - selection: A binding to the currently selected map.
-    public init(
-        online: Map,
-        selection: Binding<Map?>
-    ) {
+    public init(online: Map, selection: Binding<Map?>) {
         _mapViewModel = StateObject(wrappedValue: MapViewModel(map: online))
         _selectedMap = selection
     }
@@ -89,9 +86,7 @@ public struct OfflineMapAreasView: View {
         case .success(let models):
             if !models.isEmpty {
                 List(models) { preplannedMapModel in
-                    PreplannedListItemView(
-                        model: preplannedMapModel
-                    ) { newMap in
+                    PreplannedListItemView(model: preplannedMapModel) { newMap in
                         selectedMap = newMap
                         dismiss()
                     }
--- a/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
@@ -29,7 +29,7 @@ public struct OfflineMapAreasView: View {
     @State private var isReloadingPreplannedMapAreas = false
     
     /// The closure to perform when the map selection changes.
-    public var onMapSelectionChanged: ((Map) -> Void)?
+    private var onMapSelectionChanged: ((Map) -> Void)?
     
     /// Creates an `OfflineMapAreasView` with a given web map.
     /// - Parameter map: The web map.
--- a/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
@@ -28,6 +28,9 @@ public struct OfflineMapAreasView: View {
     /// A Boolean value indicating whether the preplanned map areas are being reloaded.
     @State private var isReloadingPreplannedMapAreas = false
     
+    /// The closure to perform when the map selection changes.
+    public var onMapSelectionChanged: ((Map) -> Void)?
+    
     /// Creates an `OfflineMapAreasView` with a given web map.
     /// - Parameter map: The web map.
     public init(map: Map) {
@@ -83,6 +86,10 @@ public struct OfflineMapAreasView: View {
                     PreplannedListItemView(
                         model: preplannedMapModel
                     )
+                    .onMapSelectionChanged { newMap in
+                        onMapSelectionChanged?(newMap)
+                        dismiss()
+                    }
                 }
             } else {
                 emptyPreplannedMapAreasView
@@ -110,6 +117,15 @@ public struct OfflineMapAreasView: View {
                 .foregroundStyle(.secondary)
         }
         .frame(maxWidth: .infinity)
+    }
+    
+    /// Sets the closure to call when the map selection changes.
+    public func onMapSelectionChanged(
+        perform action: @escaping (_ newMap: Map) -> Void
+    ) -> Self {
+        var copy = self
+        copy.onMapSelectionChanged = action
+        return copy
     }
 }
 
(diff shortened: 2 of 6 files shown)


Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
```diff
@@ -55,8 +58,21 @@ public struct PreplannedListItemView: View {
...
+                        onMapSelectionChanged?(map)
+                    }
+                }
+            } label: {
+                Text("Open")
+                    .font(.system(.headline, weight: .bold))
+                    .foregroundColor(.accentColor)
+                    .frame(width: 76, height: 32)
+                    .background(Color(UIColor.tertiarySystemFill))
+                    .cornerRadius(16)
```
yo1995, line 73: `cornerRadius` will be [deprecated](https://developer.apple.com/documentation/deviceactivity/deviceactivityreport/cornerradius(_:antialiased:)) in iOS 18. Let's use 
```suggestion
                    .clipShape(.rect(cornerRadius: 16))
```

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
```diff
@@ -55,8 +58,21 @@ public struct PreplannedListItemView: View {
...
+                    if let map = await model.loadMobileMapPackage() {
+                        onMapSelectionChanged?(map)
+                    }
+                }
+            } label: {
+                Text("Open")
+                    .font(.system(.headline, weight: .bold))
+                    .foregroundColor(.accentColor)
+                    .frame(width: 76, height: 32)
+                    .background(Color(UIColor.tertiarySystemFill))
```
yo1995, line 72: It can be `.background(Color(.tertiarySystemFill))`, but if we use the `backgroundStyle` API, then it can be 
```suggestion
                    .backgroundStyle(.tertiary)
```

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
```diff
@@ -55,8 +58,21 @@ public struct PreplannedListItemView: View {
...
+            Button {
+                Task {
+                    if let map = await model.loadMobileMapPackage() {
+                        onMapSelectionChanged?(map)
+                    }
+                }
+            } label: {
+                Text("Open")
+                    .font(.system(.headline, weight: .bold))
+                    .foregroundColor(.accentColor)
```
yo1995, line 70: `foregroundColor` is [deprecated](https://developer.apple.com/documentation/swiftui/view/foregroundcolor(_:)). Can you log an GitHub issue to clean them up later? Thanks.
```suggestion
                    .foregroundStyle(Color.accentColor)
```

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
```diff
@@ -21,6 +21,9 @@ public struct PreplannedListItemView: View {
     /// The view model for the preplanned map.
     @ObservedObject var model: PreplannedMapModel
     
+    /// The closure to perform when the map selection changes.
+    public var onMapSelectionChanged: ((Map) -> Void)?
```
yo1995, line 25:
```suggestion
    var onMapSelectionChanged: ((Map) -> Void)?
```

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
des12437: Logged an [issue](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/842).

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
philium, line 1: Should `PreplannedListItemView` be internal? It seems like it should be since it cannot be instantiated by the user (it requires an instance of `PreplannedMapModel`, which is internal).

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
des12437: `.backgroundStyle(.tertiary)` does not appear to be equivalent to `.background(Color(.tertiarySystemFill))` since the background color does not show, so I used `.background(Color(.tertiarySystemFill))`.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -111,6 +118,15 @@ public struct OfflineMapAreasView: View {
         }
         .frame(maxWidth: .infinity)
     }
+    
+    /// Sets the closure to call when the map selection changes.
+    public func onMapSelectionChanged(
+        perform action: @escaping (_ newMap: Map) -> Void
+    ) -> Self {
```
philium, line 125: Passing a selection back to a parent view via a modifier like this is highly unusual. Look through SwiftUI's interface (command click on SwiftUI wherever it is imported) and search for "selection". It is (almost?) always passed via a binding taken as an initialization parameter. Is there a strong reason not to do that with this view?

Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
```diff
@@ -284,6 +284,81 @@ class PreplannedMapModelTests: XCTestCase {
...
+            offlineMapTask: task,
+            mapArea: area,
+            portalItemID: portalItem.id!,
+            preplannedMapAreaID: areaID,
+            // User notifications in unit tests are not supported, must pass false here
+            // or the test process will crash.
+            showsUserNotificationOnCompletion: false
+        )
+        
+        var statuses = [PreplannedMapModel.Status]()
```
philium, line 318: Either `Array<PreplannedMapModel.Status>` (like you did with `Set` on the next line) or:

```suggestion
        var statuses: [PreplannedMapModel.Status] = []
```

Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
```diff
@@ -284,6 +284,81 @@ class PreplannedMapModelTests: XCTestCase {
...
+        var subscriptions = Set<AnyCancellable>()
+        model.$status
+            .receive(on: DispatchQueue.main)
+            .sink { value in
+                statuses.append(value)
+            }
```
philium, line 324: Not a big deal, but you could simplify this.

```suggestion
            .sink { statuses.append($0) }
```

Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
```diff
@@ -284,6 +284,81 @@ class PreplannedMapModelTests: XCTestCase {
...
+            .loading,
+            .packaged,
+            .downloading,
+            .downloading,
+            .downloaded
+        ]
+        
+        for (status, expected) in zip(statuses, expected) {
+            status.assertExpectedValue(expected)
+        }
```
philium, line 360: `PreplannedMapModel.Status` not being equatable makes it _very_ hard to work with. That's one reason why we didn't do something similar with `LoadStatus`. If at all possible, I strongly suggest moving the associated values out of the enum, allowing it to be equatable and greatly simplifying any code that uses it.

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
des12437: Yes, this view should be internal.

Tests/ArcGISToolkitTests/PreplannedMapModelStatusTests.swift
```diff
@@ -23,6 +23,7 @@ class PreplannedMapModelStatusTests: XCTestCase {
         XCTAssertFalse(PreplannedMapModel.Status.packaged.needsToBeLoaded)
         XCTAssertFalse(PreplannedMapModel.Status.downloading.needsToBeLoaded)
         XCTAssertFalse(PreplannedMapModel.Status.downloaded.needsToBeLoaded)
+        XCTAssertFalse(PreplannedMapModel.Status.mmpkLoadFailure(NSError()).needsToBeLoaded)
```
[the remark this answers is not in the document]
des12437: Done.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
des12437: No, I removed the view modifier and added a binding parameter to the initializer. Thanks for the suggestion.

Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
des12437: I added equatable conformance to `PreplannedMapModel.Status` and kept the associated values on the enum because it simplifies testing. The helper methods for `PreplannedMapModel.Status` in `PreplannedMapModelTests` have been removed.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
philium: You're welcome.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -28,10 +28,19 @@ public struct OfflineMapAreasView: View {
...
     /// Creates an `OfflineMapAreasView` with a given web map.
-    /// - Parameter map: The web map.
-    public init(map: Map) {
-        _mapViewModel = StateObject(wrappedValue: MapViewModel(map: map))
+    /// - Parameters:
+    ///   - onlineMap: The web map.
+    ///   - selectedMap: A binding to the currently selected map.
+    public init(
+        onlineMap: Map,
+        selectedMap: Binding<Map?>
```
philium, line 40: Remember not to repeat type information in parameter names. These should be `online` and `selection` (following SwiftUI's convention). Some with the corresponding properties.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -28,10 +28,19 @@ public struct OfflineMapAreasView: View {
...
+    ) {
+        _mapViewModel = StateObject(wrappedValue: MapViewModel(map: onlineMap))
```
philium, line 42: From the [Initialize state objects using external data](https://developer.apple.com/documentation/swiftui/stateobject#Initialize-state-objects-using-external-data) section of the doc:

> Use caution when doing this. SwiftUI only initializes a state object the first time you call its initializer in a given view. This ensures that the object provides stable storage even as the view’s inputs change. However, it might result in unexpected behavior or unwanted side effects if you explicitly initialize the state object.

If you haven't already or you haven't done so recently, please read through that section of the doc and consider whether any other changes need to be made.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -28,10 +28,19 @@ public struct OfflineMapAreasView: View {
     /// A Boolean value indicating whether the preplanned map areas are being reloaded.
     @State private var isReloadingPreplannedMapAreas = false
     
+    /// The currently selected map.
+    private var selectedMap: Binding<Map?>
```
philium, line 32: FYI: While this isn't wrong, you could use the the property wrapper here (`@Binding private var selectedMap: Map?`) by slightly modifying how the initial value is set in the initializer: `self._selectedMap = selectedMap`. Then, `.wrappedValue` wouldn't be needed later on.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -28,10 +28,19 @@ public struct OfflineMapAreasView: View {
...
+    
     /// Creates an `OfflineMapAreasView` with a given web map.
```
philium, line 34: There really isn't a need to repeat the type in initializer descriptions.

```suggestion
    /// Creates a view with a given web map.
```

By the way, I think the doc for this initializer could be improved. Isn't the selected map a web map too? What's the difference? `online` in the name of the first parameter helps, but the doc should bring that out, explaining what "online" means. For example, "The map to be taken offline."

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
```diff
@@ -17,10 +17,13 @@ import ArcGIS
 
 @MainActor
 @preconcurrency
-public struct PreplannedListItemView: View {
```
philium, line 20: Please remove the `public` modifier from all the other places in this view where it isn't needed.

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
```diff
@@ -118,6 +134,15 @@ public struct PreplannedListItemView: View {
...
     }
+    
+    /// Sets the closure to call when the map selection changes.
+    public func onMapSelectionChanged(
+        perform action: @escaping (_ newMap: Map) -> Void
+    ) -> Self {
+        var copy = self
+        copy.onMapSelectionChanged = action
+        return copy
+    }
```
philium, line 145: Using a modifier for this doesn't really make a whole lot of sense for a value that is internal and always needed by the view. Instead, just pass the selection changed handler to the view via an initializer parameter.

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -211,6 +226,26 @@ extension PreplannedMapModel {
     }
 }
 
+extension PreplannedMapModel.Status: Equatable {
```
philium, line 229: This is only needed for testing, correct? If so, please move this into the testing target. Not only does it prevent it from being compiled with and included in production code, but it also reduces the likelihood that the API will be abused (since it is a bit of a lie—it doesn't take associated values into account). That might be fine for testing, but not for production code.

Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
```diff
@@ -270,7 +270,9 @@ class PreplannedMapModelTests: XCTestCase {
         ]
         
         for (status, expected) in zip(statuses, expected) {
-            status.assertExpectedValue(expected)
+            if status != expected {
+                XCTFail("Status '\(status)' does not match expected status of '\(expected)'")
+            }
         }
```
philium, line 276: This (and what comes before) is no longer needed. After the sleep, you just need one assertion now:

```swift
XCTAssertEqual(
    statuses
    [.notLoaded, ...]
)
```

Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
```diff
@@ -282,39 +284,82 @@ class PreplannedMapModelTests: XCTestCase {
...
+            .downloading,
+            .downloading,
+            .downloaded
+        ]
+        
+        for (status, expected) in zip(statuses, expected) {
+            if status != expected {
+                XCTFail("Status '\(status)' does not match expected status of '\(expected)'")
+            }
+        }
```
philium, line 362:
```suggestion
        XCTAssertEqual(
            statuses,
            [.notLoaded, .loading, .packaged, .downloading, .downloading, .downloaded]
        )
```

philium (commented):
Remember, when writing code, start by limiting the exposure as much as possible. If it can be private, keep it private. If not, see if it can be just file private. If not, see it if can be just internal, etc. It is easy to increase the exposure of something later on down the road. It is much harder to reduce the exposure of something.

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
des12437: Yes the equatable conformance is only needed for testing. This conformance can be moved into the testing target but the conformance must be public:
Screenshot 2024-08-22 at 1 05 47 PM

Should this public conformance be added to the testing target? If not, the other options are to add back the `isMatch()` and `assertExpectedValue()` methods for the tests or to move the associated values out of the enum.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
des12437: The selected map is not a web map, its a map from a preplanned map area's mobile map package.

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
philium: Yes, just add the `public` access level to make the compiler happy. It shouldn't be necessary and may not be with never versions of Swift, but it doesn't hurt anything.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
philium: I see. Thanks.

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
des12437: Done.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
des12437: I'm not sure the `online` map is expected to change after the `OfflineMapAreasView` is declared. The view model uses the map to access it's portal item so that changes on arcgis online will show in the view. Should changing the `online` map parameter be supported?

OfflineMapAreasExample/OfflineMapAreasExample/OfflineMapAreasExampleView.swift
```diff
@@ -34,7 +37,10 @@ struct OfflineMapAreasExampleView: View {
                 }
             }
             .sheet(isPresented: $isShowingOfflineMapAreasView) {
-                OfflineMapAreasView(map: map)
+                OfflineMapAreasView(
+                    online: onlineMap,
+                    selection: $selectedMap
+                )
```
philium, line 43: Combined, the line would be shorter than 80 columns, if you want to.

```suggestion
                OfflineMapAreasView(online: onlineMap, selection: $selectedMap)
```

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -82,7 +91,10 @@ public struct OfflineMapAreasView: View {
                 List(models) { preplannedMapModel in
                     PreplannedListItemView(
                         model: preplannedMapModel
-                    )
+                    ) { newMap in
+                        selectedMap = newMap
+                        dismiss()
+                    }
```
philium, line 97: This could be simplified too:

```swift
PreplannedListItemView(model: preplannedMapModel) { newMap in
    selectedMap = newMap
    dismiss()
}
```

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
```diff
@@ -17,11 +17,14 @@ import ArcGIS
...
 @MainActor
 @preconcurrency
-public struct PreplannedListItemView: View {
+struct PreplannedListItemView: View {
     /// The view model for the preplanned map.
     @ObservedObject var model: PreplannedMapModel
     
-    public var body: some View {
+    /// The closure to perform when the map selection changes.
+    var onMapSelectionChanged: ((Map) -> Void)?
```
philium, line 25: This can be non-optional and immutable, correct?

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
```diff
@@ -55,8 +58,21 @@ public struct PreplannedListItemView: View {
...
-            Image(systemName: "checkmark.circle")
-                .foregroundStyle(.secondary)
+            Button {
+                Task {
+                    if let map = await model.loadMobileMapPackage() {
+                        onMapSelectionChanged?(map)
+                    }
+                }
+            } label: {
+                Text("Open")
```
philium, line 68: Should this be localized now or will that happen later, before the target branch is merged into v.next?

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
philium: Since this view is presented modally, most likely the online map won't change while it is presented.

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
des12437: There is a separate issue to localize the strings for this component which will be completed before merging into v.next.

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
des12437: Yes.

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
```diff
@@ -17,11 +17,14 @@ import ArcGIS
...
 @MainActor
 @preconcurrency
-public struct PreplannedListItemView: View {
+struct PreplannedListItemView: View {
     /// The view model for the preplanned map.
     @ObservedObject var model: PreplannedMapModel
     
-    public var body: some View {
+    /// The closure to perform when the map selection changes.
+    let onMapSelectionChanged: ((Map) -> Void)
```
philium, line 25: Without the question mark, the outer parentheses are no longer needed.

```suggestion
    let onMapSelectionChanged: (Map) -> Void
```

philium (approved):
Looks good. Thanks!

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
```diff
@@ -55,8 +58,21 @@ public struct PreplannedListItemView: View {
...
+                }
+            } label: {
+                Text("Open")
+                    .font(.system(.headline, weight: .bold))
+                    .foregroundStyle(Color.accentColor)
+                    .frame(width: 76, height: 32)
+                    .background(Color(.tertiarySystemFill))
+                    .clipShape(.rect(cornerRadius: 16))
+            }
+            .buttonStyle(.plain)
```
rolson, line 75: A few things in here that I'd like to see corrected:
- headline font is probably too large
- We should really try to refrain from using explicit frame sizes. Since we are using dynamic fonts, using hard-coded sizes will not scale properly.
- If you want a capsule button, there is a way to do that with the system. Don't need to set the bg color and clip shape manually.

This should do the trick:

```swift
                Button {
                     ....
                } label: {
                    Text("Open")
                        .fontWeight(.bold)
                }
                .buttonStyle(BorderedButtonStyle())
                .buttonBorderShape(.capsule)
```

yo1995: I agree, fixed frame can be changed. The reason a clipped rectangle instead of a capsule is used, is that it better mimics the App Store "Open" button. The capsule button looks too wide when used here.

philium: FYI: For the button style modifier, use `bordered`: `.buttonStyle(.bordered)`.

des12437: Here is a screenshot of the updated UI with these suggestions applied:

philium: I confess that I forgot to actually run the code to see what the UI looks like. That refresh button (I'm assuming that's what it is) is non-standard. Did you try implementing [pull to refresh](https://www.hackingwithswift.com/quick-start/swiftui/how-to-enable-pull-to-refresh)?

yo1995: Should pull-to-refresh apply to a modally presented sheet? I feel the pull down gesture will interfere with the dismiss action for this view. 🤔

philium: Pull to refresh is perfectly valid for a sheet, as long as pull to dismiss is disabled. I know that UIKit had a way of doing that. I assume that SwiftUI just does the right thing, but I don't know for certain.

rolson:
```swift
        .interactiveDismissDisabled()
```

philium: The Calendars view in the Calendar app is a sheet with pull to refresh, for example.

philium: @rolson for the win!

des12437: I will log an issue and implement pull to refresh in another PR.

des12437:
This PR is ready for re-review.

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
yo1995: Should it be `var onMapSelectionChanged: ((Map) -> Void)? = nil` with a default value, so a user doesn't have to provide a closure? I imagine that would be a rare case though, as one of the goals of this component is to allow a user to get the selected map.

Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
```diff
@@ -282,40 +268,87 @@ class PreplannedMapModelTests: XCTestCase {
...
+        var statuses: [PreplannedMapModel.Status] = []
+        var subscriptions = Set<AnyCancellable>()
+        model.$status
+            .receive(on: DispatchQueue.main)
+            .sink { statuses.append($0) }
+            .store(in: &subscriptions)
+        
+        await model.load()
+        
+        // Start downloading
```
yo1995, line 313:
```suggestion
        // Start downloading.
```

Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
```diff
@@ -282,40 +268,87 @@ class PreplannedMapModelTests: XCTestCase {
...
+            offlineMapTask: task,
+            mapArea: area,
+            portalItemID: portalItem.id!,
+            preplannedMapAreaID: areaID,
+            // User notifications in unit tests are not supported, must pass false here
+            // or the test process will crash.
+            showsUserNotificationOnCompletion: false
+        )
+        
+        var statuses: [PreplannedMapModel.Status] = []
```
yo1995, line 304: This resembles the `AsyncAccumulator` pattern in our SDK test suite. Maybe later we can change it to that. No need to change here.

yo1995 (approved):

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
philium: No, because this is an internal property and a closure is always provided.

yo1995: That's right, I misread it 🤕

yo1995 (approved):

des12437:
@philium can you please re-review this PR when you have a chance? Thank you!

philium (approved):

des12437:
Thank you both!

PR #843, opened 2024-08-21
Offline: Remove downloaded area and view details

yo1995:
## Description

This PR adds 2 features

- Remove downloaded files for a preplanned map area
    - By swiping the list row to the left, a delete button is revealed. Once tapped, the directory of the preplanned map area is removed from local disk
    - If an area is currently open but is removed, the `selectedMap` binding will be set to `nil` so the online web map is used instead
- View the metadata of a preplanned map area
    - Tapping on the list row will reveal the metadata view, similar to App Store listing
    - Currently it get all the metadata from the downloaded files. In the next on-demand workflow, some metadata will be serialized when the area is downloaded


Sources/ArcGISToolkit/Components/Offline/MetadataDetailView.swift
yo1995, line 1: |Apple Map|Our view|
|:-:|:-:|
|apple map|our view|

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -43,6 +52,11 @@ class PreplannedMapModel: ObservableObject, Identifiable {
     /// A Boolean value indicating if a user notification should be shown when a job completes.
     let showsUserNotificationOnCompletion: Bool
     
+    /// The file size of the preplanned map area.
+    var directorySize: Int? {
+        FileManager.default.sizeOfDirectory(at: mmpkDirectory)
```
yo1995, line 57: I didn't subtract the metadata.json and thumbnail.png size as they take space on disk as well.

Sources/ArcGISToolkit/Components/Offline/MetadataView.swift
```diff
@@ -0,0 +1,128 @@
...
+                        Text(model.preplannedMapArea.title)
+                            .font(.subheadline)
+                    }
+                    Divider()
+                    VStack(alignment: .leading, spacing: 0) {
+                        Text("Description")
+                            .font(.caption)
+                            .foregroundStyle(.secondary)
+                        Text(model.preplannedMapArea.description)
+                            .font(.subheadline)
```
des12437, line 53: An `isEmpty` check should be added for the description:

```suggestion
                        if !model.preplannedMapArea.description.isEmpty {
                            Text(model.preplannedMapArea.description)
                                .font(.subheadline)
                        } else {
                            Text("This area has no description.")
                                .font(.subheadline)
                                .foregroundStyle(.tertiary)
                        }
```

Sources/ArcGISToolkit/Components/Offline/MetadataView.swift
```diff
@@ -0,0 +1,128 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ArcGIS
+import SwiftUI
+
+struct MetadataView: View {
```
des12437, line 18:
```suggestion
struct PreplannedMetadataView: View {
```
This view should be specified for the preplanned map model. It may not be re-usable for the on-demand model since that version of the metadata view will support editing the title, description, and map area.

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
```diff
@@ -55,6 +74,20 @@ struct PreplannedListItemView: View {
...
     
+    @ViewBuilder private var deleteButton: some View {
+        if model.status.allowsRemoval {
+            Button {
+                Task {
+                    await model.removeDownloadedPreplannedMapArea()
+                    onMapDeletion()
+                }
+            } label: {
+                Label("Remove Area", systemImage: "trash")
```
des12437, line 85:
```suggestion
                Text("Delete")
```
`Delete` should be used here to match the "Delete Map Area" button in the metadata view. Is the icon necessary?

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -162,6 +170,14 @@ class PreplannedMapModel: ObservableObject, Identifiable {
         }
     }
     
+    /// Removes the downloaded preplanned map area from disk and reset status.
+    func removeDownloadedPreplannedMapArea() async {
```
des12437, line 174: A test should be added for this method in `PreplannedMapModelTests` to verify that the status.

yo1995:
Forgot to add test, will mark as ready for review once it is added.

Sources/ArcGISToolkit/Components/Offline/MetadataDetailView.swift
```diff
@@ -0,0 +1,128 @@
...
+                    }
+                }
+            }
+            Section {
+                HStack {
+                    ZStack {
+                        Image(systemName: "circle.fill")
+                            .foregroundStyle(.secondary)
+                            .opacity(0.1)
+                            .font(.title)
```
rolson: Instead of giving a specific size (title font), I'd probably just try giving the image a background (that will size it to the image) and clip it to a circle.

However, actually, we should try to use image `trash.circle.fill` or `trash.circle` instead. I would heavily lean towards standard images/components.

Sources/ArcGISToolkit/Components/Offline/MetadataDetailView.swift
```diff
@@ -0,0 +1,128 @@
...
+                    }
+                    Button {
+                        dismiss()
+                        Task {
+                            await model.removeDownloadedPreplannedMapArea()
+                        }
+                    } label: {
+                        Text("Delete Map Area")
+                            .font(.subheadline)
+                            .foregroundColor(.red)
```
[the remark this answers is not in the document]
rolson: What does the difference in appearance look like between destructive and setting the color manually? I would lean towards using built-in paradigms where possible instead of hard-coding colors/sizes/styles/etc.

Sources/ArcGISToolkit/Components/Offline/MetadataDetailView.swift
rolson: Looks pretty good. In comparison to Apple Maps, fonts look a bit small. Also "thumbnail" text not needed. And we can center the image.

Sources/ArcGISToolkit/Components/Offline/MetadataDetailView.swift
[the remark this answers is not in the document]
yo1995: You are right - destructive looks fine. I'll revert back to destructive.

Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
```diff
@@ -169,10 +169,14 @@ class PreplannedMapModelTests: XCTestCase {
         let areas = try await task.preplannedMapAreas
         let area = try XCTUnwrap(areas.first)
         let areaID = try XCTUnwrap(area.portalItem.id)
-        let mmpkDirectory = FileManager.default.preplannedDirectory(
```
yo1995, line 172: I refactored to make the `FileManager.preplannedDirectory` a private method:

- If the component is self contained, then a developer shouldn't touch the file structure that the component maintains
- It should be like a black box that all the offline maps are managed internally, and the developer can only access the maps via the APIs we provide them

Sources/ArcGISToolkit/Components/Offline/MetadataDetailView.swift
yo1995: 👍

yo1995:
Ready for review.

Sources/ArcGISToolkit/Components/Offline/PreplannedMetadataView.swift
```diff
@@ -0,0 +1,125 @@
...
+    public var body: some View {
+        Form {
+            Section {
+                if let image = model.preplannedMapArea.thumbnail {
+                    HStack {
+                        Spacer()
+                        LoadableImageView(loadableImage: image)
+                            .clipShape(.rect(cornerRadius: 10))
+                            .padding(.vertical, 10)
+                            .shadow(radius: 10)
```
des12437, line 34:
```suggestion
                            .shadow(radius: 5)
```
The shadow radius can be reduced:


Is there a way to make the image larger without compromising image quality?

yo1995: I have changed the max width and height to 1.5x of the thumbnail's native resolution. Let me know what you think.

des12437: Looks good :+1:

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -89,6 +89,8 @@ public struct OfflineMapAreasView: View {
                     PreplannedListItemView(model: preplannedMapModel) { newMap in
                         selectedMap = newMap
                         dismiss()
+                    } onMapDeletion: {
```
des12437, line 92:
```suggestion
                    } onDelete: {
```
SwiftUI uses the view modifier name `onDelete` for several of their views.

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -162,6 +170,13 @@ class PreplannedMapModel: ObservableObject, Identifiable {
         }
     }
     
+    /// Removes the downloaded preplanned map area from disk and reset status.
```
des12437, line 173:
```suggestion
    /// Removes the downloaded preplanned map area from disk and resets the status.
```

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
```diff
@@ -37,6 +43,19 @@ struct PreplannedListItemView: View {
                 statusView
             }
         }
+        .swipeActions {
+            deleteButton
+        }
```
des12437, line 48:
```suggestion
        }
        .contentShape(.rect)
```
This defines the content shape for hit testing. Otherwise the whitespace of the view is not tappable:

Sources/ArcGISToolkit/Components/Offline/MetadataDetailView.swift
```diff
@@ -0,0 +1,128 @@
...
+                    Divider()
+                    VStack(alignment: .leading, spacing: 0) {
+                        Text("Description")
+                            .font(.caption)
+                            .foregroundStyle(.secondary)
+                        Text(model.preplannedMapArea.description)
+                            .font(.subheadline)
+                    }
+                    Divider()
+                    if let size = model.directorySize {
```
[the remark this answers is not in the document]
des12437: Yes `.zero` could make sense as a default value, or if the value is `nil` a message could be shown  such as "File size unavailable" similar to the message "This area has no description" for the description.

Sources/ArcGISToolkit/Components/Offline/PreplannedMetadataView.swift
```diff
@@ -0,0 +1,125 @@
...
+    public var body: some View {
+        Form {
+            Section {
+                if let image = model.preplannedMapArea.thumbnail {
+                    HStack {
+                        Spacer()
+                        LoadableImageView(loadableImage: image)
+                            .clipShape(.rect(cornerRadius: 10))
+                            .padding(.vertical, 10)
+                            .shadow(radius: 5)
```
des12437, line 34:
```suggestion
```
Shadow radius                              | No shadow radius
  

The shadow may not be necessary. I think the design without the shadow radius looks better.

Sources/ArcGISToolkit/Components/Offline/PreplannedMetadataView.swift
```diff
@@ -0,0 +1,125 @@
...
+
+struct PreplannedMetadataView: View {
+    /// The view model for the preplanned map.
+    @ObservedObject var model: PreplannedMapModel
+    
+    /// The action to dismiss the view.
+    @Environment(\.dismiss) private var dismiss: DismissAction
+    
+    public var body: some View {
+        Form {
```
des12437, line 26: The top padding here between the title and form looks larger than other SwiftUI sheets. Can this padding be reduced or is it SwiftUI's automatic layout?

Sources/ArcGISToolkit/Components/Offline/PreplannedMetadataView.swift
```diff
@@ -0,0 +1,125 @@
...
+import SwiftUI
+
+struct PreplannedMetadataView: View {
+    /// The view model for the preplanned map.
+    @ObservedObject var model: PreplannedMapModel
+    
+    /// The action to dismiss the view.
+    @Environment(\.dismiss) private var dismiss: DismissAction
+    
+    public var body: some View {
```
philium, line 25:
```suggestion
    var body: some View {
```

philium (commented):
Here are a few things I saw right off the bat. This isn't my full review; I'll wait until other changes are done for that.

Sources/ArcGISToolkit/Components/Offline/PreplannedMetadataView.swift
yo1995: [Here](https://stackoverflow.com/questions/57851878/how-to-eliminate-the-space-above-section-in-swiftui-form) are some hacks to eliminate the empty space, though I don't like any of them 🤔 It seems the space is reserved for the section header and even if the header isn't specified, the space is still occupied. Maybe SwiftUI will change the appearance in the future.

des12437: Okay thanks. It's probably better to not add a workaround for this issue then.

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -313,6 +343,21 @@ extension FileManager {
         offlineMapAreasDirectory.appending(path: portalItemID.rawValue, directoryHint: .isDirectory)
     }
     
+    /// Calculates the size of a directory and all its contents.
+    /// - Parameter url: The directory's URL.
+    /// - Returns: The total size in bytes.
+    func sizeOfDirectory(at url: URL) -> Int {
+        guard let enumerator = enumerator(at: url, includingPropertiesForKeys: [.fileSizeKey]) else { return 0 }
+        var size = 0
```
philium, line 351: I would suggesting naming this `accumulatedSize` and then the other one (`fileSize`) could just be named `size`.

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -43,6 +46,11 @@ class PreplannedMapModel: ObservableObject, Identifiable {
     /// A Boolean value indicating if a user notification should be shown when a job completes.
     let showsUserNotificationOnCompletion: Bool
     
+    /// The file size of the preplanned map area.
+    var directorySize: Int {
+        FileManager.default.sizeOfDirectory(at: mmpkDirectory)
+    }
```
philium, line 52: Calculating the size could be time consuming, definitely not something to do on-demand, especially during a view update. Instead, it would be better to calculate the size as part of an asynchronous operation (inside of `downloadPreplannedMapArea()`?).

des12437 (approved):

Sources/ArcGISToolkit/Components/Offline/PreplannedMetadataView.swift
```diff
@@ -0,0 +1,118 @@
...
+                    } else {
+                        Text(model.preplannedMapArea.description)
+                            .font(.subheadline)
+                    }
+                }
+                VStack(alignment: .leading) {
+                    Text("Size")
+                        .font(.caption)
+                        .foregroundStyle(.secondary)
+                    Text(model.directorySize.formatted(.byteCount(style: .file, allowedUnits: [.kb, .mb])))
```
philium, line 61: Remember to left SwiftUI decide when a value needs to be re-formatted and only call a `formatted` method yourself when absolutely necessary.

```suggestion
                    Text(model.directorySize, format: .byteCount(style: .file, allowedUnits: [.kb, .mb]))
```

Sources/ArcGISToolkit/Components/Offline/PreplannedMetadataView.swift
```diff
@@ -0,0 +1,118 @@
...
+                        Task {
+                            await model.load()
+                        }
+                    }
+                }
+            }
+        }
+        .navigationTitle(model.preplannedMapArea.title)
+        .navigationBarTitleDisplayMode(.inline)
+        .navigationBarBackButtonHidden()
```
philium, line 83: This is not necessary because there is no back button to hide. Removing it makes no difference.

```suggestion
```

Sources/ArcGISToolkit/Components/Offline/PreplannedMetadataView.swift
```diff
@@ -0,0 +1,118 @@
...
+        .toolbar {
+            ToolbarItem(placement: .confirmationAction) {
+                Button {
+                    dismiss()
+                } label: {
+                    Text("Done")
+                        .bold()
+                }
```
philium, line 91: I'm fairly certain that SwiftUI makes the button text bold because of its placement. I don't see a noticeable different with and without the `bold()` modifier.

```suggestion
                Button("Done") {
                    dismiss()
                }
```

By the way, `Button("Done") { ... }` is used in three different places now and they will all need to be localized. I would suggest extending `Button` with a helper method to make that easier:

```swift
extension Button {
    static func done(action: @escaping () -> Void) -> Button where Label == Text {
        return .init(action: action) {
            Text(
                "Done",
                bundle: .toolkitModule,
                comment: "A label for a button for the user to indicate completion of the current task."
            )
        }
    }
}
```

Sources/ArcGISToolkit/Components/Offline/PreplannedMetadataView.swift
```diff
@@ -0,0 +1,118 @@
...
+        }
+    }
+}
+
+#Preview {
+    struct MockPreplannedMapArea: PreplannedMapAreaProtocol {
+        var packagingStatus: PreplannedMapArea.PackagingStatus? = .complete
+        var title: String = "Mock Preplanned Map Area"
+        var description: String = "This is the description text"
+        var thumbnail: LoadableImage? = nil
```
philium, line 102: It doesn't make a huge difference, but since these are all constants, they should just be computed.

```suggestion
        var packagingStatus: PreplannedMapArea.PackagingStatus? { .complete }
        var title: String { "Mock Preplanned Map Area" }
        var description: String { "This is the description text" }
        var thumbnail: LoadableImage? { nil }
```

Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
```diff
@@ -169,10 +169,14 @@ class PreplannedMapModelTests: XCTestCase {
...
-            preplannedMapAreaID: areaID
-        )
+        let mmpkDirectory = URL.documentsDirectory
+            .appending(path: "OfflineMapAreas")
+            .appending(path: portalItem.id!.rawValue)
+            .appending(path: "Preplanned")
+            .appending(
+                path: areaID.rawValue,
+                directoryHint: .isDirectory
+            )
```
philium, line 179: Apple added a `URL` API specifically so that developers don't have to call `append` with each separate component:

```suggestion
            .appending(
                components: "OfflineMapAreas",
                "\(portalItem.id!)",
                "Preplanned",
                "\(areaID)/"
            )
```

Sources/ArcGISToolkit/Components/Offline/PreplannedMetadataView.swift
yo1995: For some reason, if I apply this change, it asks me to convert the `Int` to `Int64`, such as `Int64(model.directorySize), format: .byteCount(style: .file, allowedUnits: [.kb, .mb])` while the `formatted` method doesn't require so.

I'll convert the data type.

Sources/ArcGISToolkit/Components/Offline/PreplannedMetadataView.swift
yo1995: 🤦 Copy-pasted code without scrutinizing.

Sources/ArcGISToolkit/Components/Offline/PreplannedMetadataView.swift
philium: Interesting. I wonder why the compiler didn't complain when I tried it? 🤔

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
```diff
@@ -55,6 +75,19 @@ struct PreplannedListItemView: View {
             .font(.body)
     }
     
+    @ViewBuilder private var deleteButton: some View {
+        if model.status.allowsRemoval {
+            Button("Delete") {
+                onDelete()
+                model.removeDownloadedPreplannedMapArea()
+                Task {
+                    await model.load()
```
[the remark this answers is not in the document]
yo1995: I agree that it is different from the loadable's load method. To me, it is loading a property of the data model, so maybe `model.loadPreplannedMapArea()`? Will leave this open for discussion.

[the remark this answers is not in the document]
yo1995: > it seems really strange the deleting a local cache would kick off an asynchronous operation that (presumably) goes out over the wire

I kind of agree with your assessment. What I want to achieve with the delete button is this: it deletes the local files for a preplanned map area, then "refresh" the status of the model.

A few factors can affect the new status of the model: network connection is good or not, whether the preplanned map area's packaging status was changed on AGOL. Thus I called the async `load()` method to determine the new status, which I think the is unavoidable.

I thought about calling the `load()` method in the `removeDownloadedPreplannedMapArea()` method but that seems to be overreaching. IMO, removing files from local disk should be a synchronous operation.

Not sure how calling `load()` on-demand would work here - once the delete swipe is done, the list row needs to reflect the latest model status in UI.

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -223,6 +238,21 @@ extension PreplannedMapModel {
...
+                true
+            default:
+                false
+            }
+        }
+        
+        /// A Boolean value indicating whether the preplanned map area is downloaded.
+        var isDownloaded: Bool {
+            if case .downloaded = self { true } else { false }
+        }
```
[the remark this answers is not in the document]
yo1995: The convenience computed properties solely rely on the `Status` itself. If they are moved to the `Model` type, it would be more complicated to unit test the convenience properties against each individual status cases as now they rely on `model.status` property - which means I need to test each status in a different testcase.

Is that what you suggest here? I can make the change to both these properties and their tests.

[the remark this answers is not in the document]
philium: It is up to you. If it were me, I would prioritize convenience at the point of use by not putting these on the nested type. It may seem like it would make testing more complicated, but it doesn't need to. For example, you could model the model—or part of it—with a protocol and then put those convenience properties on the protocol, making them much easier to test. But like I said, it is up to you.

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -172,7 +186,10 @@ class PreplannedMapModel: ObservableObject, Identifiable {
             let result = await job.result
             guard let self else { return }
             self.updateDownloadStatus(for: result)
-            self.mobileMapPackage = try? result.map { $0.mobileMapPackage }.get()
+            if status.isDownloaded {
+                self.mobileMapPackage = try? result.get().mobileMapPackage
+                self.directorySize = FileManager.default.sizeOfDirectory(at: mmpkDirectoryURL)
```
philium, line 191: The size is still being calculated on the main thread, but at least it isn't being done in a view update anymore.

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -313,18 +337,30 @@ extension FileManager {
...
+            }
+            accumulatedSize += size
+        }
+        return accumulatedSize
+    }
+    
     /// The path to the preplanned map areas directory for a specific portal item.
-    /// `Documents/OfflineMapAreas/<Portal Item ID>/Preplanned/<Preplanned Area ID>`
+    /// `Documents/OfflineMapAreas/<Portal Item ID>/Preplanned/<Preplanned Area ID>/`
     /// - Parameter portalItemID: The ID of the web map portal item.
```
philium, line 357: Missing doc for `preplannedMapAreaID` parameter.

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -313,18 +337,30 @@ extension FileManager {
...
     ) -> URL {
         portalItemDirectory(forPortalItemID: portalItemID)
-            .appending(
-                path: Self.preplannedDirectoryPath,
-                directoryHint: .isDirectory
-            )
+            .appending(path: "Preplanned/")
             .appending(
                 path: preplannedMapAreaID.rawValue,
                 directoryHint: .isDirectory
```
philium, line 366: These should be combined:

```swift
.appending(components: "Preplanned", "\(preplannedMapAreaID)/")
```

Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
```diff
@@ -279,18 +279,61 @@ class PreplannedMapModelTests: XCTestCase {
...
-            let directory = FileManager.default.preplannedDirectory(
-                forPortalItemID: portalItem.id!,
-                preplannedMapAreaID: areaID
-            )
-            try? FileManager.default.removeItem(at: directory)
+            model.removeDownloadedPreplannedMapArea()
         }
         
+        var statuses: [PreplannedMapModel.Status] = []
+        var subscriptions = Set<AnyCancellable>()
```
philium, line 301: Not a big deal, but I suggest using syntactic sugar for collection initialization.

```suggestion
        var subscriptions: Set<AnyCancellable> = []
```

Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
```diff
@@ -279,18 +279,61 @@ class PreplannedMapModelTests: XCTestCase {
...
+        // Verify statuses.
+        XCTAssertEqual(
+            statuses,
+            [.notLoaded, .loading, .packaged, .downloading, .downloading, .downloaded]
+        )
+    }
+    
+    @MainActor
+    func testRemoveDownloadedPreplannedMapArea() async throws {
+        let portalItem = PortalItem(portal: Portal.arcGISOnline(connection: .anonymous), id: .init("acc027394bc84c2fb04d1ed317aac674")!)
```
philium, line 331: This line is rather long.

```suggestion
        let portalItem = PortalItem(
            portal: .arcGISOnline(connection: .anonymous),
            id: .init("acc027394bc84c2fb04d1ed317aac674")!
        )
```

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
yo1995: I'm not entirely sure how to "dispatch" a long-running synchronous operation onto a background thread in Swift Concurrency. Do you have a specific suggestion in mind?

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
```diff
@@ -55,6 +74,19 @@ struct PreplannedListItemView: View {
             .font(.body)
     }
     
+    @ViewBuilder private var deleteButton: some View {
+        if model.status.allowsRemoval {
+            Button("Delete") {
+                model.removeDownloadedPreplannedMapArea()
+                onDeletion()
```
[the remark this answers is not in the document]
yo1995: Since this PR already involves many changes, I might need to open a separate PR that targets this branch to address the feedback related to selection. I'll try to implement the following:

- If a map area is currently being viewed, it cannot be "selected" opened again
- A button to unselect the currently selected area that passes a `nil` value to the parent view
- A way to unselect a selected map that is being deleted

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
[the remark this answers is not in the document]
yo1995: > But figuring out the current state should not be an async operation.

Because the status of the model is dependent on checking AGOL server to get the states such as `packaging`, `packaged`, `packageFailure`, and `downloading`, it seems it has to be asynchronous? 🤔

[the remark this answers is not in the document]
philium: So the state can't be computed. As for figuring out the state, it all depends on how you model the state. There should be a state for that, perhaps even just `unknown` or more specific `readyForPackaging`, etc. Then moving from one state to the next can't happen without an asynchronous operation (in some cases), but it is the operation itself that should drive the state transition. I hope that is making sense.

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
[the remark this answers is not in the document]
philium: Okay, I'll wait and see what you come up with.

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
philium: No, I don't. Remember that not every comment is actionable. This was just thinking out load.

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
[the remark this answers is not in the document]
yo1995: Thanks for the explanation. Unfortunately, async calls are needed for updating status, as the preplanned are on the server may change to an arbitrary state while the local files remain unchanged. Thus, it would be safer to fetch the latest state when the local files are removed.

> but it is the operation itself that should drive the state transition

This totally makes sense! Will move the state transition within the operation.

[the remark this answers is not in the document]
philium: Thanks. I still don't understand why deleting a cache needs to kick off a network request and/or why the state can't just go to `packaged`, but clearly you understand this more than I do, so I'll take your word for it that there isn't another way.

philium (approved):

des12437 (approved):
Looks good. There is a bug where `selectedMap` is set to `nil` every time a map area is deleted, but that should be fixed in the PR to improve selection logics right?

yo1995:
Thanks for approving! I was hoping #867 can get reviewed first and merged into this branch. Then I'll merge this PR.

PR #844, opened 2024-08-22
Add internal `MarkdownView` component

dfeinzimer:
Related Apollo 810

In preparation for the coming `TextFormElement` as apart of the `FeatureFormView` component.


dfeinzimer:
I'm looking into a build compatibility issue with `swift-markdown`, `ArcGIS`, and Previews in Xcode. In the meantime you'll likely need to write a small view for testing this.

Package.swift
```diff
@@ -30,13 +30,15 @@ let package = Package(
         ),
     ],
     dependencies: [
-        .package(url: "https://github.com/Esri/arcgis-maps-sdk-swift", .upToNextMinor(from: "200.5.0"))
+        .package(url: "https://github.com/Esri/arcgis-maps-sdk-swift", .upToNextMinor(from: "200.5.0")),
+        .package(url: "https://github.com/apple/swift-markdown.git", branch: "0.4.0")
```
yo1995, line 34: It seems the URL is auto redirecting to https://github.com/swiftlang/swift-markdown 👀

dfeinzimer: I'll pull .git off the URL. I don't think it's needed.

philium: As Ting pointed out, the repo has been moved from the apple org to the swiftlang org. While GitHub helpfully redirects, we should update this URL so that a redirect isn't necessary.

philium: Also, from what I've seen, most of the time Apple includes `.git` in their dependency URLs. I don't know why, but that might also be to prevent a redirect. I suggest including it.

mhdostal (dismissed):

dfeinzimer:
By also adding `swift-markdown` as a package dependency to the Test Runner project and as a library to the UITests target, the tests will run again.

While not optimal, and the versions will need to be kept in sync, this is the only fix I've found so far.

mhdostal (approved):

Test Runner/Test Runner.xcodeproj/project.pbxproj
philium, line 1: These changes are not the correct solution to the build problems. Looking at the build log, there's an indication of what is going on:

image

If we open that file, we can immediately see the culprit:

```swift
@testable import ArcGISToolkit
```

UI tests run in their own process and therefore do not have access to any of the APIs of the thing being tested. ArcGISToolkit should not be imported. (Neither should SwiftUI, by the way.) Removing the unnecessary import(s) allows the UITests target to be built successfully.

PR #845, opened 2024-08-22
`FeatureFormView` - Support `TextFormElement`

dfeinzimer:
Ref Apollo 810


dfeinzimer:
Still investigating why dynamic text isn't updating, but this can be tested with vTest `df/TextFormElement`.

mhdostal (approved):

dfeinzimer:
> Still investigating why dynamic text isn't updating, but this can be tested with vTest `df/TextFormElement`.

The last vTest I'd pulled failed without an obvious error. Validated with a local build that dynamic text is properly updating.

PR #851, opened 2024-08-30
Revise `Markdown` import to `internal`

dfeinzimer:


Sources/ArcGISToolkit/Utility/MarkdownView.swift
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import Markdown
+@_implementationOnly import Markdown
```
philium, line 15: This attribute has been replaced by scoped imports, which are enabled for the toolkit:

https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/056f64666eedf8880098ffabe0dba382e9350e6f/Package.swift#L57

 That is what should be used here:

```swift
import SwiftUI

internal import Markdown
```

dfeinzimer: I had also tried `internal` but it still leaves Xcode Previews broken when the Toolkit Examples scheme is active.

philium: I'm not sure the two are related as importing an internal dependency as internal is the right thing to do. Using an underscored attribute instead of the experiment feature that specifically replaces that attribute is not the right thing to do.

dfeinzimer: Okay, let's proceed with `internal`. Hopefully, the Preview issue will be addressed in a future Xcode release.

dfeinzimer:
It appears this will also help to fix the downstream errors in the daily Swift Samples build.

mhdostal (approved):

PR #852, opened 2024-08-30
Offline: pull to refresh

des12437:
Closes `swift/5939`

- Implements pull to refresh 
- Renames section title from "Preplanned Map Areas" to "Preplanned"

https://github.com/user-attachments/assets/8789a79e-82e0-4788-9a57-fddfb2509350

### DIFF
--- a/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
@@ -25,9 +25,6 @@ public struct OfflineMapAreasView: View {
     /// The action to dismiss the view.
     @Environment(\.dismiss) private var dismiss: DismissAction
     
-    /// A Boolean value indicating whether the preplanned map areas are being reloaded.
-    @State private var isReloadingPreplannedMapAreas = false
-    
     /// The currently selected map.
     @Binding private var selectedMap: Map?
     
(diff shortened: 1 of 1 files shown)


yo1995 (approved):

des12437:
@philium can you please review this PR when you have a chance? Thanks!

philium:
Yep! Sorry. Missed that I was requested to review.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -46,22 +43,7 @@ public struct OfflineMapAreasView: View {
...
-                        } label: {
-                            Image(systemName: "arrow.clockwise")
-                        }
-                        .controlSize(.mini)
-                        .disabled(isReloadingPreplannedMapAreas)
-                    }
-                    .frame(maxWidth: .infinity)
+                    Text("Preplanned").bold()
                 }
                 .textCase(nil)
```
philium, line 48: Is there a strong reason not to use the default style for section headers?

image

Actually, do we even need a section header? It is a bit odd to have one since the view itself has a title and there is only one section. Could we get rid of it?

```swift
Section {
    preplannedMapAreaViews
}
```

image

des12437: > Is there a strong reason not to use the default style for section headers?
> 
This change was made previously to be consistent with the section header styling used throughout Apple Maps.

> Actually, do we even need a section header? It is a bit odd to have one since the view itself has a title and there is only one section. Could we get rid of it?
> 
We need a section header to differentiate preplanned from on-demand map areas when we plan to support the on-demand workflow.

philium: Okay, thanks. In that case, could you please follow our style of putting each modifier on a single line (or more)?

```swift
Text("Preplanned")
    .bold()
```

yo1995 (approved):

philium (approved):

des12437:
Thank you both!

PR #861, opened 2024-09-05
Offline: Add support to view downloaded map areas offline

des12437:
Closes `swift/5582`

- Adds support to view downloaded preplanned map areas when the device is offline
    - Creates preplanned map models by using downloaded mmpk files for the online map. Although a downloaded map area does not need the offline map task in the PreplannedMapModel, the logic for the downloaded state is needed. An OfflinePreplannedMapModel could be made but a significant portion of the PreplannedMapModel logic would be redundantly copied over.


des12437:
Closing in favor of https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/866.

PR #862, opened 2024-09-06
Location Button

rolson:
Adds a button to the toolkit that allows a user to show their current location.

The view can be positioned anywhere, with any background. This example uses a "regular material background" in a rounded rectangle.

- Tap once, location display starts, if using a system location datasource, it will request authorization, while starting a progress indicator is shown
- Starts in auto pan mode recenter
- Long press to get a context menu
- Switch to navigation mode, any autopan mode, or turn the datasource off
- If the location display fails to start, an warning signal is shown, tapping again will try to start the datasource again
- Panning turns autopan off, tapping the button again will turn on the last autopan mode that the user had selected (barring the `off` value)

More details
- The context menu is only shown on long press if the count of the auto pan options the user passes in is greater than 1
- The "last selected auto pan mode" starts out with a default value of the first value passed to the initializer in the `autoPanOptions` that matches a value in this list: `.recenter, .compassNavigation, .navigation`

Location Display Starting
Recenter auto pan mode
Long Press Context Menu
Navigation Mode
Location Display failed to start


rolson:
Next sprint x I can write some tests for this and then we can release it PI1 of 2025.

rolson:
cc @dg0yal

rolson:
Testing, doc, and tutorial now completed.

PR #866, opened 2024-09-11
Offline: Add support to view downloaded map areas offline

des12437:
Closes `swift/5582`

- Adds support to view downloaded preplanned map areas when the device is offline
    - Adds views to show either preplanned map models or offline preplanned map models. 
    - Determines if the device is online or offline.
    - Creates preplanned map models by using downloaded mmpk files for the online map. Although a downloaded map area does not need the offline map task in the PreplannedMapModel, the logic for the downloaded state is needed. An OfflinePreplannedMapModel could be made but a significant portion of the PreplannedMapModel logic would be redundantly copied over.


Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
```diff
@@ -55,10 +67,92 @@ extension OfflineMapAreasView {
...
+            preplannedMapAreaID: PortalItem.ID
+        ) async -> OfflinePreplannedMapArea? {
+            let fileURL = FileManager.default.preplannedDirectory(
+                forPortalItemID: portalItemID,
+                preplannedMapAreaID: preplannedMapAreaID
+            )
+            guard FileManager.default.fileExists(atPath: fileURL.path()) else { return nil }
+            // Make sure the directory is not empty because the directory will exist as soon as the
+            // job starts, so if the job fails, it will look like the mmpk was downloaded.
+            guard !FileManager.default.isDirectoryEmpty(atPath: fileURL) else { return nil }
```
des12437, line 119: This check will be removed when the enhancement in `swift/5937` is implemented to download the mmpk to a temporary directory first which ensures that if the mmpk directory exists it will be non-empty.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -104,6 +116,33 @@ public struct OfflineMapAreasView: View {
...
+                .foregroundStyle(.red)
+            Text(error.localizedDescription)
+        }
+        .frame(maxWidth: .infinity)
+    }
+    
+    /// Makes the preplanned map models.
+    private func makePreplannedMapModels() async {
+        await mapViewModel.makePreplannedMapModels()
+        await mapViewModel.makeOfflinePreplannedMapModels()
```
yo1995, line 144: This isn't ideal - each time the `makePreplannedMapModels` method is called, both the method to load areas from the server, and the method to load mmpks from disk are called, but only 1 group is used. I think the problem stems in that we used the `mapViewModel.preplannedMapModels` to determine what UI is shown, which favors remote areas over local ones.

A thought that we may need some discussion:

- Instead of using `mapViewModel.preplannedMapModels`, split the `preplannedMapAreaViews` into 2 views: `onlinePreplannedMapAreaViews` and `offlinePreplannedMapAreaViews`. Then, call `try await offlineMapTask.preplannedMapAreas` or use `NWPath` to determine which view to show

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -322,7 +322,7 @@ extension PreplannedMapArea: PreplannedMapAreaProtocol {
     }
 }
 
-private extension FileManager {
+extension FileManager {
```
yo1995, line 325:
```suggestion
package extension FileManager {
```
The file URL helpers can be implementation details contained within the package scope

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
```diff
@@ -55,10 +67,93 @@ extension OfflineMapAreasView {
...
+            offlinePreplannedMapModels = mapAreas.map { mapArea in
+                PreplannedMapModel(
+                    offlineMapTask: offlineMapTask,
+                    mapArea: mapArea,
+                    portalItemID: portalItemID,
+                    preplannedMapAreaID: mapArea.id!
+                )
+            }
+            .sorted(using: KeyPathComparator(\.preplannedMapArea.title))
+            .filter({ $0.status.isDownloaded })
```
yo1995, line 100:
```suggestion
            .filter(\.status.isDownloaded)
```

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
des12437: Great suggestion. There are now two views for when the device is online or offline determined by `NWPath`, and only the respective models are created instead of both sets. I verified that the correct views are shown when the internet connection is toggled on and off on a device.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
```diff
@@ -27,19 +27,31 @@ extension OfflineMapAreasView {
         /// The offline map task.
         private let offlineMapTask: OfflineMapTask
         
-        /// The preplanned offline map information.
+        /// The preplanned map information.
         @Published private(set) var preplannedMapModels: Result<[PreplannedMapModel], Error>?
         
+        /// The offline preplanned map information sourced from downloaded mobile map packages.
+        @Published private(set) var offlinePreplannedMapModels: [PreplannedMapModel]?
```
yo1995, line 34: I'm a bit wary whether this needs to be optional or not. The online `preplannedMapModels` uses `nil` to convey that it is still loading. Whereas the loading state is not reflected in the UI for offline models.

What about adding a progress view for the offline models as well, even if it might never be visible thanks to very fast loading from disk?

Otherwise, we can use its emptiness to convey whether there are local areas or not.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
```diff
@@ -27,19 +27,31 @@ extension OfflineMapAreasView {
...
+        
         init(map: Map) {
```
yo1995, line 36: Docstring missed for this, can be added.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
```diff
@@ -27,19 +27,31 @@ extension OfflineMapAreasView {
...
+        nonisolated func requestUserNotificationAuthorization() async {
+            _ = try? await UNUserNotificationCenter.current()
+                .requestAuthorization(options: [.alert, .sound])
+        }
+        
+        /// Gets the preplanned map areas from the offline map task and creates the map models.
+        func makePreplannedMapModels() async {
             guard let portalItemID else { return }
             
+            if offlineMapTask.loadStatus == .failed {
```
yo1995, line 51: With the recent addition of `NWMonitor`, I feel that it is unlikely the `makePreplannedMapModels()` method would be called when the app is offline. However, I think it may be worth keeping this just as a sanity check. 

Can you think of edge cases that require this?

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
```diff
@@ -55,10 +67,93 @@ extension OfflineMapAreasView {
...
-            _ = try? await UNUserNotificationCenter.current()
-                .requestAuthorization(options: [.alert, .sound])
+        /// Makes offline preplanned map models with information from the downloaded mobile map
+        /// packages for the online map.
+        func makeOfflinePreplannedMapModels() async {
+            guard let portalItemID else { return }
+            
+            let preplannedDirectory = FileManager.default.preplannedDirectory(forPortalItemID: portalItemID)
+            
+            guard let mapAreaIDs = try? FileManager.default.contentsOfDirectory(atPath: preplannedDirectory.path()) else { return }
```
yo1995, line 77: [`contentsOfDirectory(at URL)`](https://developer.apple.com/documentation/foundation/filemanager/1413768-contentsofdirectory) equivalent of this method can be used.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
```diff
@@ -55,10 +67,93 @@ extension OfflineMapAreasView {
...
+            
+            offlinePreplannedMapModels = mapAreas.map { mapArea in
+                PreplannedMapModel(
+                    offlineMapTask: offlineMapTask,
+                    mapArea: mapArea,
+                    portalItemID: portalItemID,
+                    preplannedMapAreaID: mapArea.id!
+                )
+            }
+            .sorted(using: KeyPathComparator(\.preplannedMapArea.title))
```
yo1995, line 99: I see a sendability warning on this line in Xcode 16.0, it seems that it will be gone when the upcoming feature becomes formal though, so no need to fix: https://github.com/swiftlang/swift/issues/57560

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
```diff
@@ -55,10 +67,93 @@ extension OfflineMapAreasView {
...
+            .filter(\.status.isDownloaded)
+        }
+        
+        /// Creates a preplanned map area using a given portal item and map area ID to search for a corresponding
+        /// downloaded mobile map package. If the mobile map package is not found then `nil` is returned.
+        /// - Parameters:
+        ///   - portalItemID: The portal item ID.
+        ///   - preplannedMapAreaID: The preplanned map area ID.
+        /// - Returns: The preplanned map area.
+        private func createMapArea(
```
yo1995, line 109: Should it use `make` method naming?

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
```diff
@@ -55,10 +67,93 @@ extension OfflineMapAreasView {
...
+            for portalItemID: PortalItem.ID,
+            preplannedMapAreaID: PortalItem.ID
+        ) async -> OfflinePreplannedMapArea? {
+            let fileURL = FileManager.default.preplannedDirectory(
+                forPortalItemID: portalItemID,
+                preplannedMapAreaID: preplannedMapAreaID
+            )
+            guard FileManager.default.fileExists(atPath: fileURL.path()) else { return nil }
```
yo1995, line 117: Might be an oddball solution, but you can use "fileURL.[checkResourceIsReachable()](https://developer.apple.com/documentation/foundation/url/1779667-checkresourceisreachable)" here instead. It is an Apple [helper method](https://github.com/apple/swift-corelibs-foundation/blob/main/Sources/Foundation/NSURL.swift#L561) that wraps the `atPath` API.

No need to change, up to you.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
```diff
@@ -55,10 +67,93 @@ extension OfflineMapAreasView {
...
+            // Make sure the directory is not empty because the directory will exist as soon as the
+            // job starts, so if the job fails, it will look like the mmpk was downloaded.
+            guard !FileManager.default.isDirectoryEmpty(atPath: fileURL) else { return nil }
+            let mmpk = MobileMapPackage.init(fileURL: fileURL)
```
yo1995, line 121:
```suggestion
            let mmpk = MobileMapPackage(fileURL: fileURL)
```

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -81,19 +94,29 @@ public struct OfflineMapAreasView: View {
...
                 .frame(maxWidth: .infinity)
         }
     }
     
+    @ViewBuilder private var offlinePreplannedMapAreasView: some View {
+        if let models = mapViewModel.offlinePreplannedMapModels,
+           !models.isEmpty {
+            List(models) { preplannedMapModel in
+                PreplannedListItemView(model: preplannedMapModel, selectedMap: $selectedMap) {
+                    Task { await makePreplannedMapModels() }
```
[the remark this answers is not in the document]
des12437: Instead of a comment I added the `onDeletion` closure label.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
des12437: Decided to leave as-is.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
des12437: No I couldn't think of an edge case to require this, but I left this in as a sanity check.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
des12437: I added a progress view which is presented when the offlinePreplannedMapModels are `nil`. The progress view is not visible since the files load from disk quickly, but is good to have in case the files do not load quickly.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
des12437: I did not apply this suggestion since it results in `mapAreaIDs` being an array of `URL` instead of `String`. `String` is more convenient to work with in this case.

des12437:
Here is a screenshot of the offline banner view when the device is offline:

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -57,14 +63,18 @@ public struct OfflineMapAreasView: View {
...
+                await mapViewModel.requestUserNotificationAuthorization()
             }
             .task {
-                await mapViewModel.requestUserNotificationAuthorization()
+                await makePreplannedMapModels()
+                isConnected = networkMonitor.isConnected
+            }
+            .task(id: networkMonitor.isConnected) {
+                offlineBannerIsPresented = !networkMonitor.isConnected
             }
```
yo1995, line 74: This can be synchronous

```swift
            .onChange(of: networkMonitor.isConnected) { _ in
                offlineBannerIsPresented = !networkMonitor.isConnected
            }
```

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -57,14 +63,18 @@ public struct OfflineMapAreasView: View {
                 .textCase(nil)
             }
             .task {
-                await makePreplannedMapModels()
+                await mapViewModel.requestUserNotificationAuthorization()
             }
             .task {
-                await mapViewModel.requestUserNotificationAuthorization()
+                await makePreplannedMapModels()
+                isConnected = networkMonitor.isConnected
```
yo1995, line 70: I feel like the if-else in the async `makePreplannedMapModels` method should be kept, while setting `isConnected = networkMonitor.isConnected` goes before the async call.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
```diff
@@ -55,10 +69,93 @@ extension OfflineMapAreasView {
...
+        func makeOfflinePreplannedMapModels() async {
+            guard let portalItemID else { return }
+            
+            let preplannedDirectory = FileManager.default.preplannedDirectory(forPortalItemID: portalItemID)
+            
+            guard let mapAreaIDs = try? FileManager.default.contentsOfDirectory(atPath: preplannedDirectory.path()) else { return }
+            
+            var mapAreas: [OfflinePreplannedMapArea] = []
+            
+            for mapAreaID in mapAreaIDs {
```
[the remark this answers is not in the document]
des12437: I tried to implement this suggestion but could not resolve an error due to an async method `makeMapArea()` being called inside compactMap.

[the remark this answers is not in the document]
yo1995: Sorry I made a wrong suggestion here.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
```diff
@@ -55,10 +69,93 @@ extension OfflineMapAreasView {
...
+    var packagingStatus: PreplannedMapArea.PackagingStatus?
+    var title: String
+    var description: String
+    var thumbnail: LoadableImage?
+    var id: PortalItem.ID?
+    
+    func retryLoad() async throws {}
+    func makeParameters(using offlineMapTask: OfflineMapTask) async throws -> DownloadPreplannedOfflineMapParameters {
+        throw CancellationError()
+    }
```
yo1995, line 148: I don't have a suggestion, but I'd like to know other reviewer's idea on this.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -140,3 +216,26 @@ public struct OfflineMapAreasView: View {
     }
     return OfflineMapAreasViewPreview()
 }
+
+@MainActor
+private class NetworkMonitor: ObservableObject {
+    /// The path mointor to observe network changes.
```
yo1995, line 222:
```suggestion
    /// The path monitor to observe network changes.
```

yo1995 (approved):

des12437:
Here is a screenshot of the updated offline empty state view:

yo1995 (approved):

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -85,26 +105,53 @@ public struct OfflineMapAreasView: View {
...
     }
     
-    @ViewBuilder private var emptyPreplannedMapAreasView: some View {
+    @ViewBuilder private var offlinePreplannedMapAreasView: some View {
+        if let models = mapViewModel.offlinePreplannedMapModels {
+            if !models.isEmpty {
+                List(models) { preplannedMapModel in
+                    PreplannedListItemView(model: preplannedMapModel, selectedMap: $selectedMap, onDeletion: {
+                        Task { await makePreplannedMapModels() }
+                    })
```
philium, line 121: I'm wondering if the on deletion closure isn't necessary. Is there a way that a task modifier could be used here? Something like this:

```swift
PreplannedListItemView(model: preplannedMapModel, selectedMap: $selectedMap)
    .task(preplannedMapModel.status) { // Does not work since status is not equatable.
        guard preplannedMapModel.status == .needsNewModels else { return }
        await makePreplannedMapModels()
    }
```

But then having thought of it some more, this really does seem like something the model itself should do, right? I just think there needs to be a way to coalesce those updates. Think of the scenario where a user deletes multiple areas in rapid succession. That's probably out of the scope of this PR though.

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -322,7 +322,7 @@ extension PreplannedMapArea: PreplannedMapAreaProtocol {
     }
 }
 
-private extension FileManager {
+package extension FileManager {
```
philium, line 325: The `pakage` modifier is used for exposing a declaration across targets of the same package without making it public. That doesn't apply in this situation, so it doesn't make sense to use it. Remember, use the lowest access level possible for a declaration. In this case, the extension should be internal (no access level keyword). Then, since this extension is no longer private to this file, move it into a new file in this same directory called something like `FileManager+Offline.swift`. And now that they are no longer private, you can write unit tests for the enclosing declarations. Yay!

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
```diff
@@ -85,6 +88,7 @@ struct PreplannedListItemView: View {
            !isSelected {
             Button("Delete") {
                 model.removeDownloadedPreplannedMapArea()
+                onDeletion()
             }
             .tint(.red)
```
philium, line 93: There is no difference that I can see.

```suggestion
            Button("Delete", role: .destructive) {
                model.removeDownloadedPreplannedMapArea()
                onDeletion()

```

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
yo1995: Oh I mean we have a protocol method to implement for the local map areas, but it is highly likely that this would never be called. Should it be a `fatalError` or similar kind?

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -140,3 +217,26 @@ public struct OfflineMapAreasView: View {
...
+    
+    init() {
+        monitor.pathUpdateHandler = { [unowned self] path in
+            DispatchQueue.main.async {
+                self.isConnected = path.status == .satisfied
+            }
+        }
+        let queue = DispatchQueue(label: "NetworkMonitor")
+        monitor.start(queue: queue)
+    }
```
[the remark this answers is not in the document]
yo1995: In this case, we can maybe conform the `NetworkMonitor` to `Cancellable` protocol as well?

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
philium: If you're sure that the method won't be called, then fatal error is fine.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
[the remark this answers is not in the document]
philium: Uh, no. `Cancellable` is provided by `Combine`, which isn't even being used here. How would that be beneficial?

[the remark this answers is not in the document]
yo1995: Dang… I thought it was Swift Concurrency. NVM

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
```diff
@@ -27,19 +27,33 @@ extension OfflineMapAreasView {
...
-        /// offline map models.
-        func makePreplannedOfflineMapModels() async {
+        /// Requests authorization to show notifications.
+        nonisolated func requestUserNotificationAuthorization() async {
+            _ = try? await UNUserNotificationCenter.current()
+                .requestAuthorization(options: [.alert, .sound])
+        }
+        
+        /// Gets the preplanned map areas from the offline map task and creates the map models.
+        func makePreplannedMapModels() async {
```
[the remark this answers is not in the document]
des12437: I renamed these methods with a "load" prefix instead of "make".

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -67,16 +87,16 @@ public struct OfflineMapAreasView: View {
...
     }
     
-    @ViewBuilder private var preplannedMapAreaViews: some View {
+    @ViewBuilder private var preplannedMapAreasView: some View {
         switch mapViewModel.preplannedMapModels {
         case .success(let models):
             if !models.isEmpty {
                 List(models) { preplannedMapModel in
-                    PreplannedListItemView(model: preplannedMapModel, selectedMap: $selectedMap)
+                    PreplannedListItemView(model: preplannedMapModel, selectedMap: $selectedMap) {}
```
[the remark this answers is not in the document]
des12437: The closure is now optional so the empty closure has been omitted where applicable.

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
des12437: This suggestion has the side effect of removing the list item without changing the list's data source. The destructive button style works well for the offline case since the list data source does not need to be reloaded and the `onDeletion` closure can be removed. However, the destructive button style does not work for the online case where a deleted list item should not be removed from the list.

philium: In that case, consider using a different button color and title ("Remove Cache"). Having it be so similar to a destructive delete button but not function like one is rather surprising.

des12437: Okay, to support that the `PreplannedListItemView` would need a parameter to indicate that the view is online or offline to determine if the "remove cache" or "delete" button is used respectively.

Sources/ArcGISToolkit/Components/Offline/Utilities/FileManager+Offline.swift
```diff
@@ -0,0 +1,80 @@
...
+
+extension FileManager {
+    /// The path to the documents folder.
+    private var documentsDirectory: URL {
+        URL.documentsDirectory
+    }
+    
+    /// The path to the offline map areas directory within the documents directory.
+    /// `Documents/OfflineMapAreas/`
+    private var offlineMapAreasDirectory: URL {
```
philium, line 26: This is only used in the following declaration, so it should be included there:

```swift
return .documentsDirectory
    .appending(components: "OfflineMapAreas", "\(portalItemID)/")
```

Sources/ArcGISToolkit/Components/Offline/Utilities/FileManager+Offline.swift
```diff
@@ -0,0 +1,80 @@
...
+        documentsDirectory.appending(component: "OfflineMapAreas/")
+    }
+    
+    /// The path to the web map directory for a specific portal item.
+    /// `Documents/OfflineMapAreas/<Portal Item ID>/`
+    /// - Parameter portalItemID: The ID of the web map portal item.
+    private func portalItemDirectory(forPortalItemID portalItemID: PortalItem.ID) -> URL {
```
philium, line 33: This should be on `URL`, not `FileManager`. Please move this to a private URL extension below this extension.

Sources/ArcGISToolkit/Components/Offline/Utilities/FileManager+Offline.swift
```diff
@@ -0,0 +1,80 @@
...
+    ///   - portalItemID: The ID of the web map portal item.
+    ///   - preplannedMapAreaID: The ID of the preplanned map area portal item.
+    /// - Returns: A URL to the preplanned map area directory.
+    func preplannedDirectory(
+        forPortalItemID portalItemID: PortalItem.ID,
+        preplannedMapAreaID: PortalItem.ID
+    ) -> URL {
+        portalItemDirectory(forPortalItemID: portalItemID)
+            .appending(components: "Preplanned", "\(preplannedMapAreaID)/")
+    }
```
philium, line 73: Please combine (making `preplannedMapAreaID` optional) and then move to the private URL extension below this one.

Sources/ArcGISToolkit/Components/Offline/Utilities/FileManager+Offline.swift
des12437: I tried to move these methods to a URL extension but `preplannedDirectory(forPortalItemID:preplannedMapAreaID)` needs to be accessible outside this file and for some reason the method is only available as `URL.preplannedDirectory(self: URL)`.

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
des12437: I will log an issue and address this in another PR.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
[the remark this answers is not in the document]
des12437: After implementing this suggestion I realized it introduced a timing bug when the device is fully offline. Since `isConnected` is set to `true` by default, the online models are created before the network connection is set by the network monitor so an error message is presented for no internet connection. This issue is fixed by the latest commit [cbbaf44](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/866/commits/cbbaf444163c99c99662675240a14f323fb0d517).

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -223,20 +222,16 @@ private class NetworkMonitor: ObservableObject {
...
-                self.isConnected = path.status == .satisfied
-            }
+    func startMonitoring(_ onChange: @MainActor @Sendable @escaping (_ isConnected: Bool) -> Void) {
+        monitor.pathUpdateHandler = { path in
+            let isConnected = path.status == .satisfied
+            MainActor.assumeIsolated { onChange(isConnected) }
         }
-        let queue = DispatchQueue(label: "NetworkMonitor")
-        monitor.start(queue: queue)
+        monitor.start(queue: .main)
```
yo1995: Initially I think this change isn't good as we are monitoring some low priority stuff on the main queue.

But I gradually accepted that at some point the Boolean state property needs to be changed on main actor. So just require the escaping closure to run on main actor shouldn't be too much an overhead.

yo1995 (approved):

Sources/ArcGISToolkit/Components/Offline/Utilities/FileManager+Offline.swift
yo1995: Try `static func preplannedDirectory`

des12437: Nevermind, I just applied this feedback.

Sources/ArcGISToolkit/Components/Offline/Utilities/FileManager+Offline.swift
```diff
@@ -49,13 +31,19 @@ extension FileManager {
...
-    /// The path to the preplanned map areas directory for a specific portal item.
-    /// `Documents/OfflineMapAreas/<Portal Item ID>/Preplanned/`
+    /// Returns a Boolean value indicating if the specified directory is empty.
+    /// - Parameter path: The path to check.
+    func isDirectoryEmpty(atPath path: URL) -> Bool {
+        (try? FileManager.default.contentsOfDirectory(atPath: path.path()).isEmpty) ?? true
+    }
+}
+
+extension URL {
```
yo1995, line 41: Do we need another extension file `URL+Offline`? 😥

Sources/ArcGISToolkit/Components/Offline/Utilities/FileManager+Offline.swift
```diff
@@ -0,0 +1,65 @@
...
+        URL.documentsDirectory.appending(components: "OfflineMapAreas", "\(portalItemID)/")
+    }
+    
+    /// The path to the directory for a specific map area from the preplanned map areas directory for a specific portal item.
+    /// `Documents/OfflineMapAreas/<Portal Item ID>/Preplanned/<Preplanned Area ID>/`
+    /// - Parameters:
+    ///   - portalItemID: The ID of the web map portal item.
+    ///   - preplannedMapAreaID: The ID of the preplanned map area portal item.
+    /// - Returns: A URL to the preplanned map area directory.
+    static func preplannedDirectory(
```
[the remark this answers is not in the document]
des12437: > please move the private keyword from the previous declaration to the extension itself.

I kept the extension as internal since `preplannedDirectory()` needs to be accessed outside the file.

Sources/ArcGISToolkit/Components/Offline/Utilities/FileManager+Offline.swift
```diff
@@ -0,0 +1,65 @@
...
+                continue
+            }
+            accumulatedSize += size
+        }
+        return accumulatedSize
+    }
+    
+    /// Returns a Boolean value indicating if the specified directory is empty.
+    /// - Parameter path: The path to check.
+    func isDirectoryEmpty(atPath path: URL) -> Bool {
```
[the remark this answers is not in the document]
des12437: Thanks for the feedback. I removed the method since the `isDirectoryEmpty` check is no longer needed since the enhancement in #887 was made.

Now the `FileManager+Offline` file only has the `sizeOfDirectory()` method which is only used in `PreplannedMapModel`. Should the FileManager extension be moved back into that class?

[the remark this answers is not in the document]
philium: Yes, that would be fine.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -14,22 +14,26 @@
...
-    
     /// The web map to be taken offline.
     private let onlineMap: Map
-    
     /// The currently selected map.
     @Binding private var selectedMap: Map?
+    /// A Boolean value indicating whether the device has an internet connection.
+    @State private var isConnected = true
+    /// A Boolean value indicating whether the offline banner should be presented.
+    @State private var offlineBannerIsPresented = false
```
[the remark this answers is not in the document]
des12437: I refactored the network connection logic in [191a3ff](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/866/commits/191a3ff3d4a787febdb832033f0616c3ed0e69f6). `isConnected` should only be set when the view loads and upon pull to refresh while `offlineBannerIsPresented` should be updated upon network connection changes.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -47,16 +51,35 @@ public struct OfflineMapAreasView: View {
...
+                await loadPreplannedMapModels()
+            }
+            .onAppear {
+                networkMonitor.startMonitoring { isConnected in
+                    offlineBannerIsPresented = !isConnected
+                    self.isConnected = isConnected
+                }
+            }
+            .onChange(of: isConnected) { _ in
+                Task { await loadPreplannedMapModels() }
```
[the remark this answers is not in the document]
des12437: I removed this logic and added a Boolean value `connectionIsDetermined` so that the models are loaded once when the network connection is determined.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -14,22 +14,28 @@
...
     @Environment(\.dismiss) private var dismiss: DismissAction
-    
     /// The web map to be taken offline.
     private let onlineMap: Map
-    
     /// The currently selected map.
     @Binding private var selectedMap: Map?
+    /// A Boolean value indicating whether the device connection has been determined.
+    @State private var connectionIsDetermined = false
+    /// A Boolean value indicating whether the device has an internet connection.
```
philium, line 35: Does this doc need to be updated? It still sounds like it is just the inverse of `offlineBannerIsPresented`.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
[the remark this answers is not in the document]
philium: Could `isConnected` be optional? Then `nil` could mean not determined yet. I'm just looking at three different properties related to connection (`connectionIsDetermined`, `isConnected`, and `offlineBannerIsPresented`) and can't help but wonder if they could be consolidated into two.

Sources/ArcGISToolkit/Components/Offline/Utilities/URL+Offline.swift
```diff
@@ -0,0 +1,43 @@
...
+
+import ArcGIS
+import Foundation
+
+extension URL {
+    /// The path to the web map directory for a specific portal item.
+    /// `Documents/OfflineMapAreas/<Portal Item ID>/`
+    /// - Parameter portalItemID: The ID of the web map portal item.
+    static private func portalItemDirectory(forPortalItemID portalItemID: PortalItem.ID) -> URL {
+        URL.documentsDirectory.appending(components: "OfflineMapAreas", "\(portalItemID)/")
```
philium, line 23: Explicit `URL` isn't needed.

```suggestion
        return .documentsDirectory.appending(components: "OfflineMapAreas", "\(portalItemID)/")
```

Sources/ArcGISToolkit/Components/Offline/Utilities/FileManager+Offline.swift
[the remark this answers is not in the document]
philium: Okay.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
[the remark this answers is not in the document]
yo1995: While it may be good to get these 3 Booleans consolidated, upon further discussion today, Ryan suggested to not use the `NWPathMonitor` that proactively listen to network connectivity updates. Instead, we would use the loadStatus/error to determine whether the component has Internet access or not.

Thus, `offlineBannerIsPresented` would be gone shortly, and `connectionIsDetermined` might also be represented by the nullability of `isConnected` I guess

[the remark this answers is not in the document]
yo1995: p.s. I've discussed with Destiny. We can follow one of the 2 paths

- apply your suggestion here, and later refactor slightly to remove the `NWPathMonitor` related code
- branch off a separate PR (b) to remove `NWPathMonitor`, and resolve all other comments in this PR (a). First merge this PR (a), then merge the PR (b)

des12437:
> Could isConnected be optional? Then nil could mean not determined yet. I'm just looking at three different properties related to connection (connectionIsDetermined, isConnected, and offlineBannerIsPresented) and can't help but wonder if they could be consolidated into two.

I did try this implementation originally but realized that monitoring the change of `isConnected` would cause the models to reload when the network changes and not upon pull to refresh. The additional property `connectionIsDetermined` allows for the models to be loaded once, after the network connection is determined. The offline banner will be removed in a follow up PR so I will leave the documentation as-is.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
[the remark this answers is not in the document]
philium: Does that mean `NetworkMonitor` will be going away? If so, it doesn't make a whole lot of sense to merge it in this PR just to rip it out in a follow-up PR. Could it just be removed in this PR?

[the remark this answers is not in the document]
yo1995: Yes, it will go away. I was afraid that removing it would delay the review of this PR. But if you don't mind, Destiny can make the changes in this PR.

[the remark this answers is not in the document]
philium: Go for it!

des12437:
The network monitor and offline banner have been removed. The `offlinePreplannedMapAreasView` is shown when the offline map task fails to load due to a `notConnectedToInternet` error. This PR is ready for re-review.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -14,20 +14,18 @@
 
 import SwiftUI
 import ArcGIS
+import Network
```
philium, line 17: This is no longer needed, so it should be removed.

```suggestion
```

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -84,31 +82,64 @@ public struct OfflineMapAreasView: View {
...
-                Text(error.localizedDescription)
+        case .failure(let error as URLError):
+            if error.code == .notConnectedToInternet {
+                offlinePreplannedMapAreasView
+            } else {
+                view(for: error)
             }
-            .frame(maxWidth: .infinity)
+        case .failure(let error):
+            view(for: error)
```
philium, line 92: These are really both the same case (failure), so they should be combined and the error type check should be moved inside the switch case:

```swift
case .failure(let error):
    if let urlError = error as? URLError,
       urlError.code == .notConnectedToInternet {
        offlinePreplannedMapAreasView
    } else {
        view(for: error)
    }
```

philium (approved):

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -189,15 +168,10 @@ public struct OfflineMapAreasView: View {
...
     private func loadPreplannedMapModels() async {
-        isConnected = !offlineBannerIsPresented
-        
-        if isConnected {
-            await mapViewModel.loadPreplannedMapModels()
-        } else {
-            await mapViewModel.loadOfflinePreplannedMapModels()
-        }
+        await mapViewModel.loadPreplannedMapModels()
+        await mapViewModel.loadOfflinePreplannedMapModels()
```
yo1995, line 174: I'm still not convinced that we should load both online areas and local areas at the same time for each load. Do you think it makes sense to check if the online models are loaded, before loading the offline models? This way, we avoid the overhead of loading offline models that we don't need when the device is connected.

Sources/ArcGISToolkit/Components/Offline/Utilities/URL+Offline.swift
```diff
@@ -15,35 +15,12 @@
...
-}
-
 extension URL {
     /// The path to the web map directory for a specific portal item.
     /// `Documents/OfflineMapAreas/<Portal Item ID>/`
     /// - Parameter portalItemID: The ID of the web map portal item.
-    private static func portalItemDirectory(forPortalItemID portalItemID: PortalItem.ID) -> URL {
-        URL.documentsDirectory.appending(components: "OfflineMapAreas", "\(portalItemID)/")
+    static private func portalItemDirectory(forPortalItemID portalItemID: PortalItem.ID) -> URL {
+        return .documentsDirectory.appending(components: "OfflineMapAreas", "\(portalItemID)/")
```
yo1995, line 23:
```suggestion
       .documentsDirectory.appending(components: "OfflineMapAreas", "\(portalItemID)/")
```
Up to you, `return` keyword can be [omitted](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0255-omit-return.md) for single expression functions.

philium: This was my suggestion. After omitting the return keyword from single expression functions since that feature was introduced, I've found that I miss it except in cases where the entire function declaration is on a single line. The keyword is an easily identifiable clue that the function returns a value, without having to look for the method's return type, which can be much further to the right. So I've started to include them again.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
des12437: The offline map models are now loaded conditionally.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -140,6 +169,37 @@ public struct OfflineMapAreasView: View {
...
+    /// Loads the online and offline preplanned map models.
+    private func loadPreplannedMapModels() async {
+        await mapViewModel.loadPreplannedMapModels()
+        
+        switch mapViewModel.preplannedMapModels {
+        case .failure(_):
+            await mapViewModel.loadOfflinePreplannedMapModels()
+        default:
+            return
+        }
```
philium, line 201: Since only one case is being checked for and nothing is done in the other case, an if statement might be more appropriate.

```suggestion
        if case .failure = mapViewModel.preplannedMapModels {
            await mapViewModel.loadOfflinePreplannedMapModels()
        }
```

philium (approved):

yo1995 (approved):

des12437:
Thank you both!

PR #867, opened 2024-09-11
Improve selection logics

yo1995:
## Description

- Related to https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/843#discussion_r1747486303
- Close swift/issues/5968

For this PR, a few things have been done:

1. Test if the mmpk files are locked and cannot be deleted when a map created from it is being viewed
    - The result is that the files can be deleted, while the map being viewed cannot show the tiles beyond current LoD. The features can still be shown
    - Upon discussion, this behavior shouldn't be relied upon. Instead, we should disallow deletion when a map is being viewed.
2. How to go back online?
    - Upon discussion, it is not the component's responsibility to "unselect" a map area and go back to online webmap.
    - Thus, a button is added at the app level to go online to show this workflow.
3. Disallow opening again or deleting the map when it is being viewed.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
@@ -27,9 +27,6 @@ struct PreplannedListItemView: View {
     /// A Boolean value indicating whether the metadata view is presented.
     @State private var metadataViewIsPresented = false
     
-    /// The closure to perform when the map selection changes.
-    let onMapSelectionChanged: () -> Void
-    
     /// A Boolean value indicating whether the selected map area is the same
     /// as the map area from this model.
     var selectedMapFromThisModel: Bool {
@@ -98,7 +95,6 @@ struct PreplannedListItemView: View {
                 Task {
                     if let map = await model.loadMobileMapPackage() {
                         selectedMap = map
-                        onMapSelectionChanged()
                     }
                 }
             } label: {
@@ -193,6 +189,6 @@ struct PreplannedListItemView: View {
             preplannedMapAreaID: .init("preview")!
         ),
         selectedMap: .constant(nil)
-    ) { }
+    )
     .padding()
 }
--- a/Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
@@ -31,7 +31,7 @@ struct PreplannedListItemView: View {
     /// as the map area from this model.
     /// The title of a preplanned map area is guaranteed to be unique when it
     /// is created.
-    var selectedMapFromThisModel: Bool {
+    var isSelected: Bool {
         selectedMap?.item?.title == model.preplannedMapArea.title
     }
     
@@ -82,7 +82,7 @@ struct PreplannedListItemView: View {
     
     @ViewBuilder private var deleteButton: some View {
         if model.status.allowsRemoval,
-           !selectedMapFromThisModel {
+           !isSelected {
             Button("Delete") {
                 model.removeDownloadedPreplannedMapArea()
             }
@@ -106,7 +106,7 @@ struct PreplannedListItemView: View {
             }
             .buttonStyle(.bordered)
             .buttonBorderShape(.capsule)
-            .disabled(selectedMapFromThisModel)
+            .disabled(isSelected)
         case .downloading:
             if let job = model.job {
                 ProgressView(job.progress)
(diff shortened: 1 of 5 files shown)


Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -141,7 +141,14 @@ class PreplannedMapModel: ObservableObject, Identifiable {
...
         }
-        return mobileMapPackage.maps.first
+        if let map = mobileMapPackage.maps.first {
+            // Add the preplanned map area ID as a tag to differentiate between
+            // local map areas.
+            map.item!.addTag(preplannedMapAreaID.rawValue)
+            return map
+        } else {
+            return nil
+        }
```
yo1995, line 151: When these changes were added, there were some difficulties to determine if a `Map` object is created from a certain mmpk or not. A few things have been tried:

- Use the local item's various IDs. The item ID is different from its portal item ID; the `originalPortalItemID` is the portal item ID of the parent webmap, not the preplanned map areas'
- Use the local item's `fileURL`. Because a `Map` created from a mmpk is sometimes compressed, this property can be `nil`
- Use the object itself. This would require loading and creating a `Map` object from each mmpk, which is not efficient and asynchronous
- Use other metadata. Title, description, and snippet aren't guaranteed to be unique for each preplanned map area

Eventually I decided to add the preplanned map area's portal item ID as a tag to the local item for differentiating each map object, so it is easy to tell which mmpk it was created from.

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
```diff
@@ -75,7 +84,8 @@ struct PreplannedListItemView: View {
     }
     
     @ViewBuilder private var deleteButton: some View {
-        if model.status.allowsRemoval {
+        if model.status.allowsRemoval,
+           !selectedMapFromThisModel {
```
yo1995, line 88: One problem with not allowing deleting when an area is selected is that, if someone wants to delete all the area from the local disk, he has to set `selectedMap = nil` at app level, or otherwise there would always be a single area that cannot be deleted.

People have agreed on this "unselect" can be handled at app level.

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -141,7 +141,14 @@ class PreplannedMapModel: ObservableObject, Identifiable {
         } catch {
             status = .mmpkLoadFailure(error)
```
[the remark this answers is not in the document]
yo1995: Good call! I think it is worth adding that state, so that we don't need to do optional binding when the map is used. @des12437 What do you think?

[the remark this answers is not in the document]
des12437: Yes I agree a `mmpkLoaded` state would be good to add.

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
yo1995: Because a `mobileMapPackage` needs to be loaded in order to access the maps in it. In current implementation, we only load the packages "on-demand" (when the open button is tapped) so that only 1 package is loaded at a time, to save resources. Thus, when the comparison Boolean is computed, the `model.map` is `nil`.

yo1995 (commented):
I'll try some other ideas before requesting for another review. Using tags isn't ideal, but was the best I could come up with thus far.

Destiny mentioned we may held another unique ID binding "selectedMapAreaID" and I'll explore how that feels.

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
yo1995: After a long winding detour, it turned out that the local item's title is unique.

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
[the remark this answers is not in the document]
yo1995: I added it and tried it out - turned out this state is only written to, not read/used by anywhere else. So I decided to revert adding it.

yo1995:
Ready for review again.

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
```diff
@@ -21,13 +21,19 @@ struct PreplannedListItemView: View {
...
     
-    /// The closure to perform when the map selection changes.
-    let onMapSelectionChanged: (Map) -> Void
-    /// The closure to perform when the map is removed from local disk.
-    let onDeletion: () -> Void
+    /// A Boolean value indicating whether the selected map area is the same
+    /// as the map area from this model.
+    /// The title of a preplanned map area is guaranteed to be unique when it
+    /// is created.
+    var selectedMapFromThisModel: Bool {
```
des12437, line 34:
```suggestion
    var isSelected: Bool {
```
This Boolean should be renamed to read as an assertion.

philium (approved):

des12437 (approved):

yo1995:
Thanks for the feedback!

PR #869, opened 2024-09-12
Use replacement document picker method

zkline101:
Swift issue #6012
This method is [deprecated](https://developer.apple.com/documentation/uikit/uidocumentpickerdelegate/1618680-documentpicker) on iOS and unavailable on visionOS so we should use the [replacement](https://developer.apple.com/documentation/uikit/uidocumentpickerdelegate/2902364-documentpicker).


mhdostal (approved):
Thanks!!!

dfeinzimer (approved):

PR #875, opened 2024-09-17
Add visionOS as a supported platform

zkline101:
Swift #6012

This PR is adding visionOS as a supported platform so the toolkit can build against visionOS. More work is coming after this to make sure all the components work and look correctly on this platform.

Components that are unavailable for visionOS:
- All AR components. Most of the ARKit API we are using isn't available on visionOS.
- Taking a photo or video for feature forms is not available on visionOS.

Xcode 16 must be used to build successfully.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/FlyoverSceneView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/FlyoverSceneView.swift
@@ -21,7 +21,7 @@ import ArcGIS
 @preconcurrency
 @available(visionOS, unavailable)
 public struct FlyoverSceneView: View {
-#if !os(visionOS)
+#if os(iOS)
     /// The AR session.
     @StateObject private var session = ObservableARSession()
 #endif
@@ -137,7 +137,7 @@ public struct FlyoverSceneView: View {
         SceneViewReader { sceneViewProxy in
             sceneViewBuilder(sceneViewProxy)
                 .cameraController(cameraController)
-#if !os(visionOS)
+#if os(iOS)
                 .onAppear {
                     let configuration = ARPositionalTrackingConfiguration()
                     if shouldOrientToCompass {
@@ -166,7 +166,7 @@ public struct FlyoverSceneView: View {
     }
 }
 
-#if !os(visionOS)
+#if os(iOS)
 /// An observable object that wraps an `ARSession` and provides the current frame.
 private class ObservableARSession: NSObject, ObservableObject, ARSessionDelegate {
     /// The backing AR session.
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/FlyoverSceneView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/FlyoverSceneView.swift
@@ -19,9 +19,12 @@ import ArcGIS
 /// A scene view that provides an augmented reality fly over experience.
 @MainActor
 @preconcurrency
+@available(visionOS, unavailable)
 public struct FlyoverSceneView: View {
+#if !os(visionOS)
     /// The AR session.
     @StateObject private var session = ObservableARSession()
+#endif
     /// The initial camera.
     let initialCamera: Camera
     /// The translation factor.
@@ -134,6 +137,7 @@ public struct FlyoverSceneView: View {
         SceneViewReader { sceneViewProxy in
             sceneViewBuilder(sceneViewProxy)
                 .cameraController(cameraController)
+#if !os(visionOS)
                 .onAppear {
                     let configuration = ARPositionalTrackingConfiguration()
                     if shouldOrientToCompass {
@@ -142,7 +146,7 @@ public struct FlyoverSceneView: View {
                     session.start(configuration: configuration)
                 }
                 .onDisappear { session.pause() }
-                .onChange(of: session.currentFrame) { frame in
+                .onChange(session.currentFrame) { frame in
                     guard let frame, let interfaceOrientation else { return }
                     sceneViewProxy.updateCamera(
                         frame: frame,
@@ -150,10 +154,11 @@ public struct FlyoverSceneView: View {
                         orientation: interfaceOrientation
                     )
                 }
-                .onChange(of: initialCamera) { initialCamera in
+#endif
+                .onChange(initialCamera) { initialCamera in
                     cameraController.originCamera = initialCamera
                 }
-                .onChange(of: translationFactor) { translationFactor in
+                .onChange(translationFactor) { translationFactor in
                     cameraController.translationFactor = translationFactor
                 }
                 .observingInterfaceOrientation($interfaceOrientation)
@@ -161,6 +166,7 @@ public struct FlyoverSceneView: View {
     }
 }
 
+#if !os(visionOS)
 /// An observable object that wraps an `ARSession` and provides the current frame.
 private class ObservableARSession: NSObject, ObservableObject, ARSessionDelegate {
     /// The backing AR session.
@@ -190,3 +196,4 @@ private class ObservableARSession: NSObject, ObservableObject, ARSessionDelegate
         currentFrame = frame
     }
 }
+#endif
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#if !os(visionOS)
 import ArcGIS
 import ARKit
 import SwiftUI
@@ -197,3 +198,4 @@ extension ARSwiftUIViewProxy {
         return raycastMatrix
     }
 }
+#endif
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#if !os(visionOS)
+#if os(iOS)
 import ArcGIS
 import ARKit
 import SwiftUI
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/CalibrationView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/CalibrationView.swift
@@ -61,6 +61,7 @@ class WorldScaleCalibrationViewModel: ObservableObject {
     }
 }
 
+@available(visionOS, unavailable)
 extension WorldScaleSceneView {
     /// A view that allows the user to calibrate the heading of the scene view camera controller.
     @MainActor
@@ -179,6 +180,7 @@ extension WorldScaleSceneView {
     }
 }
 
+@available(visionOS, unavailable)
 private extension WorldScaleSceneView.CalibrationView {
     var calibrationLabel: String {
         String(
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/GeoTrackingSceneView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/GeoTrackingSceneView.swift
@@ -25,7 +25,7 @@ public struct GeoTrackingSceneView: View {
     @Binding var initialCameraIsSet: Bool
     /// The view model for the calibration view.
     @ObservedObject private var calibrationViewModel: WorldScaleCalibrationViewModel
-#if !os(visionOS)
+#if os(iOS)
     /// The geo-tracking configuration for the AR session.
     private let configuration = ARGeoTrackingConfiguration()
 #endif
@@ -35,7 +35,7 @@ public struct GeoTrackingSceneView: View {
     private let locationDataSource: LocationDataSource
     /// The closure that builds the scene view.
     private let sceneViewBuilder: (SceneViewProxy) -> SceneView
-#if !os(visionOS)
+#if os(iOS)
     /// The proxy for the ARSwiftUIView.
     private let arViewProxy: ARSwiftUIViewProxy
 #endif
@@ -49,7 +49,7 @@ public struct GeoTrackingSceneView: View {
     @State private var currentLocation: Location?
     /// The current interface orientation.
     @State private var interfaceOrientation: InterfaceOrientation?
-#if !os(visionOS)
+#if os(iOS)
     /// The closure to perform when the camera tracking state changes.
     private var onCameraTrackingStateChangedAction: ((ARCamera.TrackingState) -> Void)?
     /// The closure to perform when the geo tracking status changes.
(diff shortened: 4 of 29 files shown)


Sources/ArcGISToolkit/Extensions/SwiftUI/View+OnChange.swift
```diff
@@ -0,0 +1,40 @@
...
+        if #available(iOS 17.0, macCatalyst 17.0, visionOS 2.0, *) {
+            return onChange(of: value) { _, newValue in
+                action(newValue)
+            }
+        } else {
+            return onChange(of: value) { newValue in
+                action(newValue)
+            }
+        }
+    }
```
zkline101, line 39: The older onChange method we were using is deprecated on visionOS 1.0+ so when you would build the toolkit for visionOS you would see a bunch of deprecation warnings. This method makes sure the new onChange method is being used if it is available.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/TextInput.swift
```diff
@@ -106,6 +106,7 @@ private extension TextInput {
             }
             .focused($isFocused)
             .frame(maxWidth: .infinity, alignment: .leading)
+#if !os(visionOS)
             .toolbar {
                 ToolbarItemGroup(placement: .keyboard) {
                     if UIDevice.current.userInterfaceIdiom == .phone, isFocused, (element.fieldType?.isNumeric ?? false) {
```
zkline101, line 112: I am talking to David and Mark about what to do with this toolbar item on visionOS. It doesn't make sense to put it anywhere other than the keyboard which you can't because the keyboard placement isn't available on visionOS. I have another idea that we are discussing.

Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentImportMenu.swift
```diff
@@ -71,6 +71,7 @@ struct AttachmentImportMenu: View {
         }
     }
     
+    @available(visionOS, unavailable)
     private func takePhotoOrVideoButton() -> Button<some View> {
```
zkline101, line 75: We don't have camera access on the Vision Pro so we shouldn't present this option. If the user tries to use the camera, the device will use your personal persona as the camera feed. So it seems best to not allow this option.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/TextInput.swift
zkline101: Solution in this commit: https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/875/commits/7057555360e68692b2b587fc2523284ac54d61c6.

Discussed with Mark and David.

zkline101:
@philium @dfeinzimer When you have a chance this is ready for review. Thank you!

dfeinzimer:
> Components that are unavailable for visionOS:
> 
> - All AR components. Most of the ARKit API we are using isn't available on visionOS.

Should the AR examples be excluded from compilation at this time?

Screenshot 2024-09-19 at 11 40 37 AM

zkline101:
> Should the AR examples be excluded from compilation at this time?

Yes, they should be excluded and that will happen in an upcoming PR when I add the visionOS destination to the examples app.

zkline101:
Currently, in this PR that error doesn't occur because the example apps don't support visionOS.

dfeinzimer (approved):

Sources/ArcGISToolkit/Components/Augmented Reality/FlyoverSceneView.swift
```diff
@@ -19,9 +19,12 @@ import ArcGIS
 /// A scene view that provides an augmented reality fly over experience.
 @MainActor
 @preconcurrency
+@available(visionOS, unavailable)
 public struct FlyoverSceneView: View {
+#if !os(visionOS)
```
philium, line 24: Why are these compiler directives (here and elsewhere) needed when the type itself is unavailable on visionOS?

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#if !os(visionOS)
```
philium, line 15: Instead conditionally compiling this code, could `ARSwiftUIView` be marked as unavailable on visionOS?

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/CalibrationView.swift
```diff
@@ -61,6 +61,7 @@ class WorldScaleCalibrationViewModel: ObservableObject {
     }
 }
 
+@available(visionOS, unavailable)
```
philium, line 64: Instead of annotating the extension, I suggest annotating the type (`CalibrationView`) itself.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/GeoTrackingSceneView.swift
```diff
@@ -165,6 +178,7 @@ public struct GeoTrackingSceneView: View {
         cameraController.transformationMatrix = .identity
     }
     
+#if !os(visionOS)
```
philium, line 181: Could the enclosing methods just be marked as unavailable on visionOS?

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/WorldTrackingSceneView.swift
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#if !os(visionOS)
```
philium, line 15: Could `WorldTrackingSceneView` be marked as unavailable on visionOS instead?

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
```diff
@@ -15,10 +15,14 @@
 import ARKit
 import SwiftUI
 import ArcGIS
+#if os(visionOS)
+import CoreLocation
+#endif
```
philium, line 20: As far as I can tell, this framework is available on visionOS. Why does this import statement need to be conditionally compiled?

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
```diff
@@ -282,8 +297,10 @@ public struct WorldScaleSceneView: View {
         }
         return try await ARGeoTrackingConfiguration.checkAvailability()
     }
+#endif
 }
 
+@available(visionOS, unavailable)
```
philium, line 303: I suggest putting this on the type itself rather than on the extension.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
```diff
@@ -296,6 +313,7 @@ public extension WorldScaleSceneView {
     }
 }
 
+#if !os(visionOS)
```
philium, line 316: Could the enclosing function be annotated as unavailable on visionOS?

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
```diff
@@ -313,7 +331,9 @@ private extension ARGeoTrackingConfiguration {
         }
     }
 }
+#endif
 
+@available(visionOS, unavailable)
 private extension WorldScaleSceneView {
```
philium, line 337: `WorldScaleSceneView` is annotated as unavailable on visionOS, so why is the conditional compilation necessary.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
```diff
@@ -326,7 +346,9 @@ private extension WorldScaleSceneView {
     }
 }
 
+@available(visionOS, unavailable)
 public extension WorldScaleSceneView {
+#if !os(visionOS)
```
philium, line 351: Are either of these necessary? It seems like neither of them is because `WorldScaleSceneView` is already marked as unavailable on visionOS.

Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentCameraController.swift
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#if !os(visionOS)
```
philium, line 15: Could the enclosing types be annotated as unavailable on visionOS instead?

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/TextInput.swift
```diff
@@ -134,7 +136,23 @@ private extension TextInput {
...
+#else
+            if fieldType.isFloatingPoint {
+                return .decimalPad
+            } else {
+                return .numberPad
+            }
+#endif
+        } else {
+            return .default
+        }
```
philium, line 155: You could simplify this using an `if` expression.

```suggestion
        return if fieldType.isNumeric {
#if os(visionOS)
            // The 'positiveNegativeButton' doesn't show on visionOS
            // so we need to show this keyboard so the user can type
            // a negative number.
            .numbersAndPunctuation
#else
            if fieldType.isFloatingPoint { .decimalPad } else { .numberPad }
#endif
        } else {
            .default
        }
```

Sources/ArcGISToolkit/Extensions/ArcGIS/SceneViewProxy.swift
```diff
@@ -17,6 +17,7 @@ import ArcGIS
 import ARKit
 import SwiftUI
 
+#if !os(visionOS)
```
philium, line 20: Could the enclosing declarations be annotated as unavailable on visionOS instead?

philium (commented):
I asked a lot of questions. Normally I would try to answer them myself, but I don't have a build of core w/ visionOS support.

Sources/ArcGISToolkit/Components/Augmented Reality/FlyoverSceneView.swift
zkline101: I wondered that myself when I started this. Even if the type is marked as unavailable and will never be used on visionOS, the compiler will still gives errors on the areas that are unavailable or not visible on visionOS.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
zkline101: I thought about that and that is an option but you will still need the compiler directives so I thought since this type is internal we can just mark out the whole thing with the compiler directive to make things easier.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/CalibrationView.swift
zkline101: Why annotate `CalibrationView` as `unavailable` if its enclosing type is `unavailable` (`WorldScaleSceneView`)?

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/GeoTrackingSceneView.swift
zkline101: They could yes but they would still need the compiler directives to compile successfully on visionOS.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/WorldTrackingSceneView.swift
zkline101: Same response as [here](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/875/files#r1768811353).

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
zkline101: Oh this one was interesting. If I remove these few lines of code here and compile for visionOS I will get this error:

> Cannot find 'CLLocationManager' in scope

but only for visionOS not the other platforms. So that is why I added this import.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
zkline101: It is already on `WorldScaleSceneView` and all extensions of that type will need it as well.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
zkline101: Same response as [here](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/875/files#r1768811353).

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
zkline101: If a type is marked as unavailable then all extensions of that type will need that marking as well.

dfeinzimer (approved):

Sources/ArcGISToolkit/Components/Augmented Reality/FlyoverSceneView.swift
philium: I see. Following your instructions, I got a build with visionOS support and I understand now what the problem is. Since the relevant types are marked as available on iOS rather than unavailable on visionOS:

image

I suggest changing the compiler directive to `os(iOS)` instead of `!os(visionOS)`.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
philium: Seriously?? That's disappointing. Perhaps there's a valid reason for that, but I can't think of one.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
philium: I get adding the import. My question is why not import it on other platforms? I suggest doing that as the code in the file is clearly dependent on it on all platforms. It must be implicitly imported (probably by ARKit), but there's nothing wrong with making the dependency explicit.

zkline101:
@dfeinzimer @philium This is ready for another review. Thank you.

philium (approved):

dfeinzimer (approved):

zkline101:
Exciting! Thank you both for the reviews.

PR #878, opened 2024-09-19
Edit UI for Authenticator

zkline101:
cc: @rolson @philium 

Swift #6015

These edits were originally meant to give `CredentialInputSheetView` a more native feel on visionOS but then I realized these edits can also apply to iOS and Mac Catalyst. These edits make the credential sheet look more like the credential sheet in the Mail app. I explain more of my changes in the comments below.

visionOS before changes:
CredentialInput_before

visionOS after changes:
After

Apple's credential form in Mail app:
Credential_Apple

iOS after changes:
iPhone

### DIFF
--- a/Sources/ArcGISToolkit/Components/Authentication/TrustHostViewModifier.swift
+++ b/Sources/ArcGISToolkit/Components/Authentication/TrustHostViewModifier.swift
@@ -30,25 +30,20 @@ struct TrustHostViewModifier: ViewModifier {
     /// A Boolean value indicating whether or not the prompt is displayed.
     @State var isPresented: Bool = false
     
-    var title: some View {
-        Text(
-            "Certificate Trust Warning",
+    var title: String {
+        String(
+            localized: "Certificate Trust Warning",
             bundle: .toolkitModule,
             comment: "A label indicating that the remote host's certificate is not trusted."
         )
-        .font(.title)
-        .multilineTextAlignment(.center)
     }
     
-    var message: some View {
-        Text(
-            "Dangerous: The certificate provided by '\(challenge.host)' is not signed by a trusted authority.",
+    var message: String {
+        String(
+            localized: "Dangerous: The certificate provided by '\(challenge.host)' is not signed by a trusted authority.",
             bundle: .toolkitModule,
             comment: "A warning that the host service (challenge.host) is providing a potentially unsafe certificate."
         )
-        .multilineTextAlignment(.center)
-        .font(.subheadline)
-        .foregroundColor(.secondary)
     }
     
     func body(content: Content) -> some View {
@@ -60,42 +55,28 @@ struct TrustHostViewModifier: ViewModifier {
                 // it doesn't show.
                 isPresented = true
             }
-            .sheet(isPresented: $isPresented) {
-                VStack(alignment: .center) {
-                    title
-                        .padding(.vertical)
-                    message
-                        .padding(.bottom)
-                    HStack {
-                        Spacer()
-                        Button(role: .cancel) {
-                            isPresented = false
-                            challenge.resume(with: .cancel)
-                        } label: {
-                            Text(String.cancel)
-                                .padding(.horizontal)
-                        }
-                        .buttonStyle(.bordered)
-                        Spacer()
-                        Button(role: .destructive) {
-                            isPresented = false
-                            challenge.resume(with: .continueWithCredential(.serverTrust))
-                        } label: {
-                            Text(
-                                "Allow",
-                                bundle: .toolkitModule,
-                                comment: "A button indicating the user accepts a potentially dangerous action."
-                            )
-                            .padding(.horizontal)
-                        }
-                        .buttonStyle(.borderedProminent)
-                        Spacer()
+            .alert(
+                title,
+                isPresented: $isPresented,
+                actions: {
+                    Button(role: .destructive) {
+                        isPresented = false
+                        challenge.resume(with: .continueWithCredential(.serverTrust))
+                    } label: {
+                        Text(
+                            "Allow",
+                            bundle: .toolkitModule,
+                            comment: "A button indicating the user accepts a potentially dangerous action."
+                        )
                     }
-                    Spacer()
-                }
-                .padding()
-                .presentationDetents([.medium])
-                .interactiveDismissDisabled()
-            }
+                    Button(role: .cancel) {
+                        isPresented = false
+                        challenge.resume(with: .cancel)
+                    } label: {
+                        Text(String.cancel)
+                    }
+                }, message: {
+                    Text(message)
+                })
     }
 }
--- a/Sources/ArcGISToolkit/Components/Authentication/TrustHostViewModifier.swift
+++ b/Sources/ArcGISToolkit/Components/Authentication/TrustHostViewModifier.swift
@@ -30,20 +30,25 @@ struct TrustHostViewModifier: ViewModifier {
     /// A Boolean value indicating whether or not the prompt is displayed.
     @State var isPresented: Bool = false
     
-    var title: String {
-        String(
-            localized: "Certificate Trust Warning",
+    var title: some View {
+        Text(
+            "Certificate Trust Warning",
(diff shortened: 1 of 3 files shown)


Sources/ArcGISToolkit/Utility/CredentialInputView.swift
```diff
@@ -210,65 +210,58 @@ struct CredentialInputSheetView: View {
...
-                        .buttonStyle(.bordered)
-                        Spacer()
-                        Button {
-                            isPresented.wrappedValue = false
-                            continueAction.handler(username, password)
-                        } label: {
-                            Text(continueAction.title)
-                                .padding(.horizontal)
+                        .padding(.vertical)
+                        Form {
```
zkline101, line 227: Using a form provides a range of benefits here and it also allows this UI to follow this visionOS design guideline: 
Screenshot 2024-09-18 at 5 16 48 PM

Sources/ArcGISToolkit/Utility/CredentialInputView.swift
```diff
@@ -210,65 +210,58 @@ struct CredentialInputSheetView: View {
...
             }
-        }
-        .onAppear {
-            // Set initial focus of text field.
-            switch fields {
-            case .usernamePassword:
-                usernameFieldIsFocused = true
-            case .password:
-                passwordFieldIsFocused = true
+            .toolbar {
```
zkline101, line 239: The cancel and continue buttons were both oddly placed at the bottom which doesn't follow this guideline:
Screenshot 2024-09-18 at 5 19 02 PM

zkline101:
Please wait to review, I see a few more edits I want to make.

Sources/ArcGISToolkit/Components/Authentication/TrustHostViewModifier.swift
```diff
@@ -60,42 +55,28 @@ struct TrustHostViewModifier: ViewModifier {
...
-                            Text(
-                                "Allow",
-                                bundle: .toolkitModule,
-                                comment: "A button indicating the user accepts a potentially dangerous action."
-                            )
-                            .padding(.horizontal)
-                        }
-                        .buttonStyle(.borderedProminent)
-                        Spacer()
+            .alert(
```
zkline101, line 58: I was wondering why this trust host prompt looked a bit off. It is because it makes much more sense as an alert rather than a sheet. I saw an Apple alert on my phone that looked just like this. Here is what this looks like now:
Simulator Screenshot - iPhone 16 Pro Max - 2024-09-19 at 09 48 23

zkline101:
@njarecha @dfeinzimer This is now ready for review, thank you.

njarecha:
@zkline101 Please revert the changes of using alerts. There are issues with it. Please see [PR](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/381#issue-1831967874) where we made the changes from `alert` to `sheet`.

Sources/ArcGISToolkit/Components/Authentication/TrustHostViewModifier.swift
yo1995: I think it is related to https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/381 where multiple alerts being presented at the same time will cause the following alerts to disappear.

zkline101:
Dang, alright. The only changes I have then for the trust host prompt and certificate picker is to have the buttons on the top of the sheet.

zkline101:
Thanks @njarecha I reverted the alert changes, this is open for review.

Sources/ArcGISToolkit/Utility/CredentialInputView.swift
```diff
@@ -210,65 +210,58 @@ struct CredentialInputSheetView: View {
...
+                        .padding(.vertical)
+                        Form {
+                            switch fields {
+                            case .password:
+                                passwordTextField
+                            case .usernamePassword:
+                                usernameTextField
+                                passwordTextField
+                            }
                         }
```
dfeinzimer, line 235:
```suggestion
                        }
                        .scrollContentBackground(.hidden)
```

The background difference between the header is a bit jarring, thoughts?

| Before | After |
|---|---|
|  |  |

zkline101: I noticed that too but I believe I left it there because of some default behavior on visionOS, let me check.

zkline101: Nope looks good.

Sources/ArcGISToolkit/Utility/CredentialInputView.swift
rolson: Does this need to be part of the visionOS work? If not, then I'd prefer we do not do this at this time. We have a UI/UX designer starting in October and one of the things we will ask her to do is to go through this UI and make suggestions for improvements.

rolson:
> @zkline101 If we are making changes to the credential input view then I would like to see it like this,

I would prefer to hold off on making UI edits until we have suggestions from the UI designer who is starting soon.

Sources/ArcGISToolkit/Utility/CredentialInputView.swift
zkline101: > Does this need to be part of the visionOS work?

I don't think this **needs** to be. But putting the text fields into a form would need to because that adds the default background color for each text field which would be required for visionOS as per [this](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/878#discussion_r1765903581) comment. And I don't think that would change much on iOS.

rolson: The form change can be ok. But I don't want to move the buttons or do anything else

zkline101:
Ok I moved the buttons back to the original location but kept the `Form`. This is ready for review.

iOS:
Simulator Screenshot - iPhone 16 Pro Max - 2024-09-23 at 12 55 28

visionOS:
Simulator Screenshot - AVP Model 2 - 2024-09-23 at 12 53 49

dfeinzimer (dismissed):

Sources/ArcGISToolkit/Utility/CredentialInputView.swift
```diff
@@ -223,44 +223,48 @@ struct CredentialInputSheetView: View {
...
-                        .disabled(!isContinueEnabled)
-                        Spacer()
-                    }
-                    Spacer()
+                    .scrollContentBackground(.hidden)
                 }
-                .padding()
-                Spacer()
             }
+            .padding()
```
njarecha, line 267: On iOS, text field's background is getting merged with view's background so it looks odd. Don't see the same for visionOS. However, changing this makes it look much better. What is the impact of this change in visionOS? 

```suggestion
            .padding()
            .background(.thinMaterial)
```

Simulator Screenshot - iPhone 16 Pro Max - 2024-09-23 at 13 56 22

njarecha (commented):
Just one suggestion. Also, noticed that `GeometryReader { proxy in` is not used anywhere. Shall we remove it?

Sources/ArcGISToolkit/Utility/CredentialInputView.swift
zkline101: > On iOS, text field's background is getting merged with view's background so it looks odd. 

It was like that before these changes. 

> Don't see the same for visionOS.

That is because using a `Form` it adds that background automatically for visionOS to follow [this](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/878#discussion_r1765903581) guideline.

> What is the impact of this change in visionOS?

Since the `Form` already gives the text fields a background there wouldn't be a need to change that.

zkline101:
> Just one suggestion. Also, noticed that GeometryReader { proxy in is not used anywhere. Shall we remove it?

Yup looks like that can go. It wasn't used in the original implementation either.

njarecha (approved):
Looks good

zkline101:
Thanks everyone for the feedback.

PR #879, opened 2024-09-19
Offline: Refactor UI

des12437:
Closes `swift/6039`

- Renames "Offline Maps" title to "Map Areas"
- Removes "Preplanned" section header
- Rewords empty state view


yo1995 (approved):

philium (commented):
You might consider putting the "no map areas" text in the center of the view (not in a list), à la Mail:

Screenshot 2024-09-19 at 4 11 53 PM

des12437:
> You might consider putting the "no map areas" text in the center of the view (not in a list), à la Mail:

For consistency with the error view it's better to have the empty state view in the same style. If the empty state view is centered in an overlay it will need to be moved later on when the design to add an on-demand map area is implemented.

PR #890, opened 2024-09-24
UI edits for `FloatingPanel`

zkline101:
Swift #6015

Here are different screenshots of the floating panel being used:
Simulator Screenshot - Apple Vision Pro - 2024-09-24 at 11 53 14
Simulator Screenshot - Apple Vision Pro - 2024-09-24 at 11 53 56

### DIFF
--- a/Sources/ArcGISToolkit/Components/FloatingPanel/FloatingPanel.swift
+++ b/Sources/ArcGISToolkit/Components/FloatingPanel/FloatingPanel.swift
@@ -212,6 +212,7 @@ private struct Handle: View {
         RoundedRectangle(cornerRadius: 4.0)
             .foregroundColor(color)
             .frame(width: 100, height: 8.0)
+            .hoverEffect()
     }
 }
 
--- a/Sources/ArcGISToolkit/Components/FloatingPanel/FloatingPanel.swift
+++ b/Sources/ArcGISToolkit/Components/FloatingPanel/FloatingPanel.swift
@@ -25,7 +25,7 @@ struct FloatingPanel<Content>: View where Content: View {
     /// attribution bar.
     let attributionBarHeight: CGFloat
     /// The background color of the floating panel.
-    let backgroundColor: Color
+    let backgroundColor: Color?
     /// A binding to the currently selected detent.
     @Binding var selectedDetent: FloatingPanelDetent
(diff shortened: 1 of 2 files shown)


Sources/ArcGISToolkit/Components/FloatingPanel/FloatingPanel.swift
```diff
@@ -73,7 +73,12 @@ struct FloatingPanel<Content>: View where Content: View {
             }
             // Set frame width to infinity to prevent horizontal shrink on dismissal.
             .frame(maxWidth: .infinity)
+#if os(visionOS)
+            .background(.regularMaterial)
+            .compositingGroup()
```
zkline101, line 78: Without using the `compositingGroup()` modifier you end up with something like this:
Material_Showing

I believe adding this modifier lets the material layer get added **before** the clipping occurs.

Sources/ArcGISToolkit/Components/FloatingPanel/FloatingPanelModifier.swift
```diff
@@ -28,6 +28,9 @@ public extension View {
     ///   - content: A closure that returns the content of the floating panel.
     /// - Returns: A dynamic view with a presentation style similar to that of a sheet in compact
     /// environments and a popover otherwise.
+    @available(iOS 16.0, *)
+    @available(macCatalyst 16.0, *)
+    @available(visionOS, unavailable, message: "Use 'floatingPanel(attributionBarHeight:selectedDetent:horizontalAlignment:isPresented:maxWidth:_:)' instead.")
```
zkline101, line 33: Since using colors isn't really encouraged in visionOS and the default color for this modifier is `systemBackground` which doesn't exist in visionOS it doesn't seem to make sense to offer this view modifier with a background color parameter for visionOS. So instead I added a very similar view modifier below that is available on visionOS and it doesn't include the `backgroundColor` parameter.
Screenshot 2024-09-24 at 11 56 42 AM

dfeinzimer (approved):
Looks great, thanks!

zkline101:
Thanks @dfeinzimer. I just pushed one last commit where I updated the doc.

dfeinzimer (approved):

philium (approved):

PR #892, opened 2024-09-25
Offline: UI improvements

des12437:
Related to `swift/6048`

- Rephrases empty state view text
- Uses `ContentUnavailableView` when map is offline disabled
- Removes HTML tags from map area description
- Animates open button

### DIFF
--- a/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
@@ -102,9 +102,10 @@ public struct OfflineMapAreasView: View {
         VStack(alignment: .center) {
             Text("No map areas")
                 .bold()
-            Text("You don't have any map areas yet.")
+            Text("There are no map areas defined for this web map.")
                 .font(.subheadline)
                 .foregroundStyle(.secondary)
+                .multilineTextAlignment(.center)
         }
         .frame(maxWidth: .infinity)
     }
--- a/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
@@ -118,20 +118,22 @@ public struct OfflineMapAreasView: View {
         .frame(maxWidth: .infinity)
     }
     
-    private var offlineDisabledView: some View {
+    @ViewBuilder private var offlineDisabledView: some View {
+        let labelText = Text("Offline Disabled")
+        let descriptionText = Text("Please ensure the web map is offline enabled.")
         if #available(iOS 17, *) {
-            return ContentUnavailableView {
-                Text("Offline Disabled")
+            ContentUnavailableView {
+                labelText
                     .bold()
             } description: {
-                Text("Please ensure the web map is offline enabled.")
+                descriptionText
             }
         } else {
-            return VStack(alignment: .center) {
-                Text("Offline Disabled")
+            VStack(alignment: .center) {
+                labelText
                     .bold()
                     .font(.title2)
-                Text("Please ensure the web map is offline enabled.")
+                descriptionText
                     .font(.subheadline)
                     .foregroundStyle(.secondary)
             }
--- a/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
@@ -31,6 +31,11 @@ public struct OfflineMapAreasView: View {
     /// The currently selected map.
     @Binding private var selectedMap: Map?
     
+    /// A Boolean value indicating whether the web map is offline disabled.
+    private var mapIsOfflineDisabled: Bool {
+        onlineMap.loadStatus == .loaded && onlineMap.offlineSettings == nil
+    }
+    
     /// Creates a view with a given web map.
     /// - Parameters:
     ///   - online: The web map to be taken offline.
@@ -45,9 +50,7 @@ public struct OfflineMapAreasView: View {
         NavigationStack {
             Form {
                 Section {
-                    if onlineMap.loadStatus == .loaded && onlineMap.offlineSettings == nil {
-                        offlineDisabledView
-                    } else {
+                    if !mapIsOfflineDisabled {
                         preplannedMapAreaViews
                     }
                 }
@@ -65,6 +68,11 @@ public struct OfflineMapAreasView: View {
             }
             .navigationTitle("Map Areas")
             .navigationBarTitleDisplayMode(.inline)
+            .overlay {
+                if mapIsOfflineDisabled {
+                    offlineDisabledView
+                }
+            }
         }
         .refreshable {
             await mapViewModel.makePreplannedOfflineMapModels()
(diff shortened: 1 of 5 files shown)


Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -102,22 +110,33 @@ public struct OfflineMapAreasView: View {
...
-            Text("You don't have any map areas yet.")
+            Text("There are no map areas defined for this web map.")
                 .font(.subheadline)
                 .foregroundStyle(.secondary)
+                .multilineTextAlignment(.center)
         }
         .frame(maxWidth: .infinity)
     }
     
     private var offlineDisabledView: some View {
```
philium, line 121: Please remove the explicit return statements and then annotate with `ViewBuilder`. Also, consolidate the strings.

```swift
@ViewBuilder private var offlineDisabledView: some View {
    let labelText = Text("Offline Disabled")
    let descriptionText = Text("Please ensure the web map is offline enabled.")
    if #available(iOS 17, *) {
        ContentUnavailableView {
            labelText
                .bold()
        } description: {
            descriptionText
        }
    } else {
        VStack(alignment: .center) {
            labelText
                .bold()
                .font(.title2)
            descriptionText
                .font(.subheadline)
                .foregroundStyle(.secondary)
        }
        .frame(maxWidth: .infinity)
    }
}
```

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -318,7 +320,8 @@ extension PreplannedMapArea: PreplannedMapAreaProtocol {
     }
     
     var description: String {
-        portalItem.description
+        // Remove HTML tags from description.
+        portalItem.description.replacingOccurrences(of: "<[^>]+>", with: "", options: .regularExpression)
```
philium, line 324: Please use a Swift regex expression instead.

```suggestion
        portalItem.description.replacing(/<[^>]+>/, with: "")
```

Note: This suggestion will requiring enabling the `BareSlashRegexLiterals` upcoming feature:

```swift
.enableUpcomingFeature("BareSlashRegexLiterals")
```

Which can then be removed as part of the work to move to Swift 6 as that feature is enabled by default.

Tests/ArcGISToolkitTests/PreplannedMapModelTests.swift
```diff
@@ -374,6 +374,34 @@ class PreplannedMapModelTests: XCTestCase {
...
+            // User notifications in unit tests are not supported, must pass false here
+            // or the test process will crash.
+            showsUserNotificationOnCompletion: false
+        )
+        
+        // Verify description does not contain HTML tags.
+        XCTAssertEqual(model.preplannedMapArea.description,
+        """
+        A map that contains stormwater network within Naperville, IL, USA with cartography designed for web and mobile devices. This is a demo map to demonstrate offline capabilities with ArcGIS Runtime and is based on ArcGIS Solutions for Stormwater.
+        """)
```
philium, line 403: This string literal only has a single line, so a multiline string literal shouldn't be used. Also, here is how we would format/indent this assertion:

```suggestion
        XCTAssertEqual(
            model.preplannedMapArea.description,
            "A map that contains stormwater network within Naperville, IL, USA with cartography designed for web and mobile devices. This is a demo map to demonstrate offline capabilities with ArcGIS Runtime and is based on ArcGIS Solutions for Stormwater."
        )
```

yo1995 (commented):
LGTM with Phil suggested changes

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -130,7 +130,9 @@ class PreplannedMapModel: ObservableObject, Identifiable {
     func updateDownloadStatus(for downloadResult: Result<DownloadPreplannedOfflineMapResult, any Error>) {
         switch downloadResult {
         case .success:
-            status = .downloaded
+            withAnimation(.easeInOut) {
```
philium, line 133: This really shouldn't be done by the model as `withAnimation` is a SwiftUI function. I just noticed that this file imports SwiftUI, which is a huge code smell. The model should know nothing about the view; the view itself should perform the animation in response to a change in the model's state.

des12437: Yes using `withAnimation` in the model was a workaround. The `animation<V>(_ animation: Animation?, value: V)` view modifier requires a value to change to animate, and `model.status` is not equatable. I tried using `model.status.isDownloaded` but that does not trigger an animation. Do you suggest making `model.status` equatable or another option?

philium: There is another option: `View.onReceive(_:perform:)`. First, revert these changes and change `import SwiftUI` to `import Combine`. Then, in the list item view, add a new download state type and state property:

```swift
private enum DownloadState {
    case notDownloaded, downloading, downloaded
    
    init(_ state: PreplannedMapModel.Status) {
        self = switch state {
        case .downloaded: .downloaded
        case .downloading: .downloading
        default: .notDownloaded
        }
    }
}

@State private var downloadState: DownloadState = .notDownloaded
```

Then, add a couple modifiers to set the initial download state and update the download state in response to the model status changes:

```swift
.onAppear {
    downloadState = .init(model.status)
}
.onReceive(model.$status) { status in
    let downloadState = DownloadState(status)
    withAnimation(
        downloadState == .downloaded ? .easeInOut : nil
    ) {
        self.downloadState = downloadState
    }
}
```

Lastly, update the code in `downloadButton` to use the new download state.

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
```diff
@@ -27,6 +27,22 @@ struct PreplannedListItemView: View {
...
+    private enum DownloadState {
+        case notDownloaded, downloading, downloaded
+        
+        init(_ state: PreplannedMapModel.Status) {
+            self = switch state {
+            case .downloaded: .downloaded
+            case .downloading: .downloading
+            default: .notDownloaded
+            }
+        }
```
philium, line 40: I just showed this here for simplicity, but normally we would keep an enum/struct declaration clean of convenience members and instead add them in private extensions towards the bottom of the file. Consider doing that here to keep the view itself as simple and readable as possible.

philium (approved):

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
philium: You misunderstood what I meant. I was referring to this initializer (the one that I selected) only and not the entire enum. The enum itself should be where it was, but this initializer makes sense in an extension:

```swift
private extension PreplannedListItemView.DownloadState {
    init(_ state: PreplannedMapModel.Status) {
        ...
    }
}
```

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
```diff
@@ -27,6 +27,14 @@ struct PreplannedListItemView: View {
     /// A Boolean value indicating whether the metadata view is presented.
     @State private var metadataViewIsPresented = false
     
+    /// The download state of the preplanned map model.
+    enum DownloadState {
```
philium, line 31: This should still be private.

```suggestion
    fileprivate enum DownloadState {
```

philium (approved):

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -14,7 +14,7 @@
 
 import ArcGIS
 import OSLog
-import SwiftUI
+import Combine
```
yo1995, line 17: When I comment out this import, it still compiles.

```suggestion
```

It seems the import of `OSLog` implicitly introduced `Foundation`. I'm not sure if we should import Foundation here as well.

yo1995 (approved):

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
philium: Oh! Good catch. `OSLog` should not be imported; it has been superseded by `os` (which does not export Foundation, by the way). So the imports for this file should be as follows:

```swift
import ArcGIS
import Combine
import Foundation

internal import os
```

philium (approved):

yo1995 (approved):

des12437:
Thank you both!

PR #895, opened 2024-10-01
Offline: Move user notification authorization to app level

des12437:
- Moves user notification authorization logic from the offline component and places it at the app level

### DIFF
--- a/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
@@ -40,12 +40,6 @@ extension OfflineMapAreasView {
             portalItemID = map.item?.id
         }
         
-        /// Requests authorization to show notifications.
-        nonisolated func requestUserNotificationAuthorization() async {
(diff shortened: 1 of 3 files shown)


Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
```diff
@@ -15,7 +15,6 @@
 import ArcGIS
 import Combine
 import Foundation
-import UserNotifications
```
philium, line 18: Should this be imported in `OfflineMapAreasExampleView.swift`?

des12437: The notification authorization works without this import. I think the `SwiftUI` import has support for `UserNotifications`, but I'm not sure how to confirm that.

philium: I would suggest adding it anyway. It helps make the file's dependencies more explicit, which can aid readability.

yo1995 (approved):

philium (approved):

PR #897, opened 2024-10-04
`BarcodeScannerFormInput` refinements

dfeinzimer:
Ref Apollo 898


mhdostal (approved):
This looks great. I tested with both a single QR code and both a barcode and QR code and it worked great. I like the "white" torch when the flashlight is on, that works.

One thing: In the Attachment Camera display (when grabbing a new attachment from the Camera), the "Cancel" button is white with no background and is in the bottom-trailing part of the view. In the Barcode scanner, it is in the top-trailing position with a blue background. I think I like the Cancel button with no background in bottom-trailing, but not enough for it to be changed at this point. We can get more feedback, or have Peggy take a look at some point.

PR #900, opened 2024-10-08
Adds Barcode test case 11.1

mhdostal:
Adds Test Case 11.1. Needed to update the `InputFooter` to take into account the Barcode Input when retrieving the `lengthRange` property in order to correctly get the range.

I also added an `accessibilityIdentifier` to the scan button in the `TextInput` struct.

### DIFF
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/Other/InputFooter.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/Other/InputFooter.swift
@@ -217,11 +217,14 @@ extension InputFooter {
     var lengthRange: ClosedRange<Int>? {
         if let input = element.input as? TextAreaFormInput {
             return input.minLength...input.maxLength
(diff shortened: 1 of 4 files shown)


mhdostal:
cc: @kaushikrw - I add to adjust the logic in our `InputFooter` to take into account the barcode input type when getting the length range for the input. This allowed the validation message to be "Maximum 50 characters" (which the design says it should be) instead of "Maximum character length exceeded". Kotlin might need the same, if it handles validation the same way.

mhdostal:
@dfeinzimer Should `InputFooter.isShowingCharacterIndicator` be updated to include a check for `BarcodeScannerFormInput`? It currently checks for `TextAreaFormInput` and `TextBoxFormInput`?

mhdostal:
> I can go either way. It looks like we didn't specify a "focused" spec in the design but since we allow manual editing I could see how it might be useful.

Ok, thanks. I'm OK leaving it as-is. I don't think it will be very common for a "barcode scanner" to specify a number of characters, as it's all supposed to be automatic.

PR #906, opened 2024-10-12
Add visionOS as a destination to the Authentication Example

zkline101:
Swift #6014

Here is a screenshot of what it looks like in visionOS:
Simulator Screenshot - Apple Vision Pro - 2024-10-11 at 16 58 12

I highly recommend running the app and looking at the changes.


AuthenticationExample/AuthenticationExample/HomeView.swift
```diff
@@ -25,22 +25,24 @@ struct HomeView: View {
     
     var body: some View {
         if let portal = portal {
-            NavigationStack {
+            NavigationSplitView {
```
zkline101, line 28: A split view makes much more sense than a stack since a split view will automatically choose the number of columns to show. This means in visionOS and Mac Catalyst there will be two columns (which is expected) and narrow-size classes like an iPhone will collapse into a single column. This is the same thing we do in the Samples app.

AuthenticationExample/AuthenticationExample/SignInView.swift
```diff
@@ -107,7 +107,9 @@ struct SignInView: View {
             signIn()
         } label: {
             Text(signInButtonText)
+#if !os(visionOS)
                 .frame(maxWidth: .infinity)
+#endif
```
zkline101, line 112: It seemed odd having this button's width be so large in a 2D window since a window itself can be quite large.

AuthenticationExample/AuthenticationExample/ProfileView.swift
```diff
@@ -28,15 +28,28 @@ struct ProfileView: View {
...
+                Spacer()
+                signOutButton
+            }
+            .padding()
+            .toolbar {
+                ToolbarItem(placement: .topBarTrailing) {
+                    Button("Done") {
+                        dismiss()
+                    }
+                }
```
zkline101, line 50: All I did here was add a done button since on mac catalyst and visionOS there is no way to exit the profile view.

dfeinzimer (approved):

zkline101:
My rebase with v.next got missed up, give me a second to fix.

zkline101:
Alright, I fixed the problems and have a PR [up](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/911) that fixes the build errors. 

The code in this PR hasn't changed and is ready for review @dfeinzimer @njarecha.

dfeinzimer (approved):

njarecha (changes requested):
Looks good, just couple of things,

1. Can you please fix this warning?
Screenshot 2024-10-16 at 1 22 50 PM

2. The text looks very small. Is it possible to increase the size?
Screenshot 2024-10-16 at 1 27 26 PM

zkline101:
> Can you please fix this warning?

That is unrelated to these changes so that can be done in a separate PR.


>  The text looks very small. Is it possible to increase the size?

Sure I can increase it. I left it as is because it was still easy to hover over with the eye detection and select with my fingers.

njarecha (approved):

dfeinzimer (approved):

zkline101:
Thank you both

PR #922, opened 2024-10-23
`FeatureFormView` / `ComboBoxInput` - Show unsupported values

dfeinzimer:
Apollo 839

https://github.com/user-attachments/assets/4ed991a6-e349-4c54-8a0b-96118b10880f

### DIFF
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/ComboBoxInput.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/ComboBoxInput.swift
@@ -114,10 +114,6 @@ struct ComboBoxInput: View {
         .sheet(isPresented: $isPresented) {
             makePicker()
         }
-        .onTapGesture {
-            model.focusedElement = element
-            isPresented = true
-        }
         .onAppear {
             if let currentValue = element.codedValues.first(where: {
                 $0.name == element.formattedValue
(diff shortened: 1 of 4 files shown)


dfeinzimer:
Note that in its current form, this doesn't show the current unsupported type in the options list like Kotlin. Thoughts on this?

<p align="center">
  
</p>

dfeinzimer:
Used map `99ef6889b8ab42eb8cbd3b9672c7b592` / "Extrinsic_Value" to test fallback behavior successfully.

<p align="center">
  
</p>

mhdostal:
> Note that in its current form, this doesn't show the current "Unsupported Type" in the options like Kotlin. Thoughts on this?

Is there a reason we're not following what Kotlin does?

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/ComboBoxInput.swift
```diff
@@ -25,9 +25,6 @@ struct ComboBoxInput: View {
     /// The phrase to use when filtering by coded value name.
     @State private var filterPhrase = ""
     
-    /// The formatted version of the element's current value.
-    @State private var formattedValue = ""
-
```
dfeinzimer, line 30: `formattedValue` was unused

dfeinzimer:
> Is there a reason we're not following what Kotlin does?

My original thinking was that it was overcomplicating the sheet content but I can see now how it helps to provide a bit of context around the value being invalid. 

I've moved to match Kotlin.



https://github.com/user-attachments/assets/e86704b0-cbc1-4331-b2dd-c91734858dca


https://github.com/user-attachments/assets/0bbba34e-6839-4c2c-9fff-fcc6244bd1e3

mhdostal (approved):

dfeinzimer:
Verified all Feature Form tests pass on iOS 17 and 18 simulators

PR #937, opened 2024-10-25
[Update] Table top `translationFactor` doc

CalebRas:
This PR updates the `TableTopSceneView.init(anchorPoint:translationFactor:clippingDistance:sceneView:)` doc with more info that can be found in the [Implement tabletop AR](https://developers.arcgis.com/swift/scenes-3d/display-scenes-in-augmented-reality/#implement-tabletop-ar) conceptual doc.

## Linked Issue(s)

- `swift/issues/6175`

### DIFF
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/TableTopSceneView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/TableTopSceneView.swift
@@ -53,8 +53,13 @@ public struct TableTopSceneView: View {
     /// Creates a table top scene view.
     /// - Parameters:
     ///   - anchorPoint: The location point of the ArcGIS Scene that is anchored on a physical surface.
-    ///   - translationFactor: The translation factor that defines how much the scene view translates
-    ///   as the device moves.
+    ///   - translationFactor: The translation factor that defines how much the scene view
+    ///   translates as the device moves. This value can be determined by dividing the virtual
+    ///   content width by the desired physical content width (translation factor = virtual content
+    ///   width / desired physical content width). The virtual content width is the real-world size
+    ///   of the scene content, and the desired physical content width is the physical tabletop
+    ///   width. The virtual content width is determined by the clipping distance in meters around
+    ///   the camera.
     ///   - clippingDistance: Determines the clipping distance in meters around the camera. A value
     ///   of `nil` means that no data will be clipped.
     ///   - sceneView: A closure that builds the scene view to be overlayed on top of the
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/TableTopSceneView.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/TableTopSceneView.swift
@@ -58,8 +58,8 @@ public struct TableTopSceneView: View {
     ///   content width by the desired physical content width (translation factor = virtual content
     ///   width / desired physical content width). The virtual content width is the real-world size
     ///   of the scene content, and the desired physical content width is the physical tabletop
-    ///   width. The virtual content width is determined by the clipping distance in meters around
-    ///   the camera.
+    ///   width; both measurements should be in meters. The virtual content width is determined 
+    ///   by the clipping distance in meters around the camera.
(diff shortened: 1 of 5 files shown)


CalebRas:
@yo1995 You mentioned also adding this text to step 1 of the tutorial, but it made it rather cluttered since there can only be one caption paragraph. If you are still in favor of adding the tutorial, I think it would be worth breaking step 1 up. Let me know what you think!

yo1995:
I think adding it to the API reference is more important, and I like the changes in this PR so far.

I'd like to know Destiny's opinion on updating the tutorial. I prefer a slightly colloquial explanation (maybe a new step) in the tutorial, giving a numeric example of how to set the translation factor.

Sources/ArcGISToolkit/Components/Augmented Reality/TableTopSceneView.swift
```diff
@@ -53,8 +53,13 @@ public struct TableTopSceneView: View {
...
     ///   - anchorPoint: The location point of the ArcGIS Scene that is anchored on a physical surface.
-    ///   - translationFactor: The translation factor that defines how much the scene view translates
-    ///   as the device moves.
+    ///   - translationFactor: The translation factor that defines how much the scene view
+    ///   translates as the device moves. This value can be determined by dividing the virtual
+    ///   content width by the desired physical content width (translation factor = virtual content
+    ///   width / desired physical content width). The virtual content width is the real-world size
+    ///   of the scene content, and the desired physical content width is the physical tabletop
+    ///   width. The virtual content width is determined by the clipping distance in meters around
+    ///   the camera.
```
des12437, line 62:
```suggestion
    ///   - translationFactor: The translation factor that defines how much the scene view
    ///   translates as the device moves. This value can be determined by dividing the virtual
    ///   content width by the desired physical content width (translation factor = virtual content
    ///   width / desired physical content width). The virtual content width is the real-world size
    ///   of the scene content, and the desired physical content width is the physical tabletop
    ///   width; both measurements should be in meters. The virtual content width is determined 
    ///   by the clipping distance in meters around the camera.
```
The units should be specified.

des12437:
> I think adding it to the API reference is more important, and I like the changes in this PR so far.

Agreed.

> I'd like to know Destiny's opinion on updating the tutorial. I prefer a slightly colloquial explanation (maybe a new step) in the tutorial, giving a numeric example of how to set the translation factor.

Updating the tutorial with a new step would be beneficial. The tutorial currently does not follow this formula to determine the translation factor so that would need to be updated. The image example shows a scene on a table (< 1 meter wide) with a clipping distance of 400, so the translation factor should be 400 meters / 1 meter = 400 translation factor.

CalebRas:
I tried to fit it into step 1 following a suggestion from Ting and also update the tutorial to use a `400` translation factor. Let me know what you think.

yo1995 (approved):

Sources/ArcGISToolkit/Documentation.docc/Tutorials/TableTopSceneViewTutorial.tutorial
```diff
@@ -16,8 +16,10 @@
                 
                 The anchor point is the location point of the ArcGIS Scene that is anchored on a 
                 physical surface. The translation factor determines how many meters the scene view
-                translates as the device moves in the AR experience. The clipping distance is the 
-                distance in meters that the ArcGIS Scene data will be clipped to.
+                translates as the device moves in the AR experience. This value can be determined by
+                dividing the virtual content width by the desired physical content width. In this
+                case, 400 m / 1 m = 400 translation factor. The clipping distance is the distance in
+                meters that the ArcGIS Scene data will be clipped to.
```
des12437, line 22:
```suggestion
                translates as the device moves in the AR experience. This value can be determined by
                dividing the virtual content width by the desired physical content width. In this
                case, 400 m ArcGIS Scene width / 1 m table top width = 400 translation factor. The 
                clipping distance is the distance in meters that the ArcGIS Scene data will be clipped to.
```

des12437 (approved):

yo1995 (approved):

CalebRas:
Thank you both!

PR #940, opened 2024-10-29
Add visionOS support to `PopupView`

zkline101:
Swift #6156

I tried multiple pop up data sets with these changes to make sure I covered all the bases but I would still recommend looking at these changes yourself. You can test this by adding visionOS as a destination to the Examples app and just comment out the examples that aren't available on visionOS or mark the examples as `@available(visionOS, unavailable)`.

Here are some screenshots:
Simulator Screenshot - Apple Vision Pro - 2024-10-29 at 13 35 45
Simulator Screenshot - Apple Vision Pro - 2024-10-29 at 13 32 10
Simulator Screenshot - Apple Vision Pro - 2024-10-29 at 09 02 59

### DIFF
--- a/Sources/ArcGISToolkit/Components/Popups/FieldsPopupElementView.swift
+++ b/Sources/ArcGISToolkit/Components/Popups/FieldsPopupElementView.swift
@@ -72,8 +72,9 @@ struct FieldsPopupElementView: View {
         let formattedValue: String
         
         var body: some View {
-            if formattedValue.lowercased().starts(with: "http") {
-                Link("View", destination: URL(string: formattedValue)!)
+            if formattedValue.lowercased().starts(with: "http"),
+               let url = URL(string: formattedValue) {
+                Link("View", destination: url)
                     .buttonStyle(.bordered)
             } else {
                 Text(formattedValue)
--- a/Sources/ArcGISToolkit/Components/Popups/FieldsPopupElementView.swift
+++ b/Sources/ArcGISToolkit/Components/Popups/FieldsPopupElementView.swift
@@ -73,7 +73,8 @@ struct FieldsPopupElementView: View {
         
         var body: some View {
             if formattedValue.lowercased().starts(with: "http") {
-                Text(.init("[View](\(formattedValue))"))
+                Link("View", destination: URL(string: formattedValue)!)
+                    .buttonStyle(.bordered)
             } else {
                 Text(formattedValue)
             }
--- a/Sources/ArcGISToolkit/Components/Popups/PopupView.swift
+++ b/Sources/ArcGISToolkit/Components/Popups/PopupView.swift
@@ -85,7 +85,7 @@ public struct PopupView: View {
                 }
                 Spacer()
                 if showCloseButton {
-                    Button("Close", systemImage: "xmark") {
+                    Button(String.close, systemImage: "xmark") {
                         isPresented?.wrappedValue = false
                     }
                     .labelStyle(.iconOnly)
--- a/Sources/ArcGISToolkit/Components/Popups/PopupView.swift
+++ b/Sources/ArcGISToolkit/Components/Popups/PopupView.swift
@@ -51,7 +51,6 @@ import ArcGIS
 /// and refer to
 /// [PopupExampleView.swift](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/main/Examples/Examples/PopupExampleView.swift)
 /// in the project. To learn more about using the `PopupView` see the <doc:PopupViewTutorial>.
-@available(visionOS, unavailable)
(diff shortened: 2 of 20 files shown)


Sources/ArcGISToolkit/Components/Popups/FieldsPopupElementView.swift
```diff
@@ -73,7 +73,8 @@ struct FieldsPopupElementView: View {
         
         var body: some View {
             if formattedValue.lowercased().starts(with: "http") {
-                Text(.init("[View](\(formattedValue))"))
+                Link("View", destination: URL(string: formattedValue)!)
```
philium, line 76: Instead of checking the start of the string and then force casting, perhaps it would be better to conditionally unwrap the created URL:

```swift
if let url = URL(string: formattedValue) {
    Link("View", destination: url)
} else {
    ...
}
```

Sources/ArcGISToolkit/Components/Popups/PopupView.swift
```diff
@@ -89,11 +88,19 @@ public struct PopupView: View {
...
                         isPresented?.wrappedValue = false
                     }, label: {
-                        Image(systemName: "xmark.circle")
+                        let imageName: String
+#if !os(visionOS)
+                        imageName = "xmark.circle"
+#else
+                        imageName = "xmark"
+#endif
+                        return Image(systemName: imageName)
```
philium, line 97:
```suggestion
                        Image(systemName: "xmark")
#if !os(visionOS)
                            .symbolVariant(.circle)
#endif
```

Sources/ArcGISToolkit/Components/Popups/FieldsPopupElementView.swift
philium: Or perhaps both the prefix and the conditional unwrap, in case the URL initializer might treat some non-URL strings as URLs.

Sources/ArcGISToolkit/Components/Popups/PopupView.swift
philium: This modifier could actually be moved down a few more lines into the next conditional compilation block. It doesn't need to go on the image view directly.

Sources/ArcGISToolkit/Components/Popups/FieldsPopupElementView.swift
zkline101: > in case the URL initializer might treat some non-URL strings as URLs.

Yup, that is what happens. So I'll follow the second suggestion.

Sources/ArcGISToolkit/Components/Popups/PopupView.swift
zkline101: It looks like all the view modifiers on the image can be moved down to the next conditional compilation block. None of them are needed on visionOS and they are still working as expected on iOS.

Sources/ArcGISToolkit/Components/Popups/PopupView.swift
```diff
@@ -89,11 +88,14 @@ public struct PopupView: View {
                     Button(action: {
                         isPresented?.wrappedValue = false
                     }, label: {
-                        Image(systemName: "xmark.circle")
-                            .foregroundColor(.secondary)
-                            .padding([.top, .bottom, .trailing], 4)
+                        Image(systemName: "xmark")
```
philium, line 91: I know that this is somewhat beyond the scope of this PR, but not entirely. According to the documentation, we should be using a different initializer:

> image

zkline101: Ok, I can make that change here but this seems like something that should be addressed across the toolkit. Can you make an issue for it?

philium: Yes, I can make an issue.

Sources/ArcGISToolkit/Components/Popups/PopupView.swift
```diff
@@ -86,14 +85,16 @@ public struct PopupView: View {
                 }
                 Spacer()
                 if showCloseButton {
-                    Button(action: {
+                    Button("Close", systemImage: "xmark") {
```
philium, line 88: "Close" needs to be localized.

philium: You can add the localized text to `Text.swift` and then reference that here.

Sources/ArcGISToolkit/Components/Popups/PopupView.swift
philium: #942

philium (approved):

dfeinzimer (approved):
Looks good!

PR #946, opened 2024-10-30
Drop AR components from Catalyst

### LINKED ISSUE #889: AR Components should be marked as unavailable for Mac Catalyst
They don't make much sense in the context of mac catalyst so they shouldn't be available there.

dfeinzimer:
Closes #889 

cc @dg0yal

### DIFF
--- a/Examples/Examples/FlyoverExampleView.swift
+++ b/Examples/Examples/FlyoverExampleView.swift
@@ -16,6 +16,7 @@ import ArcGIS
 import ArcGISToolkit
 import SwiftUI
 
+@available(macCatalyst, unavailable)
 struct FlyoverExampleView: View {
     @State private var scene = Scene(
         item: PortalItem(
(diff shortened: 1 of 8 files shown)


zkline101 (approved):

dfeinzimer:
Does this break semantic versioning? Do we need to do a deprecation instead?

zkline101:
> Does this break semantic versioning? 

Technically yes because this is a breaking change but these components I'm pretty sure never worked for mac catalyst anyway, right? Did you running the examples?

dfeinzimer:
> Did you running the examples?

Flyover somewhat works with a mouse but it's not a great experience. Tabletop and World Scale are non-operational.

rolson:
I think it is ok to deprecate all three on Mac Catalyst. Technically a breaking change, but it shouldn't affect anybody.

rolson (approved):

mhdostal (approved):

PR #950, opened 2024-11-01
`UtilityNetworkTrace` - Starting point improvements

### LINKED ISSUE #932: Utility Network Trace: Persist starting points after trace
Starting points are currently cleared after each trace.

User stories:
> - As a dispatcher comparing incident reports, I want to be able to perform analysis on a new trace type without losing the specified starting points.
> - As a dispatcher or field engineer, I want to be able to reset the trace by clearing trace results and starting points.

dfeinzimer:
Closes #932 

Addresses lack of support for the following users stories in the design spec:
> - As a dispatcher comparing incident reports, I want to be able to perform analysis on a new trace type without losing the specified starting points.
> - As a dispatcher or field engineer, I want to be able to reset the trace by clearing trace results and starting points.

Changes:
1. Because external points are provided as a binding property it is a natural expectation that the component mutates the external set. Therefore the following changes are made:
    1. Adds a new "Delete All Starting Points" button which deletes all starting points (those created internally and those provided via the binding). <p align="center"></p><p align="center"></p>
    1. When an individual starting point is deleted via the UI it will also now be deleted from the set of external starting points, if it exists in it.
1. All starting points from the previous trace are now added to the next trace despite how they were added. (Easily clearable by the "Delete All Starting Points" button.)
1. Fixes a bug where clearing or removing an external starting point doesn't clear it from the list on the pending trace.


Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTraceViewModel.swift
```diff
@@ -438,6 +448,10 @@ import SwiftUI
     
     /// Adds programatic starting points to the pending trace.
     private func addExternalStartingPoints() async {
+        pendingTrace.startingPoints.forEach { startingPoint in
+            deleteStartingPoint(startingPoint)
+        }
+
```
dfeinzimer, line 454: One open question here. When the external set of starting points changes should we clear all of the current starting points or only those also externally set?

dfeinzimer:
To be merged following #864

Programmatic starting points can be tested with this modified example [UtilityNetworkTraceExampleView.swift.zip](https://github.com/user-attachments/files/17853648/UtilityNetworkTraceExampleView.swift.zip)

mhdostal (approved):

PR #954, opened 2024-11-01
Mark `ARTests` unavailable on Catalyst

dfeinzimer:
Followup to #946

### DIFF
--- a/Tests/ArcGISToolkitTests/ARTests.swift
+++ b/Tests/ArcGISToolkitTests/ARTests.swift
@@ -18,6 +18,7 @@ import SwiftUI
 import XCTest
 @testable import ArcGISToolkit
 
+@available(macCatalyst, unavailable)
 final class ARTests: XCTestCase {
     @MainActor
     func testFlyoverLocationInitWithDefaults() throws {
--- a/Tests/ArcGISToolkitTests/ARTests.swift
+++ b/Tests/ArcGISToolkitTests/ARTests.swift
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#if !os(visionOS) || !targetEnvironment(macCatalyst)
+#if !os(visionOS) && !targetEnvironment(macCatalyst)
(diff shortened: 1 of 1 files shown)


zkline101 (commented):
The problem here is these tests will still get run on mac catalyst. Sadly, marking a test class as unavailable for a platform won't prevent them from running on that platform.

You will need to prevent them from getting run by using compiler directives (`#if`).

zkline101:
I ran into the same situation when adding visionOS support.

Tests/ArcGISToolkitTests/ARTests.swift
```diff
@@ -12,12 +12,13 @@
...
 // limitations under the License.
 
-#if !os(visionOS)
+#if !os(visionOS) || !targetEnvironment(macCatalyst)
 import ArcGIS
 import SwiftUI
 import XCTest
 @testable import ArcGISToolkit
 
+@available(macCatalyst, unavailable)
```
zkline101, line 21: This is no longer needed
```suggestion
```

dfeinzimer: Even with `#if !os(visionOS) || !targetEnvironment(macCatalyst)`, without this in place the tests still show from the Test Navigator in Xcode with Catalyst as the destination and causes errors when you try to run them:

<p align="center">
  
</p>

With `@available(macCatalyst, unavailable)` in place, they still show in Test Navigator but don't run.

Tests/ArcGISToolkitTests/ARTests.swift
```diff
@@ -12,12 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#if !os(visionOS)
+#if !os(visionOS) || !targetEnvironment(macCatalyst)
```
zkline101, line 15:
```suggestion
#if !os(visionOS) && !targetEnvironment(macCatalyst)
```

dfeinzimer:
Closing in reference to #955

PR #956, opened 2024-11-01
Drop AR components from Catalyst

### LINKED ISSUE #889: AR Components should be marked as unavailable for Mac Catalyst

dfeinzimer:
Closes #889 

Re-removes AR support for Mac Catalyst after it had to be reverted in #955 to allow the Samples app to prepare.


dfeinzimer:
Waiting on [Samples 539](https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/539)

PR #958, opened 2024-11-05
`UtilityNetworkTrace` - Auto zoom to trace extent

### LINKED ISSUE #947: Utility Network Trace: Auto zoom to trace extent
The Swift component is misaligned with the design spec here.

> By default, the control will zoom to the trace result area.

dfeinzimer:
Closes #947 

Aligns the component with the following design spec:

> By default, the control will zoom to the trace result area. This feature can be disabled by setting the AutoZoomToTraceResults property of the control to false.

Behavioral changes:
- When a trace completes, the viewpoint is automatically set to the extent of the trace's results.
  - This can still be disabled by the user under the "Advanced Options" when creating a trace. 
- While browsing the list of completed traces, the viewpoint now changes to reflect the extent of the currently selected trace.

https://github.com/user-attachments/assets/577a0aa1-59d6-441c-bd2b-28111196dba4

### DIFF
--- a/Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTrace.swift
+++ b/Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTrace.swift
@@ -113,7 +113,7 @@ public struct UtilityNetworkTrace: View {
     @State private var currentActivity: UserActivity = .creatingTrace(nil)
     
     /// A Boolean value indicating whether the map should be zoomed to the extent of the trace result.
-    @State private var shouldZoomOnTraceCompletion = false
+    @State private var shouldZoomOnTraceCompletion = true
     
     /// A Boolean value indicating whether the Clear All Results confirmation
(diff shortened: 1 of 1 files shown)


mhdostal (approved):
This looks good. Does the design state that the extent should be buffered a bit to fully show the result (so it doesn't stretch right to the edge of the display)?

dfeinzimer:
> Does the design state that the extent should be buffered a bit to fully show the result (so it doesn't stretch right to the edge of the display)?

I'm not seeing any language about it in the spec but we do add a buffer [here](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/9afd05d7bb4f77f319516c9001e209c069337451/Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTraceViewModel%2BTrace.swift#L138). Should we expand it a bit more?

mhdostal:
> Should we expand it a bit more?

No, that's fine. I missed that a buffer was already being applied. Thanks.

PR #962, opened 2024-11-07
Text UI improvements

mhdostal:
Set line limit to 5 and the minimum height to 100 for multi-line text inputs.

image

For Apollo 961


dfeinzimer (changes requested):
Screenshot 2024-11-07 at 1 23 38 PM

Looks like this needs to be re-based

dfeinzimer (approved):

dfeinzimer (approved):

PR #991, opened 2024-12-26
add sensory feedback, haptics, to compass

### LINKED ISSUE #990: Built-in Sensory Feedback in Compass
As part of an app I am building I have added sensory feedback when the MapView rotation/heading snaps to zero. This has been a nice feature. I think it could be useful to build this feature into the Compass component.

rolson:
Closes #990

As part of an app I am building I have added sensory feedback when the MapView rotation/heading snaps to zero. This has been a nice feature. I think it could be useful to build this feature into the Compass component.

@dg0yal thoughts on this?

### DIFF
--- a/Sources/ArcGISToolkit/Components/Compass/Compass.swift
+++ b/Sources/ArcGISToolkit/Components/Compass/Compass.swift
@@ -55,7 +55,6 @@ public struct Compass: View {
     /// when the heading snaps to zero.
     private var snapToZeroSensoryFeedbackEnabled: Bool = false
     
-    
     /// Creates a compass with a heading based on compass directions (0° indicates a direction
     /// toward true North, 90° indicates a direction toward true East, etc.).
     /// - Parameters:
--- a/Sources/ArcGISToolkit/Components/Compass/Compass.swift
+++ b/Sources/ArcGISToolkit/Components/Compass/Compass.swift
@@ -122,6 +122,7 @@ public struct Compass: View {
 private extension View {
     /// Enables the snap to zero sensory feedback
     /// when it is available.
+    @available(visionOS, unavailable)
     @ViewBuilder
     func snapToZeroSensoryFeedback(enabled: Bool, heading: Double) -> some View {
         if #available(iOS 17.0, *) {
@@ -142,7 +143,6 @@ private extension View {
             self
         }
     }
-
 }
 
 @available(visionOS, unavailable)
--- a/Sources/ArcGISToolkit/Components/Compass/Compass.swift
+++ b/Sources/ArcGISToolkit/Components/Compass/Compass.swift
@@ -35,9 +35,8 @@ public struct Compass: View {
     /// The opacity of the compass.
     @State private var opacity: Double = .zero
     
-    /// A Boolean value indicating whether sensory feedback is enabled
-    /// when the heading snaps to zero.
-    private var snapToZeroSensoryFeedbackEnabled: Bool = false
+    /// An action to perform when the compass is tapped.
+    private var action: (() -> Void)?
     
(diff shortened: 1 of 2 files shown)


Sources/ArcGISToolkit/Components/Compass/Compass.swift
```diff
@@ -110,10 +114,37 @@ public struct Compass: View {
...
+                }
+            } else {
+                self
+            }
+        } else {
+            // Fallback on earlier versions
+            self
+        }
+    }
+
```
dfeinzimer, line 145:
```suggestion
```

Sources/ArcGISToolkit/Components/Compass/Compass.swift
```diff
@@ -35,7 +35,11 @@ public struct Compass: View {
     /// The opacity of the compass.
     @State private var opacity: Double = .zero
     
-    /// A Boolean value indicating whether  the compass should automatically
+    /// A Boolean value indicating whether sensory feedback is enabled
+    /// when the heading snaps to zero.
+    private var snapToZeroSensoryFeedbackEnabled: Bool = false
```
dfeinzimer, line 40: I'm not sure why the `action` property is out of order, but my thinking is that it is best we keep these properties alphabetized. Should we move this down below `size` and move `action` up above `autoHide`?

Sources/ArcGISToolkit/Components/Compass/Compass.swift
```diff
@@ -110,10 +114,37 @@ public struct Compass: View {
...
+private extension View {
+    /// Enables the snap to zero sensory feedback
+    /// when it is available.
+    @ViewBuilder
+    func snapToZeroSensoryFeedback(enabled: Bool, heading: Double) -> some View {
+        if #available(iOS 17.0, *) {
+            if enabled {
+                sensoryFeedback(.selection, trigger: heading) { oldValue, newValue in
+                    if (!oldValue.isZero && newValue.isZero) ||
+                        (oldValue.isZero && !newValue.isZero) {
```
dfeinzimer, line 131: I'm on the fence about snapping when coming off zero. 

I can't quite put my finger on it, but I think the slight delay in the feedback once you're already off zero is perceivable.

Do you have any thoughts on having it just when entering zero?

Sources/ArcGISToolkit/Components/Compass/Compass.swift
```diff
@@ -110,10 +114,37 @@ public struct Compass: View {
...
                 )
+                .snapToZeroSensoryFeedback(enabled: snapToZeroSensoryFeedbackEnabled, heading: heading)
         }
     }
 }
 
+private extension View {
+    /// Enables the snap to zero sensory feedback
+    /// when it is available.
+    @ViewBuilder
```
dfeinzimer, line 125: We'll need to mark this unavailable for `visionOS` to not break the build.

```suggestion
    @available(visionOS, unavailable)
    @ViewBuilder
```

Sources/ArcGISToolkit/Components/Compass/Compass.swift
rolson: The perceptible delay seems to me to be in when the compass starts animating to visible from coming off zero. The haptic seems to coincide well with when the map rotates, IMO.

I tested it both ways and both seem good to me. I could go either way.

The [HIG](https://developer.apple.com/design/human-interface-guidelines/playing-haptics#Selection) doesn't give a lot of guidance:
> Selection haptics provide feedback while the values of a UI element are changing.

dfeinzimer (dismissed):

mhdostal (dismissed):
That's cool.

dfeinzimer (approved):

PR #993, opened 2024-12-28
Feature template picker

rolson:
We have one of these in the [objc toolkit](https://github.com/Esri/arcgis-runtime-toolkit-ios/blob/main/Sources/ArcGISToolkit/TemplatePickerViewController.swift).

This is an initial PR. I still need to add tests and a tutorial. Also the map that I use in the example is not a good map for this. I need to find a better map.

image


Sources/ArcGISToolkit/Components/FeatureTemplatePicker/FeatureTemplatePicker.swift
```diff
@@ -0,0 +1,334 @@
...
+            localized: "No Feature Templates",
+            bundle: .toolkitModule,
+            comment: """
+                 A title for showing a view that tells the user there are no feature templates.
+                 """
+        )
+    }
+    static var noFeatureTemplatesDetail: String {
+        String(
+            localized: "There are no feature templates available for this map.",
```
dfeinzimer, line 101: Should we generalize this for use with scenes?

```suggestion
            localized: "There are no feature templates available.",
```

Sources/ArcGISToolkit/Components/FeatureTemplatePicker/FeatureTemplatePicker.swift
```diff
@@ -0,0 +1,334 @@
...
+        featureLayers
+            .map { (layer: $0, table: $0.featureTable) }
+            .compactMap { tuple in
+                if let table = tuple.table as? ArcGISFeatureTable {
+                    return (layer: tuple.layer, table: table)
+                } else {
+                    return nil
+                }
+            }
+    }
```
dfeinzimer, line 293: These properties can be alphabetized.

```suggestion
    /// A list containing tuples of the feature layers and the associated
    /// ArcGIS feature tables in the geo model.
    var arcGISFeatureLayersAndTables: [(layer: FeatureLayer, table: ArcGISFeatureTable)] {
        featureLayers
            .map { (layer: $0, table: $0.featureTable) }
            .compactMap { tuple in
                if let table = tuple.table as? ArcGISFeatureTable {
                    return (layer: tuple.layer, table: table)
                } else {
                    return nil
                }
            }
    }
    
    /// All the feature layers in the geo model.
    var featureLayers: [FeatureLayer] {
        flattenedLayers
            .compactMap { $0 as? FeatureLayer }
    }
    
    /// A flattened list of the layers in the geo model.
    var flattenedLayers: [Layer] {
        layers.flattened
    }
    
    /// All the layers in the geo model.
    var layers: [Layer] {
        operationalLayers
        + (basemap?.baseLayers ?? [])
        + (basemap?.referenceLayers ?? [])
    }
```

Sources/ArcGISToolkit/Components/FeatureTemplatePicker/FeatureTemplatePicker.swift
```diff
@@ -0,0 +1,334 @@
...
+/// A value that represents a section in the feature template picker.
+private struct FeatureTemplateSectionInfo: Identifiable {
+    let table: ArcGISFeatureTable
+    var infos: [FeatureTemplateInfo]
+    let id: UUID = UUID()
+    
+    func filtered(by searchText: String) -> FeatureTemplateSectionInfo {
+        guard !searchText.isEmpty else { return self }
+        var copy = self
+        copy.infos = copy.infos.filter { $0.template.name.lowercased().contains(searchText.lowercased()) }
```
dfeinzimer, line 315: Elsewhere in the the Toolkit we've used [`localizedStandardContains(_:)`](https://developer.apple.com/documentation/swift/stringprotocol/localizedstandardcontains(_:)) for search similar search functionalities.

```suggestion
        copy.infos = copy.infos.filter { $0.template.name.localizedStandardContains(searchText) }
```

> This is the most appropriate method for doing user-level string searches, similar to how searches are done generally in the system. The search is locale-aware, case and diacritic insensitive. The exact list of search options applied may change over time.

Sources/ArcGISToolkit/Components/FeatureTemplatePicker/FeatureTemplatePicker.swift
```diff
@@ -0,0 +1,334 @@
...
+            var sections = [FeatureTemplateSectionInfo]()
+            let layersAndTables = geoModel.arcGISFeatureLayersAndTables
+            await layersAndTables.map(\.table).load()
+            for (layer, table) in layersAndTables {
+                guard includeNonCreatableFeatureTemplates || table.canAddFeature else { continue }
+                var infos = [FeatureTemplateInfo]()
+                for template in table.allTemplates {
+                    let feature = table.makeFeature(template: template)
+                    let symbol = layer.renderer?.symbol(for: feature)
+                    let image = try? await symbol?.makeSwatch(scale: UIScreen.main.scale)
```
dfeinzimer, line 223: Screenshot 2024-12-30 at 1 50 56 PM

We'll need to either find an alternative or mark the component unavailable for Vision for now to avoid breaking the build.

rolson:
I added tests

rolson:
ok, added tutorial now as well

Sources/ArcGISToolkit/Components/FeatureTemplatePicker/FeatureTemplatePicker.swift
```diff
@@ -0,0 +1,340 @@
...
+    @Binding var selection: FeatureTemplateInfo?
+    
+    var body: some View {
+        HStack {
+            Label {
+                Text(info.template.name)
+                    .lineLimit(1)
+            } icon: {
+                if let image = info.image {
+                    Image(uiImage: image)
```
mhdostal, line 153: Big symbol images can get unwieldy.

```suggestion
                    Image(uiImage: image)
                        .resizable()
                        .aspectRatio(contentMode: .fit)
```
Before
image

After:
image

And it still works with "regular" symbols, including showing size gradation:

image

Sources/ArcGISToolkit/Components/FeatureTemplatePicker/FeatureTemplatePicker.swift
```diff
@@ -0,0 +1,340 @@
...
+        private func makeFeatureTemplateSections() async -> [FeatureTemplateSectionInfo] {
+            try? await geoModel.load()
+            var sections = [FeatureTemplateSectionInfo]()
+            let layersAndTables = geoModel.arcGISFeatureLayersAndTables
+            await layersAndTables.map(\.table).load()
+            for (layer, table) in layersAndTables {
+                guard includeNonCreatableFeatureTemplates || table.canAddFeature else { continue }
+                var infos = [FeatureTemplateInfo]()
+                for template in table.allTemplates {
+                    let feature = table.makeFeature(template: template)
```
mhdostal, line 221: I got a crash here with a specific map: `TreeTrimmingHealthCheck`. Ping me for specifics, as it's internal. One of the tables in the map failed to load, which caused `table.makeFeature` to fail.

```
ArcGIS/ArcGISFeatureTable+Generated.swift:375: Fatal error: Precondition failed. Invalid call. Can only call this method on a loaded table.

(lldb) po table
note: object description requested, but type doesn't implement a custom object description. Consider using "p" instead of "po" (this note will only be shown once per debug session).
<ServiceFeatureTable: 0x6000004fb1b0>

(lldb) po table.loadStatus
ArcGIS.LoadStatus.failed

(lldb) po table.loadError
▿ Optional<Error>
  ▿ some : GeodatabaseError
    ▿ domainAlreadyExists : 1 element
      - details : "The domain \"WorkFlow\" already exists but has a conflicting definition. Make sure the fields using the domain have the same field type"
      - ```

File: Sources/ArcGISToolkit/Components/FeatureTemplatePicker/FeatureTemplatePicker.swift [hunk 7]
```diff
@@ -0,0 +1,340 @@
...
+        }
+        
+        /// The feature template section information for the geo model.
+        private func makeFeatureTemplateSections() async -> [FeatureTemplateSectionInfo] {
+            try? await geoModel.load()
+            var sections = [FeatureTemplateSectionInfo]()
+            let layersAndTables = geoModel.arcGISFeatureLayersAndTables
+            await layersAndTables.map(\.table).load()
+            for (layer, table) in layersAndTables {
+                guard includeNonCreatableFeatureTemplates || table.canAddFeature else { continue }
```
mhdostal, line 218: This fixes the crash I noted with unloaded tables.
```suggestion
                guard includeNonCreatableFeatureTemplates || table.canAddFeature,
                      table.loadStatus == .loaded,
                      layer.loadStatus == .loaded else { continue }
```

mhdostal (changes requested):
I've tested this with the Feature Forms Micro App. It looks and works great! I noticed a couple of issues, which I've commented on. I'll look at the code in another go-round.

Sources/ArcGISToolkit/Components/FeatureTemplatePicker/FeatureTemplatePicker.swift
rolson: Does the layer need to be loaded as well to fix this crash or only the table?

Sources/ArcGISToolkit/Components/FeatureTemplatePicker/FeatureTemplatePicker.swift
rolson: Nice one!

Sources/ArcGISToolkit/Components/FeatureTemplatePicker/FeatureTemplatePicker.swift
mhdostal: Only the table for this specific crash. I added the layer for completeness, but not sure it's necessary.

rolson:
Actually, I'm going to look for a different solution because I'm not sure I like how it resizes some of these:

With the fix:
image

Before the fix:
image

Sources/ArcGISToolkit/Components/FeatureTemplatePicker/FeatureTemplatePicker.swift
rolson: thanks

rolson:
Ok, I fixed the image size so that it will limit it to 50x50, but not enlarge the smaller icon sizes.

mhdostal:
> Ok, I fixed the image size so that it will limit it to 50x50, but not enlarge the smaller icon sizes.

The only issue I see with that is when Dynamic Type is used the text will increase in size but not the icons. That would be "nice-to-have", but it's not a dealbreaker for me.

image

I did add padding around the image to keep the size in check.

```
Image(uiImage: image)
    .resizable()
    .aspectRatio(contentMode: .fit)
    .padding(4)
```
Here's what it looks like with Dynamic Type:

image

And without it:

image

rolson:
> The only issue I see with that is when Dynamic Type is used the text will increase in size but not the icons. That would be "nice-to-have", but it's not a dealbreaker for me.

I'd rather have features that are under 50x50 draw correctly in their correct size. To me that's important. If there is a way to do that and let the larger features grow up to the dynamic font size, that would be ideal. But I couldn't find a way to do that unfortunately.

philium:
The tool for that job is [`ScaledMetric`](https://developer.apple.com/documentation/swiftui/scaledmetric).

mhdostal (changes requested):

Sources/ArcGISToolkit/Components/FeatureTemplatePicker/FeatureTemplatePicker.swift
```diff
@@ -0,0 +1,364 @@
...
+                                }
+                            }
+                    } else {
+                        // Fallback on earlier versions
+                    }
+                } else {
+                    Image(systemName: "minus")
+                        .foregroundStyle(.secondary)
+                }
+            }
```
dfeinzimer, line 177: Should the conditions on 155 and 156 be combined so that iOS 16 at least gets the minus icon?

```suggestion
                if let image = info.image, #available(iOS 17.0, *) {
                        Image(uiImage: image)
                            .resizable()
                            .scaledToFit()
                            .containerRelativeFrame(.horizontal) { size, axis in
                                let maxSize = maxImageSize
                                if image.size.width > maxSize {
                                    // Limit icon to max size.
                                    return maxSize
                                } else {
                                    // But don't restrict how small the icon can be.
                                    return image.size.width
                                }
                            }
                } else {
                    Image(systemName: "minus")
                        .foregroundStyle(.secondary)
                }
            }
```

Sources/ArcGISToolkit/Documentation.docc/Resources/FeatureTemplatePicker/FeatureTemplatePicker.png
dfeinzimer, line 1: Other components exclude the device frame. I've recaptured the image.

FeatureTemplatePicker

Sources/ArcGISToolkit/Documentation.docc/Tutorials/FeatureTemplatePickerTutorial.tutorial
```diff
@@ -0,0 +1,39 @@
...
+        }
+        
+        @Steps {
+            @Step {
+                Create a new view with a `MapView`.
+                @Code(name: "FeatureTemplatePicker.swift", file: FeatureTemplatePickerStep1.swift)
+            }
+            
+            @Step {
+                Add button that, when tapped, will display the feature template picker.
```
dfeinzimer, line 19:
```suggestion
                Add a button that, when tapped, will display the feature template picker.
```

Tests/ArcGISToolkitTests/FeatureTemplatePickerTests.swift
```diff
@@ -0,0 +1,95 @@
...
+        try await map.load()
+        
+        let model = FeatureTemplatePicker.Model(geoModel: map, includeNonCreatableFeatureTemplates: true)
+        #expect(model.includeNonCreatableFeatureTemplates)
+        #expect(!model.isGeneratingFeatureTemplates)
+        #expect(!model.showContentUnavailable)
+        #expect(!model.showNoTemplatesFound)
+        
+        let lockedValues = OSAllocatedUnfairLock(initialState: Array<Bool>())
+        let subscription = model.$isGeneratingFeatureTemplates.sink { isGeneratingFeatureTemplates in
```
dfeinzimer, line 36: Screenshot 2025-01-03 at 1 23 26 PM

Unfortunately, this property and the same property in the test below generate a warning, and applying the suggestion causes the test to fail.

How about calling `subscription.cancel()` at the end of each test?

Sources/ArcGISToolkit/Components/FeatureTemplatePicker/FeatureTemplatePicker.swift
rolson: I still need to find a solution for iOS 16, working on that now.

Sources/ArcGISToolkit/Documentation.docc/Resources/FeatureTemplatePicker/FeatureTemplatePicker.png
rolson: thanks

Sources/ArcGISToolkit/Components/FeatureTemplatePicker/FeatureTemplatePicker.swift
```diff
@@ -0,0 +1,365 @@
...
+                    let image = try? await symbol?.makeSwatch(scale: scale)
+                    infos.append(
+                        FeatureTemplateInfo(
+                            layer: layer,
+                            table: table,
+                            template: template,
+                            image: image
+                        )
+                    )
+                }
```
dfeinzimer, line 263: Should the templates be sorted alphabetically by name?

```suggestion
                    )
                    infos.sort { $0.template.name < $1.template.name }
                }
```

Sources/ArcGISToolkit/Documentation.docc/Resources/FeatureTemplatePicker/FeatureTemplatePicker.png
dfeinzimer, line 1: Looks like the image is still the same. Not sure if you wanted to use the new one or not but no worries if you want to keep this one.

dfeinzimer (commented):
Just 2 last comments and I think that'll be it.

Sources/ArcGISToolkit/Components/FeatureTemplatePicker/FeatureTemplatePicker.swift
dfeinzimer: Perhaps the sections too?

mhdostal:
One enhancement request would be to only show the templates that apply to a specific geometry type. In the Swft FF test app, we are only allowing creating point symbols, so limiting the feature templates to those supporting point geometries would be nice.

mhdostal (dismissed):

Sources/ArcGISToolkit/Documentation.docc/Resources/FeatureTemplatePicker/FeatureTemplatePicker.png
rolson: For some reason I thought you had committed the new image to the branch. If not, where is it?

Sources/ArcGISToolkit/Components/FeatureTemplatePicker/FeatureTemplatePicker.swift
rolson: I'm not sure. I could see alphabetizing them. Currently it is in the order of layers and then the templates that the layers report. So I'm not sure what is desirable. @mhdostal do you recall what the obj-c based version did?

rolson:
> One enhancement request would be to only show the templates that apply to a specific geometry type. In the Swft FF test app, we are only allowing creating point symbols, so limiting the feature templates to those supporting point geometries would be nice.

Cool. Did the original template picker in the other toolkit do this as well?

mhdostal:
> > One enhancement request would be to only show the templates that apply to a specific geometry type. In the Swft FF test app, we are only allowing creating point symbols, so limiting the feature templates to those supporting point geometries would be nice.
> 
> Cool. Did the original template picker in the other toolkit do this as well?

No, but it's something that I'm doing for the Swift Feature Form test app; we're not concerned with the full geometry-editing experience yet, so we're only supporting adding point features.

dfeinzimer (dismissed):
Updated the image. PR looks good 👍

Sources/ArcGISToolkit/Components/FeatureTemplatePicker/FeatureTemplatePicker.swift
mhdostal: > I'm not sure. I could see alphabetizing them. Currently it is in the order of layers and then the templates that the layers report. So I'm not sure what is desirable. @mhdostal do you recall what the obj-c based version did?

There was no sorting in the obj-c template picker.

Sources/ArcGISToolkit/Components/FeatureTemplatePicker/FeatureTemplatePicker.swift
```diff
@@ -0,0 +1,365 @@
...
+    /// Returns the size that the icon should be in the feature template view.
+    /// This will scale down images that are bigger than the max,
+    /// but for images smaller than that it will not scale them up.
+    private func size(for image: UIImage, maxSize: CGFloat) -> (width: CGFloat, height: CGFloat) {
+        let xScale = min(1, maxSize / image.size.width)
+        let yScale = min(1, maxSize / image.size.height)
+        let scale = min(xScale, yScale)
+        return (
+            width: image.size.width * xScale,
+            height: image.size.height * yScale
```
rolson, line 188:
```suggestion
            width: image.size.width * scale,
            height: image.size.height * scale
```

mhdostal (dismissed):

mhdostal:
@rolson @dfeinzimer - since this is almost ready, it would be nice to get this into v.next so it can go in the upcoming release.

dfeinzimer (dismissed):

dfeinzimer:
This would be nice to have for Apollo 1740 as it needs to support picking a template as part of creating the new ArcGISFeature. 

I put together a rudimentry picker in the meantime. In my use case, though, I need to initialize it with an `ArcGISFeatureTable` instead of a `GeoModel`.

PR #997, opened 2025-01-08
Add visionOS support to Examples

philium:


Examples/Examples/FeatureFormExampleView.swift
```diff
@@ -88,7 +88,7 @@ struct FeatureFormExampleView: View {
                 .alert(
                     "The form wasn't submitted",
                     isPresented: model.alertIsPresented
-                ) {
+                ) { // swiftlint:disable:this vertical_parameter_alignment_on_call
```
dfeinzimer, line 91: Screenshot 2025-01-07 at 4 41 29 PM

This generates a new warning for me. I'm on Swiftlint 0.57.1.

Examples/Examples/TableTopExampleView.swift
```diff
@@ -16,7 +16,6 @@ import ArcGIS
 import ArcGISToolkit
 import SwiftUI
 
-@available(macCatalyst, unavailable)
```
zkline101, line 19: I think you can make the same change for `WorldScaleExampleView`  & `FlyoverExampleView` then change both target membership for those files to just iOS.

Examples/ExamplesApp/Examples.swift
```diff
@@ -42,31 +42,43 @@ struct Examples: View {
...
-    @available(macCatalyst, unavailable)
-    static let augmentedReality = Self(
-        name: "Augmented Reality",
-        examples: [
-            AnyExample("Flyover", content: FlyoverExampleView()),
-            AnyExample("Tabletop", content: TableTopExampleView()),
-            AnyExample("World Scale", content: WorldScaleExampleView())
-        ]
-    )
+    #if os(iOS) && !targetEnvironment(macCatalyst)
```
zkline101, line 45: Since line 56 starts at the beginning of this line, it should probably be the same here.

philium: Good catch! Will fix.

Examples/Examples/TableTopExampleView.swift
philium: Sure, I'll make those changes.

Examples/Examples/FeatureFormExampleView.swift
philium: That didn't work. Since that comment is unrelated to these changes, I'll remove it, which then causes a warning for me. I'll have to look into that more.

philium: Somehow I had an older version of SwiftLint installed:

image

I don't know how it got there, but it isn't there anymore. Problem solved!

zkline101 (approved):

dfeinzimer (approved):

zkline101:
@dfeinzimer Once we are done adding visionOS support to all the components we will revisit the examples app to give it a more native visionOS experience.

PR #998, opened 2025-01-08
Add visionOS support to `FloorFilter`

zkline101:
Swift #6156

Here is a picture of it but I highly recommend testing it out in the examples app:

Simulator Screenshot - Apple Vision Pro - 2025-01-10 at 13 02 19

### DIFF
--- a/Examples/Examples.xcodeproj/project.pbxproj
+++ b/Examples/Examples.xcodeproj/project.pbxproj
@@ -17,7 +17,7 @@
 		75D41B2B27C6F21400624D7C /* ScalebarExampleView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 75D41B2A27C6F21400624D7C /* ScalebarExampleView.swift */; };
 		882899FD2AB5099300A0BDC1 /* FlyoverExampleView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 882899FC2AB5099300A0BDC1 /* FlyoverExampleView.swift */; platformFilter = ios; };
 		E42BFBE92672BF9500159107 /* SearchExampleView.swift in Sources */ = {isa = PBXBuildFile; fileRef = E42BFBE82672BF9500159107 /* SearchExampleView.swift */; platformFilters = (ios, maccatalyst, ); };
-		E4624A25278CE815000D2A38 /* FloorFilterExampleView.swift in Sources */ = {isa = PBXBuildFile; fileRef = E4624A24278CE815000D2A38 /* FloorFilterExampleView.swift */; platformFilters = (ios, maccatalyst, ); };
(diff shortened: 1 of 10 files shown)


zkline101:
I just realized that level selection doesn't work in the component where the selected level isn't highlighted. Let me fix that.

zkline101:
The fix looks like this now:
Simulator Screenshot - Apple Vision Pro - 2025-01-08 at 15 55 29

The current level is now highlighted in the list. I made the current level have a background of white with black text because that is what Apple uses here:
Simulator Screenshot - Apple Vision Pro - 2025-01-08 at 15 58 31

zkline101:
@dfeinzimer @philium This is ready for review.

philium (commented):
On iOS, the currently selected facility is highlighted, but I don't see that on visionOS. Is that intentional?

zkline101:
> On iOS, the currently selected facility is highlighted, but I don't see that on visionOS. Is that intentional?

No that was not intentional. I fixed that in my last commit, do you still see this now?

philium:
I'm pretty sure I had your latest changes, but I'll try again just to be sure.

dfeinzimer (changes requested):
Looks like this changes button styles in the Catalyst level selector

| Before | After |
|---|---|
|  |  |

zkline101:
Lovely. Let me look into it.

zkline101:
@dfeinzimer I fixed it let me know what you think, thanks.

dfeinzimer:
> @dfeinzimer I fixed it let me know what you think, thanks.

Hmm, something is off still. This is what I'm seeing.

<p align="center">
  
</p>

dfeinzimer (commented):
Hmm, something is off still. This is what I'm seeing.

<p align="center">
  
</p>

zkline101:
@dfeinzimer Thanks for the help, I think I covered all the bases now. Please review.

@philium We moved away from `Button(_:action:)` in favor of `Button(action:label:)` so we could customize the text that is being used for the label. That allowed us to keep the same button design that was there previously for iOS and Mac Catalyst.

dfeinzimer (dismissed):

philium (dismissed):

philium (dismissed):

dfeinzimer (dismissed):

zkline101:
I was re-reading the design guidelines for visionOS [buttons](https://developer.apple.com/design/human-interface-guidelines/buttons#visionOS) and realized there were some areas my changes were lacking. Now the level selector buttons look consistent across platforms and match up more with what is expected from the guidelines.

These were the two areas the level selector buttons were lacking on visionOS:
Screenshot 2025-01-10 at 12 43 52 PM

Please re-review, thank you.

dfeinzimer (approved):
Looks good!

philium (approved):

PR #999, opened 2025-01-09
Add visionOS support to `Compass`

zkline101:
Swift #6156

### DIFF
--- a/Sources/ArcGISToolkit/Components/Compass/Compass.swift
+++ b/Sources/ArcGISToolkit/Components/Compass/Compass.swift
@@ -30,7 +30,6 @@ import SwiftUI
 /// To see it in action, try out the [Examples](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/tree/main/Examples/Examples)
 /// and refer to [CompassExampleView.swift](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/main/Examples/Examples/CompassExampleView.swift)
 /// in the project. To learn more about using the `Compass` see the <doc:CompassTutorial>.
-@available(visionOS, unavailable)
 public struct Compass: View {
     /// The opacity of the compass.
     @State private var opacity: Double = .zero
@@ -110,11 +109,15 @@ public struct Compass: View {
                                  """
                     )
                 )
+#if os(visionOS)
+                .hoverEffect { effect, isActive, _ in
+                    effect.scaleEffect(isActive ? 1.05 : 1.0)
+                }
+#endif
         }
     }
 }
 
-@available(visionOS, unavailable)
 extension Compass {
     /// Returns a Boolean value indicating whether the compass should hide based on the
     /// provided heading and whether the compass has been configured to automatically hide.
@@ -125,7 +128,6 @@ extension Compass {
     }
 }
 
-@available(visionOS, unavailable)
 public extension Compass {
     /// Creates a compass with a rotation (0° indicates a direction toward true North, 90° indicates
     /// a direction toward true West, etc.).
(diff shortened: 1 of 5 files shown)


Sources/ArcGISToolkit/Components/Compass/Needle.swift
```diff
@@ -23,22 +23,51 @@ struct Needle: View {
...
+                                axis: (x: 0, y: 1, z: 0)
+                            )
+#else
                             .rotation3DEffect(
-                                Angle(degrees: 180),
-                                axis: (x: 0, y: 1, z: 0))
+                                .degrees(180),
+                                axis: (x: 0, y: 1, z: 0)
+                            )
+#endif
```
zkline101, line 36: Using `rotation3DEffect` was causing parts of the needle to flicker on a Vision Pro device not a simulator. And it turns out we shouldn't use this on visionOS according to [Apple](https://developer.apple.com/documentation/swiftui/view/rotation3deffect(_:axis:anchor:anchorz:perspective:)):
Screenshot 2025-01-09 at 11 53 26 AM

Using the method they recommend resolves this problem.

Sources/ArcGISToolkit/Components/Compass/Compass.swift
```diff
@@ -110,11 +109,15 @@ public struct Compass: View {
                                  """
                     )
                 )
+#if os(visionOS)
+                .hoverEffect { effect, isActive, _ in
+                    effect.scaleEffect(isActive ? 1.05 : 1.0)
+                }
+#endif
```
zkline101, line 116: This is our first time using a custom hover effect so let me know your thoughts. The original hover effect I found was not even noticeable on a device (it was a little better on a simulator). This hover effect will increase the size of the compass very slightly when you look at it to indicate you can tap it. Apple has used something similar to this in some places. Here is what it looks like.
Simulator Screen Recording - Apple Vision Pro - 2025-01-09 at 11 56 31

zkline101: @philium Please test this on your device too.

philium: I did test it and it was barely visible to me, but perhaps other can see it better.

philium (approved):

Sources/ArcGISToolkit/Components/Compass/Compass.swift
zkline101: Do you think the scale should be larger when the hover effect is active? I'm fine with how it is now but increasing it a bit couldn't hurt. Just as long as it isn't too big where it becomes distracting.

philium: It's up to you. Probably more significant would be a highlight while the compass is clicked/tapped, before the action is actually triggered.

zkline101: >  before the action is actually triggered.

That sounds like what the original hover effect was like and it was hard to notice in my testing.

philium: Could we highlight and scale?

zkline101: I bet we can, I can see what that looks like.

zkline101: I think it looks good:
Simulator Screen Recording - Apple Vision Pro - 2025-01-09 at 14 44 12

philium (approved):

dfeinzimer (approved):

PR #1000, opened 2025-01-10
Prototype `UtilityNetworkAssociationFormElementView`

dfeinzimer:
Apollo 1030

<p align="center">
  
</p>

### DIFF
--- a/Examples/Examples/FeatureFormExampleView.swift
+++ b/Examples/Examples/FeatureFormExampleView.swift
@@ -27,7 +27,7 @@ struct FeatureFormExampleView: View {
     @State private var identifyScreenPoint: CGPoint?
     
     /// The `Map` displayed in the `MapView`.
-    @State private var map = Map(url: .sampleData)!
+    @State private var map = makeMap()
     
     /// The validation error visibility configuration of the form.
     @State private var validationErrorVisibility = FeatureFormView.ValidationErrorVisibility.automatic
@@ -56,6 +56,10 @@ struct FeatureFormExampleView: View {
                         model.state = .editing(FeatureForm(feature: feature))
                     }
                 }
+                .task {
+                    let publicSample = try? await ArcGISCredential.publicSample
+                    ArcGISEnvironment.authenticationManager.arcGISCredentialStore.add(publicSample!)
+                }
                 .ignoresSafeArea(.keyboard)
                 .floatingPanel(
                     attributionBarHeight: attributionBarHeight,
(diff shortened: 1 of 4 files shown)


Examples/Examples/FeatureFormExampleView.swift
dfeinzimer, line 1: Changes to the Example are just for ease of testing. This is just borrowing the data from the Utility Network Trace Tool example. These changes will be backed out before we merge everything into v.next.

Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView.swift
```diff
@@ -85,8 +91,9 @@ public struct FeatureFormView: View {
     /// Initializes a form view.
     /// - Parameters:
     ///   - featureForm: The feature form defining the editing experience.
-    public init(featureForm: FeatureForm) {
+    public init(featureForm: FeatureForm, utilityNetwork: UtilityNetwork? = nil) {
```
mhdostal, line 94: Instead of having the FF constructor expand to encompass new features, we should probably have the `UtilityNetworkAssociationsFormElement` have a `utilityNetwork` property. This would be keeping with the idea that the elements are the "models" for the display. For now, this is fine.

Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView.swift
```diff
@@ -131,6 +146,42 @@ public struct FeatureFormView: View {
...
+                                } else {
+                                    return nil
+                                }
+                            }
+                            associations.append(newAssociation)
+                        }
+                        groups.append(.init(associations: associations, description: "[Network Source Description]", name: "\(source.name)".capitalized))
+                    }
+                    self.groups = groups
+                }
```
mhdostal, line 180: I'm hoping a lot of this logic can be moved into the UN Associations form element.

mhdostal (approved):
Just a couple of comments.

Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView.swift
dfeinzimer: 100%, this was to get the element functional for the time being. I am definitely looking to avoid initializer changes if they can be avoided.

Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView.swift
dfeinzimer: Yep same here, this is just temporary code to get things up and running for now. Planning to pull this out before we merge into v.next.

PR #1003, opened 2025-01-13
Fix visionOS build

dfeinzimer:
Followup #991


rolson (approved):

zkline101 (commented):
I can't test because my Xcode is unusable right now but this looks ok.

zkline101 (approved):

Issue #842, opened 2024-08-21
Remove use of deprecated view modifiers


des12437:
The following view modifiers are deprecated and should be replaced with the following:
  - `.foregroundColor(.accentColor)` -> `.foregroundStyle(Color.accentColor)`
  - `.background(Color(UIColor.tertiarySystemFill))` -> `. backgroundStyle(.tertiary)`
  - `.cornerRadius(5)` -> `.clipShape(.rect(cornerRadius: 16))`

PR #1005, opened 2025-01-15
Remove use of `View.foregroundColor(_:)`

philium:
Related to #842. The replacement is `foregroundStyle(_:)`. While working on visionOS support, we have already done this in many places. This just does it across the board.

Note that there is one more use of what appears to be this modifier here:

https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/453250b297bddd6e587b3f6536ee4b684dd26a00/Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/TextInput.swift#L119

But it is actually a different modifier, `Text.foregroundStyle(_:)`, which was deprecated in iOS 17. I chose to leave it as-is.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Compass/CompassBody.swift
+++ b/Sources/ArcGISToolkit/Components/Compass/CompassBody.swift
@@ -22,8 +22,11 @@ struct CompassBody: View {
             Circle()
                 .inset(by: borderWidth / 2.0)
                 .stroke(lineWidth: borderWidth)
-                .foregroundColor(.outline)
-                .background(Circle().foregroundColor(.fill))
+                .foregroundStyle(Color.outline)
+                .background {
+                    Circle()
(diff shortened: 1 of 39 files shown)


Sources/ArcGISToolkit/Components/Compass/CompassBody.swift
```diff
@@ -22,8 +22,11 @@ struct CompassBody: View {
             Circle()
                 .inset(by: borderWidth / 2.0)
                 .stroke(lineWidth: borderWidth)
-                .foregroundColor(.outline)
-                .background(Circle().foregroundColor(.fill))
```
philium, line 26: This modifier was also deprecated, which is why the code was changed to use the non-deprecated closure-based replacement.

mhdostal (approved):

Issue #1004, opened 2025-01-14
OfflineMapAreasView: Ambiguous Buttons


rolson:
The buttons to get details of the offline map view are ambiguous. Take a look at the following screen shot:

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/8e268522-9ffb-4a60-8313-17fccca1b73f" />

The top two items are not tappable. The bottom two items are tappable and show you item details. For example:

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/36a98759-af69-454a-a331-cc1de10e0f38" />

This ambiguity is not a good user experience. Something needs to denote that the cell is tappable, or there needs to be another button available to tap to get details.

des12437:
Closed as completed: https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/1010

PR #1010, opened 2025-01-15
Offline: Add download button to metadata view

des12437:
Closes https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/1004.

Adds tap gesture to shows metadata view for all map areas.

- When map area is not downloaded:
  - Shows "Download" button
- When map area is downloading:
  - "Download" button is disabled
- When map area is downloaded:
  - Shows file size
  - Shows "Remove Download" button

Screenshots:

### DIFF
--- a/Sources/ArcGISToolkit/Components/Offline/PreplannedMetadataView.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/PreplannedMetadataView.swift
@@ -59,7 +59,6 @@ struct PreplannedMetadataView: View {
                 }
                 if model.status.isDownloaded {
                     VStack(alignment: .leading) {
-                        
                         Text("Size")
                             .font(.caption)
                             .foregroundStyle(.secondary)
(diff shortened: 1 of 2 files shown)


rolson (approved):

Sources/ArcGISToolkit/Components/Offline/PreplannedMetadataView.swift
```diff
@@ -57,26 +57,43 @@ struct PreplannedMetadataView: View {
...
                 }
-                VStack(alignment: .leading) {
-                    Text("Size")
-                        .font(.caption)
-                        .foregroundStyle(.secondary)
-                    Text(Int64(model.directorySize), format: .byteCount(style: .file, allowedUnits: [.kb, .mb]))
-                        .font(.subheadline)
+                if model.status.isDownloaded {
+                    VStack(alignment: .leading) {
+
```
yo1995, line 62:
```suggestion
```

Sources/ArcGISToolkit/Components/Offline/PreplannedMetadataView.swift
```diff
@@ -57,26 +57,43 @@ struct PreplannedMetadataView: View {
...
+                        Task { await model.downloadPreplannedMapArea() }
+                    } label: {
+                        HStack {
+                            Image(systemName: "arrow.down.circle")
+                            Text("Download")
                         }
                     }
+                    .foregroundStyle(Color.accentColor)
+                    .disabled(!model.status.allowsDownload)
                 }
```
yo1995, line 97: This branch seems to me should not be nested, maybe sth like

```swift
if model.status.isDownloaded && !isSelected {
    // Show the trash button
}
if !model.status.isDownloaded {
    // Show the download button
}
```

because an area not downloaded won't be selected.

Sources/ArcGISToolkit/Components/Offline/PreplannedMetadataView.swift
```diff
@@ -79,6 +81,19 @@ struct PreplannedMetadataView: View {
                     }
                 }
             }
+            if !model.status.isDownloaded {
+                Button {
```
yo1995, line 85: The view for the trash button is an HStack (In a Section) with image and button, while this is a button with a HStack label. Slightly different structure but no visible differences per se. Up to you to keep consistent or not.

Though I was a bit surprised when I tapped on the trash icon instead of the button and the button still worked.

yo1995 (approved):

Sources/ArcGISToolkit/Components/Offline/PreplannedMetadataView.swift
des12437: The Button with a label was used so the entire view could be disabled via `.disabled(!model.status.allowsDownload)`.

des12437: I will keep the UI as-is. I notice that applying the disabled modifier to the HStack or Section makes the button and label colors look non-uniform:

philium: FYI: This button could be much simpler. Here's a comparison of the two ways, showing that they are indeed equivalent:

image

des12437: I tried this approach and the UI does not look equivalent. I prefer the UI of the first button with the smaller icon size and spacing between the button and label. Am I missing something?

Screenshot 2025-01-16 at 11 12 19 AM

philium: You can do what you want, but what I've found is that usually a simpler, system provided approach is better than a custom one. That's especially been true as we've been adding visionOS support to various components. For example, a list that uses either the list selection mechanism or a button just works, whereas a list with text (or an H stack) and a tap gesture doesn't work as expected on visionOS.

There's a reason Apple included the `Button` initializers that take a system image, and my guess is so that there is a consistent appearance for buttons with both an image and text. Using one of those initializers means that the button's appearance will be more in line with the stand appearance elsewhere in iOS. But if you want a custom appearance with a more complex implementation, then so be it.

des12437: Thanks for pointing out the difference in behavior between a system button and a custom implementation. I don't think the customization is worth the tradeoff. I removed the custom button.

yo1995 (approved):

rolson (approved):

PR #1011, opened 2025-01-16
Add visionOS support to `UtilityNetworkTrace`

philium:


dfeinzimer (approved):

zkline101 (commented):
Some feedback:
- [This](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/aaac44f6a7f86621158fbf210c404849fe59784e/Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTrace.swift#L314) can be removed for visionOS so the button has a default circle around it.
- In the example file, the `UtilityNetworkTrace` could use a bit of top padding since it is cramped on the top of the floating panel
- The "Cancel Starting Point Selection" is a bit of an eye sore and should look like this. Where the background of the button is red and the text is white.
Screenshot 2025-01-17 at 9 53 53 AM

- Under "Advanced Options", it isn't clear the "Name" row is selectable to type in text. So adding a hover effect to the `TextField` there could probably help.

zkline101:
Also, I was expecting each section to look like it does in iOS:
Simulator Screenshot - iPad Pro 13-inch (M4) - 2025-01-17 at 10 12 39

where there is a background color for each row. Similar to how the settings app is in visionOS or iOS. I'm assuming there is something in the code that is preventing that default behavior on visionOS .

philium:
> The "Cancel Starting Point Selection" is a bit of an eye sore and should look like this. Where the background of the button is red and the text is white.

The dialogue is created using the built-in confirmation dialog and the delete button has a destructive role. While it may look odd compared to other buttons, it matches other system confirmation dialogues, such as this one:

image

philium:
I suspect that the lists aren't styled as grouped because the list is not the root of the view in the navigation stack.

zkline101:
> I suspect that the lists aren't styled as grouped because the list is not the root of the view in the navigation stack.

Should we try that? Or do you think that would require too many changes?

zkline101:
Please remove`#if !os(visionOS)` from`UtilityNetworkTraceViewModelTests` so those tests can be run on visionOS.

zkline101:
@mhdostal @dfeinzimer 

I am helping Phil finish these PRs while he is out on vacation, can you review this? Thank you.

zkline101:
> I suspect that the lists aren't styled as grouped because the list is not the root of the view in the navigation stack.

I created an [issue](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/1021) to come back to this.

dfeinzimer (approved):

mhdostal (approved):

zkline101:
Fixed merge conflict, please re-review.

mhdostal (approved):

dfeinzimer (approved):

PR #1012, opened 2025-01-16
cleanup, move more into the model

rolson:
Please review asap.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
@@ -21,7 +21,7 @@ extension OfflineMapAreasView {
     @MainActor
     class MapViewModel: ObservableObject {
         /// The portal item ID of the web map.
-        private let portalItemID: PortalItem.ID?
+        private let portalItemID: PortalItem.ID
         
         /// The offline map task.
         private let offlineMapTask: OfflineMapTask
@@ -29,47 +29,63 @@ extension OfflineMapAreasView {
         /// The preplanned map information.
         @Published private(set) var preplannedMapModels: Result<[PreplannedMapModel], Error>?
         
-        /// The offline preplanned map information sourced from downloaded mobile map packages.
-        @Published private(set) var offlinePreplannedMapModels: [PreplannedMapModel]?
+        /// A Boolean value indicating if only offline models are being shown.
+        @Published private(set) var isShowingOnlyOfflineModels = false
+        
+        /// The online map.
+        private let onlineMap: Map
         
         /// Creates an offline map areas view model for a given web map.
         /// - Parameter map: The web map.
+        /// - Precondition: `map.item?.id` is not `nil`.
         init(map: Map) {
+            precondition(map.item?.id != nil)
             offlineMapTask = OfflineMapTask(onlineMap: map)
-            portalItemID = map.item?.id
+            onlineMap = map
+            portalItemID = map.item!.id!
         }
         
         /// Gets the preplanned map areas from the offline map task and loads the map models.
         func loadPreplannedMapModels() async {
-            guard let portalItemID else { return }
-            
             if offlineMapTask.loadStatus != .loaded {
                 try? await offlineMapTask.retryLoad()
             }
             
+            // Reset flag
+            isShowingOnlyOfflineModels = false
+            
             preplannedMapModels = await Result { @MainActor in
-                try await offlineMapTask.preplannedMapAreas
-                    .filter { $0.portalItem.id != nil }
-                    .sorted(using: KeyPathComparator(\.portalItem.title))
-                    .map {
-                        PreplannedMapModel(
-                            offlineMapTask: offlineMapTask,
-                            mapArea: $0,
-                            portalItemID: portalItemID,
-                            preplannedMapAreaID: $0.portalItem.id!
-                        )
+                do {
+                    return try await offlineMapTask.preplannedMapAreas
+                        .filter { $0.portalItem.id != nil }
+                        .sorted(using: KeyPathComparator(\.portalItem.title))
+                        .map {
+                            PreplannedMapModel(
+                                offlineMapTask: offlineMapTask,
+                                mapArea: $0,
+                                portalItemID: portalItemID,
+                                preplannedMapAreaID: $0.portalItem.id!
+                            )
+                        }
+                } catch {
+                    // If not connected to the internet, then return only the offline models.
+                    if let urlError = error as? URLError,
+                       urlError.code == .notConnectedToInternet {
+                        isShowingOnlyOfflineModels = true
+                        return await loadOfflinePreplannedMapModels()
+                    } else {
+                        throw error
                     }
+                }
             }
         }
         
         /// Loads the offline preplanned map models with information from the downloaded mobile map
         /// packages for the online map.
-        func loadOfflinePreplannedMapModels() async {
-            guard let portalItemID else { return }
-            
+        func loadOfflinePreplannedMapModels() async -> [PreplannedMapModel] {
             let preplannedDirectory = URL.preplannedDirectory(forPortalItemID: portalItemID)
             
-            guard let mapAreaIDs = try? FileManager.default.contentsOfDirectory(atPath: preplannedDirectory.path()) else { return }
+            guard let mapAreaIDs = try? FileManager.default.contentsOfDirectory(atPath: preplannedDirectory.path()) else { return [] }
             
             var preplannedMapModels: [PreplannedMapModel] = []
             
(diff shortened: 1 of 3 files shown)


Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -109,47 +104,14 @@ public struct OfflineMapAreasView: View {
...
-                view(for: error)
-            }
+            view(for: error)
         case .none:
             ProgressView()
                 .frame(maxWidth: .infinity)
         }
     }
     
-    @ViewBuilder private var offlinePreplannedMapAreasView: some View {
```
rolson, line 124: This code was basically duplicated

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -161,18 +123,6 @@ public struct OfflineMapAreasView: View {
         .frame(maxWidth: .infinity)
     }
     
-    private var emptyOfflinePreplannedMapAreasView: some View {
```
rolson, line 164: duplicated

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -196,15 +146,6 @@ public struct OfflineMapAreasView: View {
         }
     }
     
-    private var offlineBannerView: some View {
```
rolson, line 199: wasn't used

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -214,14 +155,6 @@ public struct OfflineMapAreasView: View {
         }
         .frame(maxWidth: .infinity)
     }
-    
-    /// Loads the online and offline preplanned map models.
-    private func loadPreplannedMapModels() async {
```
rolson, line 219: no longer necessary

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
```diff
@@ -122,6 +138,12 @@ extension OfflineMapAreasView {
                 thumbnail: item.thumbnail
             )
         }
+        
+        
+        /// A Boolean value indicating whether the web map is offline disabled.
+        var mapIsOfflineDisabled: Bool {
+            onlineMap.loadStatus == .loaded && onlineMap.offlineSettings == nil
+        }
```
rolson, line 146: this was in the view, but it's logic so it's best to go in the model

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
```diff
@@ -21,55 +21,71 @@ extension OfflineMapAreasView {
...
+        @Published private(set) var isShowingOnlyOfflineModels = false
+        
+        /// The online map.
+        private let onlineMap: Map
         
         /// Creates an offline map areas view model for a given web map.
         /// - Parameter map: The web map.
+        /// - Precondition: `map.item?.id` is not `nil`.
         init(map: Map) {
+            precondition(map.item?.id != nil)
```
rolson, line 42: This should be safe right?

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
```diff
@@ -21,55 +21,71 @@ extension OfflineMapAreasView {
...
+                                portalItemID: portalItemID,
+                                preplannedMapAreaID: $0.portalItem.id!
+                            )
+                        }
+                } catch {
+                    // If not connected to the internet, then return only the offline models.
+                    if let urlError = error as? URLError,
+                       urlError.code == .notConnectedToInternet {
+                        isShowingOnlyOfflineModels = true
+                        return await loadOfflinePreplannedMapModels()
```
rolson, line 75: This is logic, thus it should be in the model.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
yo1995: Had same finding in swift/issues/6467. Per my test it should be safe - If a webmap is offline enabled, it's guaranteed to have an ID.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
```diff
@@ -90,7 +106,7 @@ extension OfflineMapAreasView {
                 preplannedMapModels.append(model)
             }
             
-            offlinePreplannedMapModels = preplannedMapModels
+            return preplannedMapModels
                 .filter(\.status.isDownloaded)
```
des12437, line 110:
```suggestion
```
This filter is not necessary.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
des12437, line 146: This computed property can be placed near the top of the file along with other property declarations.

des12437 (approved):
Looks good just a few comments.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
yo1995: Had the same finding in swift/issues/6467. Per my test it should be safe - If a webmap is offline enabled (i.e. not a purely local map), it's guaranteed to have an ID.

However, I don't think it is safe to write it this way. For example, if I simply pass in a map `Map(basemapStyle: .arcGISLightGray)`, it will crash on this line due to the map doesn't have an item or ID.

I would probably do it like this:

- Make the `portalItemID` an implicitly unwrapped optional, `private let portalItemID: PortalItem.ID!`
    - Because we know that whenever we need to access it, it doesn't make sense to be `nil`
    - A map can have `nil` item, but whichever map has a `nil` item, it doesn't support offline workflow
- Make sure no method will be called when a map's item is `nil` in UI

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
```diff
@@ -21,55 +21,71 @@ extension OfflineMapAreasView {
...
         
         /// Gets the preplanned map areas from the offline map task and loads the map models.
         func loadPreplannedMapModels() async {
-            guard let portalItemID else { return }
-            
             if offlineMapTask.loadStatus != .loaded {
                 try? await offlineMapTask.retryLoad()
             }
             
+            // Reset flag
```
yo1995, line 54:
```suggestion
            // Reset flag.
```

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
rolson: I don't think it's necessary, but I'd prefer an issue to double check it and come back to it. Because the directory could have gotten corrupted or the contents of the directory removed.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
yo1995: Should this precondition be removed or do you want to keep it as a second fail-safe? Since there is only 1 place that calls this internal initializer and a precondition is in that place, this precondition won't be triggered?

yo1995 (approved):

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -33,18 +33,15 @@ public struct OfflineMapAreasView: View {
...
     }
     
-    /// A Boolean value indicating whether the web map is offline disabled.
-    private var mapIsOfflineDisabled: Bool {
-        onlineMap.loadStatus == .loaded && onlineMap.offlineSettings == nil
-    }
-    
     /// Creates a view with a given web map.
     /// - Parameters:
     ///   - online: The web map to be taken offline.
```
des12437, line 38:
```suggestion
    ///   - onlineMap: The web map to be taken offline.
```

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
rolson: best to stay

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
des12437: Okay I can log an issue.

PR #1014, opened 2025-01-17
Offline manager: Move models into manager

rolson:


Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
```diff
@@ -38,9 +38,15 @@ public class OfflineManager: ObservableObject {
...
     static private let defaultsKey = "com.esri.ArcGISToolkit.offlineManager.offlineMaps"
     
+    /// The available offline map view models.
+    private var models: [Item.ID: OfflineMapViewModel] = [:]
+    
     private init() {
         Logger.offlineManager.debug("Initializing OfflineManager")
         
+        // Retrieves the offline map infos from the user defaults.
+        loadOfflineMapInfosFromDefaults()
```
rolson, line 48: this should happen before resuming jobs.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
```diff
@@ -16,133 +16,143 @@ import ArcGIS
...
-        
-        /// The online map.
-        private let onlineMap: Map
-        
-        /// A Boolean value indicating whether the web map is offline disabled.
-        var mapIsOfflineDisabled: Bool {
-            onlineMap.loadStatus == .loaded && onlineMap.offlineSettings == nil
+/// The model class that represents information for an offline map.
+@MainActor
+class OfflineMapViewModel: ObservableObject {
```
rolson, line 21: don't embed this in the view as the manager will now use it

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
```diff
@@ -16,133 +16,143 @@ import ArcGIS
...
+                return try await offlineMapTask.preplannedMapAreas
+                    .filter { $0.portalItem.id != nil }
+                    .sorted(using: KeyPathComparator(\.portalItem.title))
+                    .map {
+                        PreplannedMapModel(
+                            offlineMapTask: offlineMapTask,
+                            mapArea: $0,
+                            portalItemID: portalItemID,
+                            preplannedMapAreaID: $0.portalItem.id!,
+                            onRemoveDownload: onRemoveDownloadOfPreplannedArea(withID:)
```
rolson, line 72: add closure to when a download is removed so we can remove the saved state.

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
```diff
@@ -95,14 +92,6 @@ struct PreplannedListItemView: View {
                 self.downloadState = downloadState
             }
         }
-        .task(id: downloadState) {
-            if previousDownloadState == .notDownloaded && downloadState == .downloading {
-                onDownloadAction?()
-            }
-            if previousDownloadState == .downloaded && downloadState == .notDownloaded {
```
rolson, line 102: it's not great to check the state for determining if it was removed. The reason being is that we have a function to remove the download so we can know for certain when it happens.

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -52,12 +55,12 @@ class PreplannedMapModel: ObservableObject, Identifiable {
     var map: Map? { 
         get async {
             if let mobileMapPackage {
-                if mobileMapPackage.loadStatus != .loaded {
```
rolson, line 55: You don't need to check the load status because calling load will do that.

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -52,12 +55,12 @@ class PreplannedMapModel: ObservableObject, Identifiable {
...
-                    do {
-                        try await mobileMapPackage.load()
-                    } catch {
-                        status = .mmpkLoadFailure(error)
-                    }
+                do {
+                    try await mobileMapPackage.load()
+                } catch {
+                    // TODO: it's odd to have this as a side effect of simply asking
+                    // for the map. This needs to be fixed.
```
rolson, line 62: I'll fix this later.

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -86,9 +77,24 @@ class PreplannedMapModel: ObservableObject, Identifiable {
...
+        }
+    }
+    
+    /// Tries to load a mobile map package and if successful, then updates state
+    /// associated with it.
+    private func loadAndUpdateMobileMapPackage(mmpk: MobileMapPackage) async {
+        do {
+            try await mmpk.load()
             mobileMapPackage = mmpk
             directorySize = FileManager.default.sizeOfDirectory(at: mmpkDirectoryURL)
```
rolson, line 91: This was being set in multiple places. The loadAndUpdateMobileMapPackage consolidates it.

Sources/ArcGISToolkit/Components/Offline/PreplannedMapModel.swift
```diff
@@ -49,33 +52,21 @@ class PreplannedMapModel: ObservableObject, Identifiable {
...
     
     /// The first map from the mobile map package.
-    var map: Map? { 
-        get async {
-            if let mobileMapPackage {
-                if mobileMapPackage.loadStatus != .loaded {
-                    do {
-                        try await mobileMapPackage.load()
-                    } catch {
-                        status = .mmpkLoadFailure(error)
```
rolson, line 59: Setting a load failure here is a side-effect of the asynchronous property getter which isn't good. Best to update this property when a mobile map package is set, and make this property published.

Sources/ArcGISToolkit/Components/Offline/PreplannedListItemView.swift
```diff
@@ -95,14 +92,6 @@ struct PreplannedListItemView: View {
                 self.downloadState = downloadState
             }
         }
-        .task(id: downloadState) {
-            if previousDownloadState == .notDownloaded && downloadState == .downloading {
```
rolson, line 99: It's not ideal to check the status for when a download happens because we should know exactly when it happens in the model. Also this logic should be in a model if possible and not a view. I moved it.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift
```diff
@@ -89,18 +102,6 @@ public struct OfflineMapAreasView: View {
...
-                            guard let portalItem else { return }
-                            OfflineManager.shared.saveMapInfo(for: portalItem)
-                        }
-                        .onRemoveDownload {
-                            guard let portalItemID = portalItem?.id,
-                                  models.filter(\.status.isDownloaded).isEmpty else { return }
-                            OfflineManager.shared.deleteMapInfo(for: portalItemID)
-                        }
-                        .onChange(of: selectedMap) { _ in
-                            dismiss()
```
rolson, line 102: dismiss can happen when we set it in the child view, no need for the onChange modifier here

des12437 (commented):
Changes look good. Just a few comments.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.MapViewModel.swift
```diff
@@ -16,133 +16,143 @@ import ArcGIS
...
+            preplannedMapModels.append(model)
         }
+        
+        return preplannedMapModels
+            .filter(\.status.isDownloaded)
+            .sorted(by: { $0.preplannedMapArea.title < $1.preplannedMapArea.title })
+    }
+    
+    /// The function called when a downloaded preplanned map area is removed.
+    func onRemoveDownloadOfPreplannedArea(withID preplannedAreaID: Item.ID) {
```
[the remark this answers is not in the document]
rolson: I actually think all of them should change to Item.ID since Item is where ID is defined. It's on my list to change all those to Item, I was going to do that later.

des12437 (approved):

PR #1008, opened 2025-01-15
Prototype `UtilityNetworkAssociationFormElementView` external navigation

dfeinzimer:
Apollo 1054

This is the first of two planned prototypes for testing different navigation patterns around association selection. This prototype places navigation responsibility on the component user (developer).

See also #1015, #1025

https://github.com/user-attachments/assets/841c17ae-21fc-4d67-b0bd-b153661b4aa9

PR #1015, opened 2025-01-17
Prototype `UtilityNetworkAssociationFormElementView` with custom navigation

dfeinzimer:
Apollo 1054

This is the second of ~two~ three planned prototypes for testing different navigation patterns around association selection. This prototype implements internal navigation within the FeatureFormView.

See also #1008, #1025

https://github.com/user-attachments/assets/a9858c1b-3ddf-491e-8a37-81cfabf45798


dfeinzimer:
We've decided to proceed with this internal navigation prototype.

Issue #1022, opened 2025-01-21
Crash when requesting camera access


mhdostal:
There is a situation where requesting camera/video access when adding an attachment will fail.

- New install only (so you get the "request access" alert and not the "go to settings" alert)
- Open a map with attachments
- Open a Feature Form
- In the "Attachments" section tap the "Add Attachment" button and select "Take Photo or Video"
- Deny access to the camera when prompted
- Crash

This only happens for new installs and only if you deny access in the alert.

dfeinzimer:
I'm not seeing a crash immediately upon denying camera access. 

Using the FF micro-app, I do see a background crash on my 9th gen iPad in landscape when leaving the app to go to device settings to grant camera access:
```
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Layout requested for visible navigation bar, <SwiftUI.UIKitNavigationBar: 0x120cc2580; baseClass = UINavigationBar; frame = (0 0; 320 56); opaque = NO; autoresize = W; gestureRecognizers = <NSArray: 0x11bde3630>; layer = <CALayer: 0x1152d7040>> delegate=0x106ae2a00, when the top item belongs to a different navigation bar. topItem = <UINavigationItem: 0x120cc2800> title='testPoints' style=navigator hidesBackButton leftItemsSupplementBackButton trailingItemGroups=0x106b4e340 largeTitleDisplayMode=never, navigation bar = <SwiftUI.UIKitNavigationBar: 0x1041dde00; baseClass = UINavigationBar; frame = (0 0; 535 56); opaque = NO; autoresize = W; gestureRecognizers = <NSArray: 0x120c6a700>; layer = <CALayer: 0x1151f48c0>> delegate=0x106ae1500, possibly from a client attempt to nest wrapped navigation controllers.'
*** First throw call stack:
(0x19dbe80c0 ... 0x1c4571f08)
libc++abi: terminating due to uncaught exception of type NSException
```

On the contrary, when approving camera access I'm seeing a blank screen upon approving camera access.

PR #1023, opened 2025-01-21
Mhd/camera requester crash

mhdostal:
#1022

Uses async version of `AVCaptureDevice.requestAccess(for:)`.

### DIFF
--- a/Sources/ArcGISToolkit/Common/CameraRequester.swift
+++ b/Sources/ArcGISToolkit/Common/CameraRequester.swift
@@ -29,8 +29,9 @@ import SwiftUI
         case .authorized:
             onAccessGranted()
         case .notDetermined:
-            AVCaptureDevice.requestAccess(for: .video) { granted in
-                if granted {
+            Task {
+                let isAuthorized = await AVCaptureDevice.requestAccess(for: .video)
+                if isAuthorized {
                     onAccessGranted()
                 } else {
                     onAccessDenied()
--- a/Sources/ArcGISToolkit/Common/CameraRequester.swift
+++ b/Sources/ArcGISToolkit/Common/CameraRequester.swift
@@ -23,19 +23,17 @@ import SwiftUI
     
     var onAccessDenied: (() -> Void)?
     
-    func request(onAccessGranted: @escaping () -> Void, onAccessDenied: @escaping () -> Void) {
+    func request(onAccessGranted: @escaping () -> Void, onAccessDenied: @escaping () -> Void) async {
         self.onAccessDenied = onAccessDenied
         switch AVCaptureDevice.authorizationStatus(for: .video) {
         case .authorized:
             onAccessGranted()
         case .notDetermined:
-            Task {
-                let isAuthorized = await AVCaptureDevice.requestAccess(for: .video)
-                if isAuthorized {
-                    onAccessGranted()
-                } else {
-                    onAccessDenied()
-                }
+            let isAuthorized = await AVCaptureDevice.requestAccess(for: .video)
+            if isAuthorized {
+                onAccessGranted()
+            } else {
+                onAccessDenied()
             }
         default:
             alertIsPresented = true
--- a/Sources/ArcGISToolkit/Common/CodeScanner.swift
+++ b/Sources/ArcGISToolkit/Common/CodeScanner.swift
@@ -46,10 +46,12 @@ struct CodeScanner: View {
         } else {
             Color.clear
                 .onAppear {
-                    cameraRequester.request {
-                        cameraAccessIsAuthorized = true
-                    } onAccessDenied: {
-                        isPresented = false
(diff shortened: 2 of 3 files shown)


Sources/ArcGISToolkit/Common/CameraRequester.swift
```diff
@@ -29,8 +29,9 @@ import SwiftUI
         case .authorized:
             onAccessGranted()
         case .notDetermined:
-            AVCaptureDevice.requestAccess(for: .video) { granted in
-                if granted {
+            Task {
+                let isAuthorized = await AVCaptureDevice.requestAccess(for: .video)
```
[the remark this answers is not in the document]
mhdostal: Ok, I made that change.

Sources/ArcGISToolkit/Common/CodeScanner.swift
```diff
@@ -46,10 +46,12 @@ struct CodeScanner: View {
         } else {
             Color.clear
                 .onAppear {
```
dfeinzimer, line 48: onAppear can be replaced with a `.task`:

```swift
Color.clear
    .task {
        await cameraRequester.request {
            cameraAccessIsAuthorized = true
        } onAccessDenied: {
            isPresented = false
        }
    }
    .cameraRequester(cameraRequester)
```

dfeinzimer (approved):

Sources/ArcGISToolkit/Common/CameraRequester.swift
```diff
@@ -23,18 +23,17 @@ import SwiftUI
     
     var onAccessDenied: (() -> Void)?
     
-    func request(onAccessGranted: @escaping () -> Void, onAccessDenied: @escaping () -> Void) {
+    func request(onAccessGranted: @escaping () -> Void, onAccessDenied: @escaping () -> Void) async {
```
rolson, line 26: I'm confused as to why this method is taking 2 closures. Since it's asynchronous can we not just return a boolean for `isAuthorized`?

mhdostal: > I'm confused as to why this method is taking 2 closures. Since it's asynchronous can we not just return a boolean for `isAuthorized`?

There are code paths in the `AVCaptureDevice.authorizationStatus` switch that do not specifically set `isAuthorized`. That's why it can't simply return a boolean.

rolson: One thing to consider is that isAuthorized is not a return value but a published property:

```swift
@Published private(set) var isAuthorized = false
```

I think that might simplify code looking at where the closures are used?

dfeinzimer: I investigated using the proposed published `isAuthorized` property. At face value it does seem like we could potentially simplify this:

```swift
await cameraRequester.request {
    cameraAccessIsAuthorized = true
} onAccessDenied: {
    isPresented = false
}
```
to this:
```swift
await cameraRequester.request()
if cameraRequester.isAuthorized {
    cameraAccessIsAuthorized = true
} else {
    isPresented = false
}
```

but as Mark noted there is a third possibility we have to account for where the `CameraRequester` itself needs to remain presented in order to present this alert (when we get [`AVAuthorizationStatus.denied`](https://developer.apple.com/documentation/avfoundation/avauthorizationstatus/denied) or [`AVAuthorizationStatus.restricted`](https://developer.apple.com/documentation/avfoundation/avauthorizationstatus/restricted) back):

<p align="center">
  
</p>

The current callback structure makes it simple to keep `CameraRequester` presented so that the user can respond to this alert. 

With this, my recommendation would be to keep the PR as is.

rolson: So maybe an isPublished Bool is not enough, but an enum might be?

dfeinzimer: An enum could work yes, but if we use [AVAuthorizationStatus](https://developer.apple.com/documentation/avfoundation/avauthorizationstatus) directly then the `CameraRequester` user has four possible outcomes to handle, rather than the current two.

Alternatively, we could use the proposed boolean property (or create our own simplified enumeration) and just set it `false` or `denied` once the user has pressed Cancel but this mean the developer now also needs to install a `.task(id: cameraRequester.isAuthorized) { ... }` elsewhere to monitor the value, scattering permission handling logic into more locations.

dfeinzimer:
> Approving. I feel like the closure api is a bit of a strange metaphor for SwiftUI in this case, but maybe I'm just not understanding the nuances here.

Agreed, I think the strangeness is rooted in the fact that the class has an async method and it's used in conjunction with a view modifier that may or may not ever need to present an alert. The good news is that this is all internal so I've opened #1077 so we can revisit this later on.

PR #1031, opened 2025-01-28
Add `FeatureFormGroupedContentView`

dfeinzimer:
If we proceed with the proposal to integrate `Form` into `FeatureFormView`, this view will not be needed and we can just use `Section { }` instead. 

In the meantime, this allows us to replicate what `Form { Section { } }` provides to match the proposed entry point for navigating associations in `UtilityNetworkAssociationFormElement`.

<p align="center">
  
</p>


mhdostal (approved):
Would this be incorporated after Zach's Form PR is done?

dfeinzimer:
> Would this be incorporated after Zach's Form PR is done?

No, the intention is to merge this right now to bring us in line with Peggy's work ASAP under the assumption that the `Form` PR is going to have a longer review process.

If we do proceed with incorporating `Form` into `FeatureFormView` this can just be reverted back out of the targeted feature branch.

Issue #1033, opened 2025-01-28
The today or now button should be aligned with the leading edge of the `DatePicker`


zkline101:
(Feedback from @PeggyDu09)

In Feature Forms, there is a button to get the current date (the "Today" button):

![Image](https://github.com/user-attachments/assets/e4b545b0-0a9c-4fc5-bbc9-cad24c22a58c)

It looks out of place and would look better if it was aligned with the calendar like this:

![Image](https://github.com/user-attachments/assets/f98979e6-bf85-4383-96db-db3292d72551)

I tried a few different ways to achieve this but I wasn't successful. SwiftUI won't let me move the `DatePicker` so both that and the button are aligned with the leading edge. Also, I don't know where the leading edge of the `DatePicker` is so I don't know where to move the button. The only solution I found was to add a large number of trailing padding to the button but that doesn't seem like an ideal solution.

Issue #1032, opened 2025-01-28
Give `GroupFormElement` a flat design


zkline101:
(Feedback from @PeggyDu09)
After the `Form` changes, I tried to achieve this but wasn't able to. Right now it looks like each group form element is grouped together when they should be separate.

![Image](https://github.com/user-attachments/assets/2c8eb817-b55f-4a11-a2eb-0e6653dda9d0)

When it would look better with a flat design like this:

![Image](https://github.com/user-attachments/assets/1307c121-95dd-422d-b1b8-5ca19666adb8)

So the goal would be to take out the background color from each disclosure group. I tried a few different ways to achieve this and none of them worked.

dfeinzimer:
@zkline101 I'm not sure I see the difference in the images above except for maybe the visibility of the Dividers and that the first uses two different font colors while the second uses only one.

My understanding is that the main issue with `Form` and `GroupFormElement` is that when the DisclosureGroup is expanded, there is no 2nd-level background color to connect each inner FieldFormElement visually. Is that a separate issue from this issue?

zkline101:
> @zkline101 I'm not sure I see the difference in the images above except for maybe the visibility of the Dividers and that the first uses two different font colors while the second uses only one.

Yeah I guess it is hard to tell from the pictures. But for the first picture there is a background behind the form and the header of each disclosure group has a separate background. The second picture only shows the background behind the form and the header for each disclosure group doesn't have background color. Does that help?

> My understanding is that the main issue with Form and GroupFormElement is that when the DisclosureGroup is expanded, there is no 2nd-level background color to connect each inner FieldFormElement visually. Is that a separate issue from this issue?

That sounds like a separate issue to me. The issue was surrounding the background color behind each disclosure group header.

dfeinzimer:
@zkline101, can this be closed? Groups received a revised appearance in #1283 with a good amount of input from Ryan. 

<img width="1916" height="2632" alt="Image" src="https://github.com/user-attachments/assets/336586a0-f732-449f-8ec9-e57ba5beb97f" />

zkline101:
Looks good to me. Closing.

PR #1034, opened 2025-01-28
Use `Form` in `FeatureFormView`

zkline101:
This allows us to use all the default formatting and styling that comes along with using a `Form` rather than adding all the behavior ourselves. Using a `Form` solves all the problems I was seeing when using this component in visionOS as well.

Another benefit I noticed is that a `Section` is considered lazy so it won't render until it is shown on the screen. Where before everything was rendered at once.

In addition, I crossed this by @PeggyDu09 and got her feedback. I addressed her feedback in this PR and made issues for the things that I couldn't get to work. I wouldn't consider either a show stopper. Here are the issues I made for them:
- https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/1032
- https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/1033

I tested this heavily on Mac Catalyst and iOS (more iPad than iPhone). But I highly recommend you test this on different platforms and data sets as well.

Last remaining task:
- Fix `FeatureFormViewTests.testCase_8_1()`


zkline101:
@mhdostal @dfeinzimer 

This is ready for review. I recommend testing feature forms with all the different types of datasets. I only have one more test to fix and I noted that in the PR description. But this can be reviewed in the meantime since it is a decent amount to get reviewed already.

Sources/ArcGISToolkit/Common/AttachmentsFeatureElementView.swift
```diff
@@ -67,18 +67,33 @@ struct AttachmentsFeatureElementView: View {
     @State private var isExpanded = true
     
     var body: some View {
-        Group {
+        Section {
```
zkline101, line 70: @mhdostal @dfeinzimer 

I had to make this change compared to the previous change I had because there was some strange behavior where the attachments weren't getting loaded and a cancellation error was getting thrown. So if I wrap this whole view in a Section that solves the problem and maybe this is the way SwiftUI expects it to be.

The main thing I wanted to point out was `AttachmentsFeatureElementView` is used for feature forms and popups so I made edits here to prevent any new obvious changes happening to popups.

zkline101:
I ran all the live UI tests and they all pass now.

Sources/ArcGISToolkit/Common/AttachmentsFeatureElementView.swift
```diff
@@ -67,18 +67,33 @@ struct AttachmentsFeatureElementView: View {
...
+                                    thumbnailSize: thumbnailSize
+                                )
+                            }
+                        attachmentModelsState = .initialized(attachmentModels)
+                    }
             case .initialized(let attachmentModels):
                 if isShowingAttachmentsFormElement {
                     // If showing a form element, don't show attachments in
                     // a disclosure group, but also ALWAYS show
-                    // the list of attachments, even if there are none.
```
dfeinzimer, line 79: Looks like this comment line was accidentally removed?

Sources/ArcGISToolkit/Common/AttachmentsFeatureElementView.swift
```diff
@@ -67,18 +67,33 @@ struct AttachmentsFeatureElementView: View {
...
-                    attachmentHeader
-                    attachmentBody(attachmentModels: attachmentModels)
+                        attachmentBody(attachmentModels: attachmentModels)
```
dfeinzimer, line 96: Please decrease the indentation here.

```suggestion
                    attachmentBody(attachmentModels: attachmentModels)
```

Sources/ArcGISToolkit/Common/AttachmentsFeatureElementView.swift
```diff
@@ -67,18 +67,33 @@ struct AttachmentsFeatureElementView: View {
...
     
     var body: some View {
-        Group {
+        Section {
             switch attachmentModelsState {
             case .notInitialized, .initializing:
                 ProgressView()
                     .padding()
+                    .task {
+                        guard case .notInitialized = attachmentModelsState else { return }
```
dfeinzimer, line 76: Since this task should always only run once before the view appears, I wonder if we need this guard. Theoretically, the condition should always be true.

I know this is also on `v.next`. It is probably an optimization that we initially missed, so this might be a good opportunity to fix it.

dfeinzimer (changes requested):

Sources/ArcGISToolkit/Common/AttachmentsFeatureElementView.swift
dfeinzimer: I wonder if the unexpected cancellation is related to the Form content being lazy loaded in some way.

Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView.swift
```diff
@@ -126,6 +129,20 @@ public struct FeatureFormView: View {
...
+                FullScreenTextInput(
+                    text: $model.fullScreenText,
+                    element: element,
+                    model: model
+                )
+#if targetEnvironment(macCatalyst)
+                .environmentObject(model)
+#endif
+            }
+        }
```
dfeinzimer, line 145: This is unfortunate. Maybe it would be worth putting this into an internal modifier to keep `evaluatedForm` as slim as possible by reducing this to 1 line.

Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView.swift
```diff
@@ -169,15 +186,13 @@ extension FeatureFormView {
...
-            Divider()
         }
     }
     
     /// Makes UI for a text form element including a divider beneath it.
     /// - Parameter element: The element to generate UI for.
     @ViewBuilder func makeTextElement(_ element: TextFormElement) -> some View {
         TextFormElementView(element: element)
-        Divider()
     }
```
dfeinzimer, line 196: Now that this method is a single line, perhaps it should be removed.

Examples/Examples/FeatureFormExampleView.swift
```diff
@@ -66,8 +66,6 @@ struct FeatureFormExampleView: View {
                     if let featureForm = model.featureForm {
                         FeatureFormView(featureForm: featureForm)
                             .validationErrors(validationErrorVisibility)
-                            .padding(.horizontal)
-                            .padding(.top, 16)
```
dfeinzimer, line 70: Please also remove these lines from `FeatureFormViewStep13.swift` onward to keep the tutorial in sync with the example.

Sources/ArcGISToolkit/Common/AttachmentsFeatureElementView.swift
zkline101: Yeah that was one of the things Phil and I discussed. It could be.

Sources/ArcGISToolkit/Common/AttachmentsFeatureElementView.swift
zkline101: Ok yeah it looks like it could be removed.

Sources/ArcGISToolkit/Common/AttachmentsFeatureElementView.swift
```diff
@@ -67,18 +67,33 @@ struct AttachmentsFeatureElementView: View {
...
+                                )
+                            }
+                        attachmentModelsState = .initialized(attachmentModels)
+                    }
             case .initialized(let attachmentModels):
                 if isShowingAttachmentsFormElement {
                     // If showing a form element, don't show attachments in
                     // a disclosure group, but also ALWAYS show
-                    // the list of attachments, even if there are none.
-                    attachmentHeader
```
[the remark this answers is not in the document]
zkline101: I see. Let me make some adjustments.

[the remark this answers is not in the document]
zkline101: The padding that is being added to the `DisclosureGroup` doesn't seem to be making a difference so I'm going to remove that too.

Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView.swift
```diff
@@ -99,22 +99,25 @@ public struct FeatureFormView: View {
...
-                        // case added for AttachmentsFormElement.
-                        AttachmentsFeatureElementView(featureElement: attachmentsElement)
-                    }
+            Form {
+                if !title.isEmpty && headerVisibility != .hidden {
+                    FormHeader(title: title)
+                        // Get rid of background color.
+                        .listRowBackground(Color.clear)
+                        // Make header align with the leading edge of the Form.
+                        .listRowInsets(.init())
```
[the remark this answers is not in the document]
zkline101: I'll move the comments. I think the background color comment can go but the row insets one I think is helpful.

Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView.swift
zkline101: Yeah it is a bit odd that this sheet modifier had to be on the `Form` to work properly while other sheets in feature form work properly. 

> Maybe it would be worth putting this into an internal modifier to keep evaluatedForm as slim as possible by reducing this to 1 line.

I'm not sure I see much benefit of that right now but if I can make the change if that is what you want.

zkline101: I'm not sure if a sheet modifier is something that should be hidden in a custom view modifier. It seems like an important detail. Thoughts?

Sources/ArcGISToolkit/Components/FeatureFormView/FormViewModel.swift
```diff
@@ -31,6 +31,12 @@ import SwiftUI
     /// The list of visible form elements.
     @Published var visibleElements = [FormElement]()
     
+    /// A Boolean value indicating whether the full screen text input is presented.
+    @Published var fullScreenTextInputIsPresented = false
+        
+    /// Text that will be displayed in a view showing multi line text.
+    @Published var fullScreenText = ""
```
dfeinzimer, line 38: This property is never written to. Currently, when the sheet is brought up, the element's current value is not injected.

<p align="center">
  
</p>

dfeinzimer (changes requested):
Overall the changes aren't too bad and I think for the most part the Form looks pretty great. I do still have some reservations:

### Group Form Elements

Take for example two consecutive Group Form Elements: 

<p align="center">
  
</p>

1.  Group 1 and Group 2 are two distinct Form Elements. Not being able to place them in a `Section` and therefore losing the separation we get between other Form Elements is rough.

2. When a group is expanded, it's hard to discern any visual connection to the header and its elements apart from the indentation.

1 and 2 together almosts inverting the whole purpose of groups. We're grouping headers where they should otherwise be separated and separating group members where they should be grouped.

### Scrolling

This is a smaller concern but Form Element headers are cutoff when they become the focused element by the auto-scroll behavior. This isn't an issue with the current `FeatureFormView`.

| Before | After |
|---|---|
|  |  |

_"What was the question asking again?"_ 😅 

---

The loss of fine-grained control we have over these things when we use more rudimentary elements like `VStack` makes me apprehensive for future things we may need to support in `FeatureFormView`, but, at present, the upsides of `Form` do seem promising.

It would be my preference gather feedback on these changes (perhaps from Apollo before proceeding with the current changes.)

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/TextInput.swift
```diff
@@ -201,37 +191,28 @@ private extension TextInput {
     }
 }
 
-@available(visionOS, unavailable)
-private extension TextInput {
```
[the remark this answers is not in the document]
zkline101: We don't have access to the `TextInput ` from the outer feature form view. So that is why I took it out.

Sources/ArcGISToolkit/Common/AttachmentsFeatureElementView.swift
```diff
@@ -67,48 +67,75 @@ struct AttachmentsFeatureElementView: View {
     @State private var isExpanded = true
     
     var body: some View {
-        Group {
+        Section {
             switch attachmentModelsState {
             case .notInitialized, .initializing:
                 ProgressView()
                     .padding()
+                    .task {
```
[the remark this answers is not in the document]
philium: I'm comment since this was my suggestion. Does it _need_ to change? No. But _should_ it change? Yes. The reason is that the task is not needed when the state is `initialized`. This is common practice and even done elsewhere (`FormView`, I think?).

zkline101:
This work has been called off for now. So closing out the PR, but I will keep the branch alive so we can reference it in the future.

dfeinzimer:
See also Apollo 332

PR #1038, opened 2025-01-29
Replace navigation stack with navigation split view

philium:
This puts the category and example lists in a sidebar, rather than taking up the whole window, on iPadOS, macOS, and visionOS. It also improves category and example selection, eliminates unnecessary types, and reduces use of deprecated APIs.


zkline101 (approved):
Looks good and I tested it on all platforms.

Examples/ExamplesApp/Examples.swift
```diff
@@ -15,21 +15,48 @@
...
+    /// The category selected by the user.
+    @State private var selectedCategory: Category?
+    /// The example selected by the user.
+    @State private var selectedExample: Example?
     
     var body: some View {
-        NavigationStack {
-            List(lists) { (list) in
-                NavigationLink(list.name, destination: list)
+        NavigationSplitView {
```
dfeinzimer, line 27: How would you feel about using [`init(columnVisibility:sidebar:detail:)`](https://developer.apple.com/documentation/swiftui/navigationsplitview/init(columnvisibility:sidebar:detail:)) here, specifying [`doubleColumn`](https://developer.apple.com/documentation/swiftui/navigationsplitviewvisibility/doublecolumn) as the starting `columnVisibility` value?

I feel it's a bit jarring to have the application open to a largely blank screen in portrait iPadOS and require the user's eye to wander up and find the column button in the top left.

philium: I'm fine with that.

philium: So I looked into this and the options aren't great. We could set the column visibility to `all` and just leave it, but then there's no show/hide sidebar button, which is really nice to have. We could use `all` initially then set it to `automatic` when an example is selected, but then the sidebar dismisses (in portrait orientation), which could be unexpected and undesirable. We could do that only when the first example is selected, but then the behavior would be inconsistent (depending on whether the first or a subsequent example is selected). I wish there was a way to programmatically show the sidebar, but I couldn't find one. I think what we have is not the best (specifically for iPad in portrait orientation), but the best of the alternatives.

dfeinzimer: Thank you for looking into it more, Phil.

dfeinzimer (approved):

Examples/ExamplesApp/Examples.swift
philium: You're welcome.

PR #1040, opened 2025-01-30
Make top level `Model` class private

philium:
This class is specific to a particular example, but, by being at the top level as non-private, it was exposed to all examples. This fixes that.

Note: The one declaration that uses the type is already private.


mhdostal (approved):
Thanks!

dfeinzimer (approved):

PR #1043, opened 2025-01-31
update notification for other job types

rolson:


des12437 (approved):
Tested notifications with on demand web map.

PR #1046, opened 2025-02-01
OfflineMapAreasView - OnDemand Cleanup

rolson:
- cancel jobs
- show corrupted jobs in the list so they can be deleted


Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandListItemView.swift
```diff
@@ -36,24 +36,21 @@ struct OnDemandListItemView: View {
...
             }
-        }
-        .contentShape(.rect)
-        .onTapGesture {
-            if model.status.isDownloaded {
-                metadataViewIsPresented = true
+            .contentShape(.rect)
+            .onTapGesture {
+                if model.status.isDownloaded {
+                    metadataViewIsPresented = true
```
des12437, line 49: I'm not sure why, the the metadata view will not present for downloaded on-demand map areas. When a map area job is cancelled and can be deleted, tapping anywhere on the list row rather than just the delete button removed the download.

des12437 (approved):

PR #1047, opened 2025-02-01
Adjust background on floating panel for visionOS

### LINKED ISSUE #1021: Refactor `UtilityNetworkTrace` so the default styling for each platform is used
Currently, if you look at `UtilityNetworkTrace` on visionOS it looks like this:
Image

And on iOS it looks like this:
Image

visionOS should have a darker background material (According to the [HIG](https://developer.apple.com/design/human-interface-guidelines/materials#visionOS)) for each row to help visually separate sections of the app but it currently doesn't by default because we are using custom views to build this component. One reason why this could be [this](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/1011#issuecomment-2599345093). It would be worth looking into this and doing any refactoring that would result in more of the default styling being used.

zkline101:
Closes #1021

The problem here was the background on the floating panel was somehow overriding some default styling on the Utility Network component. If you put the Utility Network component in a sheet you would see it looks as expected but it would look incorrect in a floating panel. This fixes that. This doesn't make a visual difference to the Popup Component example.

Utility Network Before:
Before

Utility Network After:
After

### DIFF
--- a/Sources/ArcGISToolkit/Components/FloatingPanel/FloatingPanel.swift
+++ b/Sources/ArcGISToolkit/Components/FloatingPanel/FloatingPanel.swift
@@ -73,7 +73,7 @@ struct FloatingPanel<Content>: View where Content: View {
             // Set frame width to infinity to prevent horizontal shrink on dismissal.
             .frame(maxWidth: .infinity)
 #if os(visionOS)
-            .background(.regularMaterial)
+            .glassBackgroundEffect()
             .compositingGroup()
 #else
             .background(backgroundColor)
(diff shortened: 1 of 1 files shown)


Sources/ArcGISToolkit/Components/FloatingPanel/FloatingPanel.swift
```diff
@@ -73,7 +73,7 @@ struct FloatingPanel<Content>: View where Content: View {
             // Set frame width to infinity to prevent horizontal shrink on dismissal.
             .frame(maxWidth: .infinity)
 #if os(visionOS)
-            .background(.regularMaterial)
+            .glassBackgroundEffect()
             .compositingGroup()
```
philium, line 77: Is this necessary? Methinks not, but I'm not entirely certain.

dfeinzimer (dismissed):
Looks good on the simulator.

Sources/ArcGISToolkit/Components/FloatingPanel/FloatingPanel.swift
zkline101: Yep if we remove `compositingGroup` and `clipShape` it looks like this:
Screenshot 2025-01-31 at 4 38 01 PM

philium (approved):

dfeinzimer (approved):

PR #1053, opened 2025-02-05
On Demand: Map area selector

des12437:
Implements a map area selector for an on-demand map area:


Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandMapAreaSelectorView.swift
```diff
@@ -0,0 +1,289 @@
...
+    
+    /// Updates the handle locations using the boudning rect.
+    private func updateHandles() {
+        topRight = CGPoint(x: boundingRect.maxX, y: boundingRect.minY)
+        topLeft = CGPoint(x: boundingRect.minX, y: boundingRect.minY)
+        bottomLeft = CGPoint(x: boundingRect.minX, y: boundingRect.maxY)
+        bottomRight = CGPoint(x: boundingRect.maxX, y: boundingRect.maxY)
+    }
+    
+    struct Handle: View {
```
rolson, line 216: doc

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandMapAreaSelectorView.swift
```diff
@@ -0,0 +1,289 @@
...
+        let position: CGPoint
+        let color: Color
+        let resize: (CGPoint) -> Void
+        @GestureState var gestureState: Progress = .started
+        enum Progress {
```
rolson, line 221: doc

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandMapAreaSelectorView.swift
```diff
@@ -0,0 +1,289 @@
...
+                            UISelectionFeedbackGenerator().selectionChanged()
+                        case .changed:
+                            resize(value.location)
+                        }
+                    }
+                )
+        }
+    }
+    
+    struct RoundedRectangleWithRoundedCorners: Shape {
```
rolson, line 244: doc

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandMapAreaSelectorView.swift
```diff
@@ -0,0 +1,289 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ArcGIS
+import SwiftUI
+
+struct OnDemandMapAreaSelectorView: View {
+    /// The map view proxy.
+    private let mapViewProxy: MapViewProxy
```
rolson, line 20: I think this view shouldn't need to know anything about the mapview proxy. It can simply report the selection and the parent view can calculate the envelope.

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandConfigurationView.swift
```diff
@@ -223,3 +252,26 @@ private struct BottomCard<Content: View, Background: ShapeStyle>: View {
...
     }
 }
+
+private extension UIImage {
+    /// Crops a UIImage to a certain CGRect of the screen's coordinates.
+    /// - Parameter rect: A CGRect in screen coordinates.
+    /// - Returns: The cropped image.
+    @MainActor
+    func crop(to rect: CGRect) -> UIImage? {
+        let scale = UIScreen.main.scale
```
rolson, line 262: Doesn't UIImage have a scale? IIRC we shouldn't have to get the scale from the screen. That could be different than the image's scale.

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandConfigurationView.swift
```diff
@@ -35,12 +35,30 @@ struct OnDemandConfigurationView: View {
...
     /// A Boolean value indicating if the download button is disabled.
     private var downloadIsDisabled: Bool {
-        visibleArea == nil
+        mapAreaSelection == nil
+    }
+    
+    /// The scale or the map area selector.
+    private let mapAreaScale = 0.8
+    
+    /// The scaled downn visible area for the map area selector.
```
rolson, line 55:
```suggestion
    /// The scaled down visible area for the map area selector.
```

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandConfigurationView.swift
```diff
@@ -35,12 +35,30 @@ struct OnDemandConfigurationView: View {
...
     @Environment(\.dismiss) private var dismiss
     
     /// A Boolean value indicating if the download button is disabled.
     private var downloadIsDisabled: Bool {
-        visibleArea == nil
+        mapAreaSelection == nil
+    }
+    
+    /// The scale or the map area selector.
+    private let mapAreaScale = 0.8
```
rolson, line 53: This could be static

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandConfigurationView.swift
```diff
@@ -35,12 +35,30 @@ struct OnDemandConfigurationView: View {
...
+    
+    /// The scaled downn visible area for the map area selector.
+    private var scaledVisibleArea: Envelope? {
```
rolson, line 56: I don't quite understand what this is for. Can you explain it please?

des12437: Removed.

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandMapAreaSelectorView.swift
des12437: Removed.

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandConfigurationView.swift
des12437: This property was removed.

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandConfigurationView.swift
des12437: UIImage has a scale, and it is different from the screen scale. The image should be cropped relative to the screen scale. For example, on my test device the UIImage scale is 1 and the screen scale is 3, and using the image scale does crop the right image.

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandMapAreaSelectorView.swift
```diff
@@ -0,0 +1,310 @@
...
+                    x: safeAreaInsets.leading,
+                    y: safeAreaInsets.top,
+                    width: geometry.size.width - safeAreaInsets.trailing - safeAreaInsets.leading,
+                    height: geometry.size.height - safeAreaInsets.bottom - safeAreaInsets.top
+                )
+                
+                maxRect = frame
+                    .insetBy(
+                        dx: frame.width * 0.1,
+                        dy: frame.height * 0.1
```
rolson, line 117: I'd prefer hard-coded point values here instead of a percentage because a percentage will give different values for the width and height inset.

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandMapAreaSelectorView.swift
```diff
@@ -0,0 +1,310 @@
...
+
+import ArcGIS
+import SwiftUI
+
+struct OnDemandMapAreaSelectorView: View {
+    /// The maximum size of the area selector view.
+    @State private var maxRect: CGRect = .zero
+    
+    /// A Binding to the CGRect of the selected map area.
+    @Binding var selectedMapRect: CGRect
```
rolson, line 23:
```suggestion
    @Binding var selectedRect: CGRect
```

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandMapAreaSelectorView.swift
```diff
@@ -0,0 +1,310 @@
...
+// limitations under the License.
+
+import ArcGIS
+import SwiftUI
+
+struct OnDemandMapAreaSelectorView: View {
+    /// The maximum size of the area selector view.
+    @State private var maxRect: CGRect = .zero
+    
+    /// A Binding to the CGRect of the selected map area.
```
rolson, line 22:
```suggestion
    /// A Binding to the CGRect of the selected area.
```

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandMapAreaSelectorView.swift
```diff
@@ -0,0 +1,310 @@
...
+    @Binding var selectedMapRect: CGRect
+    
+    /// The bounding rectangle for the area selector view.
+    @State private var boundingRect: CGRect = .zero
```
rolson, line 26: what is the difference between this and the selectedRect? If you look at the code in the branch I posted I believe there is only selectedRect. I'm not sure why there is a need for both.

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandMapAreaSelectorView.swift
```diff
@@ -0,0 +1,310 @@
...
+    
+    /// The inset rectangle for the area selector view.
+    @State private var insetRect: CGRect = .zero
```
rolson, line 29: The name of this and the doc are not descriptive of what this rect is actually for. It looks like it's not an "inset" rect but actually an "outset" rect that is for positioning the rounded corners. If that's the case, it probably doesn't need to exist as a state variable at all. And if it does need to exist as a state variable, then I think it should be called something that describes what it is for. Like handlesRect or something.

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandMapAreaSelectorView.swift
```diff
@@ -0,0 +1,310 @@
...
+    @State private var bottomRight: CGPoint = .zero
+    
+    /// The safe area insets of the view.
+    @State private var safeAreaInsets = EdgeInsets()
+    
+    /// The corner radius of the area selector view.
+    private let cordnerRadius: CGFloat = 16
+    
+    /// The drag point for a drag gesture.
+    private enum DragPoint {
```
rolson, line 50: This is more of a "handle kind" or "handle location" it seems to me.

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandMapAreaSelectorView.swift
```diff
@@ -0,0 +1,310 @@
...
+                                .frame(width: boundingRect.width, height: boundingRect.height)
+                                .position(boundingRect.center)
+                            
+                        }
+                    RoundedCorners(cornerRadius: cordnerRadius)
+                        .stroke(.ultraThickMaterial, style: StrokeStyle(lineWidth: 6, lineCap: .butt))
+                        .frame(width: insetRect.width, height: insetRect.height)
+                        .position(insetRect.center)
+                    
+                    ZStack {
```
rolson, line 71: what is this ZStack needed for?

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandMapAreaSelectorView.swift
```diff
@@ -0,0 +1,310 @@
...
+                    .frame(width: boundingRect.width, height: boundingRect.height)
+                    .position(boundingRect.center)
+                }
+                .ignoresSafeArea()
+                .allowsHitTesting(false)
+                .onAppear {
+                    boundingRect = maxRect
+                    insetRect = boundingRect.insetBy(dx: -2, dy: -2)
+                    updateHandles()
+                }
```
rolson, line 85: This shouldn't be needed.

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandMapAreaSelectorView.swift
```diff
@@ -0,0 +1,310 @@
...
+                        case .changed:
+                            resize(value.location)
+                        }
+                    }
+                )
+        }
+    }
+    
+    /// A view that displays rounded corners for a rectangle view.
+    struct RoundedCorners: Shape {
```
rolson, line 264: It's fine to leave it like this for now, but I would've expected the Handle view to draw the rounded corner. So the handle view would have been initialized with a HandleLocation and it would draw the corner based on its location. Instead of the handle being a clear view. Maybe something for a future PR.

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandMapAreaSelectorView.swift
```diff
@@ -0,0 +1,310 @@
...
+    
+    /// The drag point for a drag gesture.
+    private enum DragPoint {
+        case topLeft, topRight, bottomLeft, bottomRight
+    }
+    
+    var body: some View {
+        GeometryReader { geometry in
+            ZStack {
+                ZStack {
```
rolson, line 57: I'm not sure both of these ZStacks are needed. If you look at the branch `ryan/area-selector-rect` there is only one ZStack. It might be simpler to use one ZStack and apply a background or overlay where necessary just so that logically the view reads better.

That's why I had this background line commented out, because you might've wanted to apply a that black background.

```
        //.background(Color.black.opacity(0.1).ignoresSafeArea().allowsHitTesting(false))
```

rolson:
Can you give the stroke on these a rounded edge? 
image

ie:

```swift
/// A circular gauge progress view style.
struct GaugeProgressViewStyle: ProgressViewStyle {
    private var strokeStyle: StrokeStyle { .init(lineWidth: 3, lineCap: .round) }
    
    func makeBody(configuration: Configuration) -> some View {
        if let fractionCompleted = configuration.fractionCompleted {
            ZStack {
                Circle()
                    .stroke(Color(.systemGray5), style: strokeStyle)
...
```

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandMapAreaSelectorView.swift
des12437: `boundingRect` is now removed.

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandMapAreaSelectorView.swift
des12437: Removed.

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandMapAreaSelectorView.swift
des12437: The second ZStack contains the dark background, area selector rectangle, and handles, which is a view that covers the whole screen. If this view is not nested as it is, the view will capture all gestures so they will not pass through to the map view. If hit testing is disabled, the gestures still do not pass through to the map view.

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandConfigurationView.swift
rolson: This can be done later. It's fine for now.

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandMapAreaSelectorView.swift
rolson: did you try this?
```
 //.background(Color.black.opacity(0.1).ignoresSafeArea().allowsHitTesting(false))
```

des12437:
Here is a screenshot of the updated stroke style with a darker background color:

Screenshot 2025-02-07 at 11 39 30 AM

Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandMapAreaSelectorView.swift
des12437: Yes I tried that and the UI looks right, but the gestures do not pass through the map view due to this view:
```
Rectangle()
  .fill(.black.opacity(0.1))
  .reverseMask {
        Rectangle()
            .frame(width: selectedRect.width, height: selectedRect.height)
            .position(selectedRect.center)
    }
```

rolson (approved):

PR #1055, opened 2025-02-06
Add visionOS as a supported platform in Test Runner

zkline101:
Swift #6016


dfeinzimer (approved):
Changes look okay but I see at least one test (`BookmarksTests.testCase1()`) not passing on Vision that passes okay on iOS. 

Should we look into test failures as apart of this work?

zkline101:
All unit tests passed for me when I ran them. `BookmarksTests` is a live UI test and we won't be running those on visionOS.

PR #1056, opened 2025-02-07
Small `FeatureFormView` tweaks

dfeinzimer:
Makes some small improvements found during #1034 review.

### DIFF
--- a/Sources/ArcGISToolkit/Common/AttachmentsFeatureElementView.swift
+++ b/Sources/ArcGISToolkit/Common/AttachmentsFeatureElementView.swift
@@ -72,6 +72,22 @@ struct AttachmentsFeatureElementView: View {
             case .notInitialized, .initializing:
                 ProgressView()
                     .padding()
+                    .task {
+                        guard case .notInitialized = attachmentModelsState else { return }
+                        attachmentModelsState = .initializing
+                        let attachments = (try? await featureElement.featureAttachments) ?? []
+                        
+                        let attachmentModels = attachments
+                            .reversed()
+                            .map {
+                                AttachmentModel(
+                                    attachment: $0,
+                                    displayScale: displayScale,
+                                    thumbnailSize: thumbnailSize
+                                )
+                            }
+                        attachmentModelsState = .initialized(attachmentModels)
+                    }
             case .initialized(let attachmentModels):
                 if isShowingAttachmentsFormElement {
                     // If showing a form element, don't show attachments in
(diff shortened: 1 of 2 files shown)


Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/SwitchInput.swift
```diff
@@ -73,6 +73,9 @@ struct SwitchInput: View {
             .onChange(isOn) { isOn in
                 element.updateValue(isOn ? input.onValue.code : input.offValue.code)
                 model.evaluateExpressions()
+            }
+            .onTapGesture {
+                isOn.toggle()
```
dfeinzimer, line 78: `SwitchInput` needs to use a tap gesture to know when to set the focused element.

If we set the focusedElement during the onChange callback, the element is unexpectedly focused and scrolled to when a form first opens.

Sources/ArcGISToolkit/Common/AttachmentsFeatureElementView.swift
```diff
@@ -72,6 +72,20 @@ struct AttachmentsFeatureElementView: View {
             case .notInitialized, .initializing:
                 ProgressView()
                     .padding()
+                    .task {
```
dfeinzimer, line 75: Moving the task here was suggested in #1034 but that PR was closed.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/SwitchInput.swift
philium: Using a tap gesture on a switch (and setting the value bound to the switch in the handler) is extremely unusual. Is the unexpected focus and scroll behavior due to this:

https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/49a1b26be1ac612f3b78f0b273062dad60bd153f/Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView.swift#L114-L118

If so, the unintended behavior should be addressed at the source rather than hacking a switch like this.

dfeinzimer: Sorry, I meant unexpected as in unexpected _to the user_. They'll open a form and it'll be scrolled down to a switch and they won't know why.

What we want to avoid is the element setting itself as the focused element when this line runs for the first time:
https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/bcdf01233e1f98e14c068819bb756ba8555091e2/Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/SwitchInput.swift#L73

By using `onTapGesture` the element isn't set as the focused element in the model until the user taps on it.

My change here reverted #1048 to the previous implementation, but I will look for a better solution.

dfeinzimer: @philium I've committed another approach. This is ready for review again.

zkline101 (dismissed):

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/SwitchInput.swift
```diff
@@ -22,6 +22,9 @@ struct SwitchInput: View {
     /// The view model for the form.
     @EnvironmentObject var model: FormViewModel
     
+    /// A Boolean value indicating whether the initial element value was received.
+    @State private var didReceiveInitialValue = false
```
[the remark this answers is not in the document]
dfeinzimer: https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/1056/commits/76b3bf31bf9c5375856c0556eefd70450de4ae6d

zkline101 (approved):

mhdostal (approved):

PR #1057, opened 2025-02-07
OfflineMapAreas: Consolidate code for gauge progress view

rolson:
part of the work for UI design feedback


des12437 (approved):
Looks good. The `cancelGauge` progress view style is not applied to the preplanned list item view, but that can be done in a follow up PR to consolidate the list item views.

PR #1059, opened 2025-02-08
OfflineMapAreasView: design feedback round 2

rolson:
There is still more to do. But this takes care of the majority of the list view and metadata view.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreaItemView.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreaItemView.swift
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import SwiftUI
+import ArcGIS
 
 /// A view that shows information for an offline area for use in a List.
 @MainActor
@@ -125,6 +126,9 @@ protocol OfflineMapAreaListItemInfo: ObservableObject, OfflineMapAreaMetadata {
     var listItemDescription: String { get }
     var statusText: String { get }
     var statusSystemImage: String { get }
+    var jobProgress: Progress? { get }
+    
+    func cancelJob()
 }
 
 #Preview {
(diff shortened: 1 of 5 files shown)


Sources/ArcGISToolkit/Components/Offline/OfflineMapAreaMetadataView.swift
```diff
@@ -39,26 +39,20 @@ struct OfflineMapAreaMetadataView<Metadata: OfflineMapAreaMetadata>: View {
                 }
```
des12437, line 39:
```suggestion
                }
                .textCase(nil)
```
Suggested to remove upper case section title, as per the design.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreaMetadataView.swift
```diff
@@ -77,25 +71,36 @@ struct OfflineMapAreaMetadataView<Metadata: OfflineMapAreaMetadata>: View {
...
             }
+            .padding()
             
             Text(model.title)
                 .lineLimit(1)
-                .font(.largeTitle)
+                .font(.title2)
                 .fontWeight(.bold)
+            
+            Text("Size: \(model.directorySizeText)")
```
des12437, line 101: The size should only be shown if the status is downloaded, so "Size: Zero kB" does not show for areas not downloaded.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreaMetadataView.swift
```diff
@@ -117,38 +122,22 @@ protocol OfflineMapAreaMetadata: ObservableObject {
     var allowsDownload: Bool { get }
     /// The size of the area on disk.
     var directorySize: Int { get }
+    /// Whether the metadata view should be dismissed when the map area is deleted.
```
des12437, line 125:
```suggestion
    /// A Boolean value indicating whether the metadata view should be dismissed when the map area is deleted.
```

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreaItemView.swift
```diff
@@ -0,0 +1,236 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import SwiftUI
+import ArcGIS
+
+/// A view that shows information for an offline area for use in a List.
+@MainActor
+struct OfflineMapAreaItemView<Model: OfflineMapAreaListItemInfo, TrailingContent: View>: View {
```
des12437, line 20:
```suggestion
struct OfflineMapAreaListItemView<Model: OfflineMapAreaListItemInfo, TrailingContent: View>: View {
```
Can be renamed for consistency with `OnDemandListItemView` and `PreplannedListItemView`. Up to you.

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreaMetadataView.swift
rolson: I'll check with Peggy, but she probably doesn't know that uppercase is the default for section headers. I'd rather not mess with defaults.

des12437 (approved):

PR #1068, opened 2025-02-12
Offline: Ignore keyboard safe area

des12437:
Fixes a bug where the keyboard presentation causes the map view visible area to change.


rolson (approved):

dfeinzimer:
Related: Swift 1451

PR #1069, opened 2025-02-12
Offline: fix issue with updating safe area insets

rolson:


des12437 (approved):
Looks good. Fixes bottom sheet drag gesture bug.

PR #1075, opened 2025-02-15
Offline: Add on demand map model tests

des12437:
Adds tests for the `OnDemandMapModel`. 
   - Test suite runs in ~100 seconds due to test case that downloads mmpk.

### DIFF
--- a/Tests/ArcGISToolkitTests/OnDemandMapModelTests.swift
+++ b/Tests/ArcGISToolkitTests/OnDemandMapModelTests.swift
@@ -47,6 +47,10 @@ class OnDemandMapModelTests: XCTestCase {
         let areaID = OnDemandMapModel.makeAreaID()
         let mmpkDirectory = URL.onDemandDirectory(forPortalItemID: portalItemID, onDemandMapAreaID: areaID)
         
+        let job = try await OnDemandMapModel.makeJob(for: mmpkDirectory, task: offlineMapTask)
+        
+        OfflineManager.shared.jobManager.jobs.append(job)
+        
         defer {
             // Clean up JobManager.
             OfflineManager.shared.jobManager.jobs.removeAll()
@@ -54,10 +58,6 @@ class OnDemandMapModelTests: XCTestCase {
             // Clean up folder.
             try? FileManager.default.removeItem(at: mmpkDirectory)
         }
-        
-        let job = try await OnDemandMapModel.makeJob(for: mmpkDirectory, task: offlineMapTask)
-        
-        OfflineManager.shared.jobManager.jobs.append(job)
         
         let model = OnDemandMapModel(
             job: job,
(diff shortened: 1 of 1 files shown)


Tests/ArcGISToolkitTests/OnDemandMapModelTests.swift
```diff
@@ -0,0 +1,353 @@
...
+            areaID: areaID,
+            portalItemID: portalItemID,
+            onRemoveDownload: { _ in }
+        )
+        XCTAssertNil(model)
+        
+        let directory = URL.onDemandDirectory(forPortalItemID: portalItemID, onDemandMapAreaID: areaID)
+        let mmpkDirectory = OnDemandMapModel.mmpkDirectory(forOnDemandDirectory: directory)
+        try FileManager.default.createDirectory(at: mmpkDirectory, withIntermediateDirectories: true)
+        XCTAssertTrue(FileManager.default.fileExists(atPath: mmpkDirectory.path()))
```
rolson, line 107: please comment this block and others like it.

Tests/ArcGISToolkitTests/OnDemandMapModelTests.swift
```diff
@@ -0,0 +1,353 @@
...
+        XCTAssertNotEqual(downloadedModel.status, .initialized)
+        
+        // Clean up folder.
+        downloadedModel.removeDownloadedArea()
+        
+        // Verify statuses after remove.
+        // First give time for final status to come in.
+        try? await Task.yield(timeout: 0.1) { @MainActor in
+            statuses.last == .initialized
+        }
```
rolson, line 304: This shouldn't be necessary to wait like this here. Is it?

Tests/ArcGISToolkitTests/OnDemandMapModelTests.swift
```diff
@@ -0,0 +1,353 @@
...
+}
+
+extension OnDemandMapModel.Status: Equatable {
+    public static func == (lhs: OnDemandMapModel.Status, rhs: OnDemandMapModel.Status) -> Bool {
+        return switch (lhs, rhs) {
+        case (.initialized, .initialized),
+            (.downloading, .downloading),
+            (.downloaded, .downloaded),
+            (.downloadFailure, .downloadFailure),
+            (.mmpkLoadFailure, .mmpkLoadFailure):
```
rolson, line 316: Just FYI - This violates the equatable contract because it doesn't consider the error. But I'm not sure there is a better way to do this.

des12437: Good point. These cases can be removed from the switch statement and return false by default without impacting the tests.

Tests/ArcGISToolkitTests/OnDemandMapModelTests.swift
des12437: This wait is necessary for the status to reach `initialized`.

rolson (approved):

PR #1084, opened 2025-02-25
Add `NavigationLayer`

dfeinzimer:
Proposes `NavigationLayer`, an internal `NavigationStack` alternative to work around the [nested `NavigationStack` limitation](https://developer.apple.com/forums/thread/772650).

<p align="center">
  
</p>

### DIFF
--- a/Sources/ArcGISToolkit/Common/NavigationLayer/NavigationLayer+DestinationHeader.swift
+++ b/Sources/ArcGISToolkit/Common/NavigationLayer/NavigationLayer+DestinationHeader.swift
@@ -43,6 +43,7 @@ extension NavigationLayer {
                         if let subtitle = model.presented?.subtitle {
                             Text(subtitle)
                                 .font(.caption)
+                                .foregroundStyle(.secondary)
                         }
                     }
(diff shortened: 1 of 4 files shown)


Sources/ArcGISToolkit/Common/NavigationLayer/NavigationLayer+DestinationHeader.swift
```diff
@@ -0,0 +1,57 @@
...
+                }
+                if let title = model.presented?.title {
+                    Divider()
+                        .frame(height: size)
+                    VStack(alignment: .leading) {
+                        Text(title)
+                            .bold()
+                        if let subtitle = model.presented?.subtitle {
+                            Text(subtitle)
+                                .font(.caption)
```
mhdostal, line 45: This may depend on the design, but do we want the subtitle in the `.secondary` color?
```suggestion
                                .font(.caption)
                                .foregroundStyle(.secondary)
```

mhdostal (approved):
Just the one comment, otherwise it looks great! Much simpler than I had anticipated.

PR #1088, opened 2025-02-28
`FeatureFormView` - View utility network associations

dfeinzimer:
Apollo 1014

## ⚪ Overview:
- 🟢  Additions:
  - Support for utility association browsing via `FeatureFormView`.
  - The `FeatureFormView(featureForm: Binding<FeatureForm?>)` initializer.
  - The `FeatureFormView.closeButton(_: Visibility)` modifier.
  - The `FeatureFormView.editingButtons(_: Visibility)` modifier.
  - The `FeatureFormView.onFormEditingEvent(perform:)` modifier.
  - The `FeatureFormView.EditingEvent` enum.
- 🟡 Deprecations:
  - The `FeatureFormView(featureForm: FeatureForm)` initializer.
  - The `FeatureFormView.validationErrors(_:)` modifier.
  - Deprecates the `ValidationErrorVisibility` enum.


mhdostal (approved):

mhdostal (approved):

mhdostal (approved):

CalebRas:
I was prototyping these changes in `PopupView` and found that using padding outside of the `NavigationLayer` causes some animation issues (watch leading and trailing edges in video). 

https://github.com/user-attachments/assets/bf1db1a7-0478-41e4-bad8-f5a4284244d1



One possible solution would be to move the navigation layer up to wrap the [`FloatingPanel` content](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/v.next/Sources/ArcGISToolkit/Components/FloatingPanel/FloatingPanel.swift#L63), which would then make the navigation layer work more like a `NavigationStack` and also allow any component inside of a floating panel to use navigation from the environment. That solution would probably require a decent amount of refactoring, though, and special handling for when a component view (FeatureFormView, PopupView, etc.) is not used inside a floating panel, so I am leaning toward leaving your implementation as is.

The downside to that would be that padding (and possibly other views) would have to be removed from the component view call sites. In the case of PopupView, adding padding appears to be a very common workflow, since it is done [in the Examples](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/v.next/Examples/Examples/PopupExampleView.swift#L73) and everywhere it is used in the Samples. As a result, using the current navigation layer implementation in PopupView would likely break a lot of user implementations. 

I figured I would bring this up as something to consider while the NavigationLayer is still in development.

PR #1090, opened 2025-03-03
Update Localizable.strings

des12437:
Generated `Localizable.strings` file and some strings are reordered.


Sources/ArcGISToolkit/Resources/en.lproj/Localizable.strings
```diff
@@ -165,7 +166,7 @@ east, etc.). */
 /* A label for a button to delete a map area. */
 "Delete Map Area" = "Delete Map Area";
 
-/* A label for the description of the map area. */
```
dfeinzimer, line 168: Other similar comments have a period. Was this accidental?

des12437: Yes, the period is now added back.

dfeinzimer (approved):

Issue #1092, opened 2025-03-04
App freezes when trying to dismiss a popup in a sheet


zkline101:
I have an iOS app that is showing a popup in a sheet and when I try to dismiss that sheet by swiping down the whole app freezes and I see the memory starts to increase without stopping.

I have a reproducer for this that I can send if that helps.

zkline101:
I am using an iOS 18.2 simulator, Xcode 16.2 and I am using the Toolkit branch `OfflineMapAreasView` and I am using 200.6 of the SDK.

zkline101:
In my app, if I remove the scale bar from the app then this problem doesn't occur. This could be an iOS bug but not sure. It seems like if I have some toolkit components overlaid on my map then I show a popup in a sheet and I try to swipe to dismiss the sheet then it freezes.

rolson:
This is bad bug affecting the epic app and other dev summit demos.

dfeinzimer:
This seems possibly related to #802.

Another workaround would be to remove the Scalebar from the hierarchy while the sheet is presented.

```swift
.overlay {
    if !sheetIsPresented {
        Scalebar(
            ...
        )
    }
}
```

rolson:
I tried that in my app and it never came back... strange. Yes, probably best to just remove scalebar.

cc @nixta

dfeinzimer:
Well, this is weird. Using `overlay(alignment:content:)` vs `overlay(content:)` seems to make all the difference. The alignment is throwing something off somehow. I've distilled the reproducer down to this:

```swift
import ArcGIS
import ArcGISToolkit
import SwiftUI

struct PopupExampleView: View {
    let map = Map(basemapStyle: .arcGISCommunity)
    
    @State private var showPopup = false
    
    @State private var spatialReference: SpatialReference?
    @State private var unitsPerPoint: Double?
    @State private var viewpoint: Viewpoint?
    
    var body: some View {
        MapView(map: map)
            .onSpatialReferenceChanged { spatialReference = $0 }
            .onUnitsPerPointChanged { unitsPerPoint = $0 }
            .onViewpointChanged(kind: .centerAndScale) { viewpoint = $0 }
            .onSingleTapGesture { _, _ in
                showPopup = true
            }
            .overlay {                          // <=== NO FREEZE
//            .overlay(alignment: .bottomLeading) { // <=== FREEZES
                Scalebar(
                    maxWidth: 175,
[...]

PR #1095, opened 2025-03-08
fix scalebar freeze

rolson:
fixes https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/1092

This seems to fix the freeze. And it seems that the code was reading the height and then setting the height to what was read.


mhdostal (approved):
Looks good. All the tests pass.

Thanks for digging into this some more!

dfeinzimer (approved):

PR #1097, opened 2025-03-17
Offline: Persist map area configuration

des12437:
Closes swift/6631.

Persists map area configuration data such as title and extent when app is backgrounded.


rolson (commented):
This turns a simple struct that is meant as the return value into a view model object. That seems to violate the [single responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle).

PR #1103, opened 2025-03-18
Offline: Done button visibility

des12437:
Adds support to set the visibility of the "Done" button with the `doneButton(Visibility)` view modifier on the `OfflineMapAreasView`. This allows various types of navigation to be supported.

Closes `swift/6471`.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandListItemView.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandListItemView.swift
@@ -32,6 +32,9 @@ struct OnDemandListItemView: View {
         selectedMap?.item?.title == model.title
     }
     
+    /// A Boolean value indicating whether the view should dismiss.
+    var shouldDismiss = true
+    
     var body: some View {
         OfflineMapAreaListItemView(model: model, isSelected: isSelected) {
             trailingButton
@@ -47,7 +50,7 @@ struct OnDemandListItemView: View {
                 selectedMap: $selectedMap,
                 map: model.map,
                 isSelected: isSelected,
-                dismiss: dismiss
+                dismiss: shouldDismiss ? dismiss : nil
             )
         case .initialized:
             EmptyView()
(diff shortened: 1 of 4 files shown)


Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandListItemView.swift
```diff
@@ -32,6 +32,9 @@ struct OnDemandListItemView: View {
         selectedMap?.item?.title == model.title
     }
     
+    /// A Boolean value indicating whether the view should dismiss.
+    var shouldDismiss = true
```
rolson, line 36: It would be better to not have multiple places where the default value exists, thus requiring passing it to the constructor of this view. Same for the other view.

```suggestion
    /// A Boolean value indicating whether the view should dismiss.
    let shouldDismiss: Bool
```

rolson (dismissed):

Sources/ArcGISToolkit/Components/Offline/Preplanned/PreplannedListItemView.swift
```diff
@@ -59,7 +62,7 @@ struct PreplannedListItemView: View {
                 selectedMap: $selectedMap,
                 map: model.map,
                 isSelected: isSelected,
-                dismiss: dismiss
+                dismiss: shouldDismiss ? dismiss : nil
```
dfeinzimer, line 65: Is it intended that the visibility of the Done button changes the behavior of the Open button? If Done is hidden, it seems the user is stuck on `PreplannedListItemView`?

<p align="center">
  
</p>

It seems like Open should either continue to always dismiss or, at the least, dismiss when Done is hidden.

des12437: Yes, this is intentional. When the visibility is `hidden` then the "Done" button does not show and the open button does not dismiss. This is to support a modal presentation navigation workflow. The view can be dismissed at the app-level when the selectedMap value changes if needed.

For example, in this demo app the "Done" button is hidden since the view is presented as a view pushed onto the navigation stack and the back button is shown. When the "open" button is tapped, the view does not dismiss, otherwise the view would be popped off of the navigation stack.


https://github.com/user-attachments/assets/20e9326a-6205-4685-a400-53fa203bc81d

dfeinzimer: Got it, thank you for confirming!

rolson (approved):

PR #1128, opened 2025-04-01
Fix intermittent crash accessing credential in Example

des12437:
Fixes a crash in the `UtilityNetworkTraceExampleView` where accessing the credential causes a crash when it is `nil`.

### DIFF
--- a/Examples/Examples/UtilityNetworkTraceExampleView.swift
+++ b/Examples/Examples/UtilityNetworkTraceExampleView.swift
@@ -37,6 +37,13 @@ struct UtilityNetworkTraceExampleView: View {
     /// A container for graphical trace results.
     @State private var resultGraphicsOverlay = GraphicsOverlay()
     
+    init() {
+        Task {
+            let publicSample = try? await ArcGISCredential.publicSample
+            ArcGISEnvironment.authenticationManager.arcGISCredentialStore.add(publicSample!)
+        }
+    }
+    
     var body: some View {
         GeometryReader { geometryProxy in
             MapViewReader { mapViewProxy in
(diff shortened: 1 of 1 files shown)


yo1995 (dismissed):
This fix is OK, but doesn't explain the root cause. If you replace the task modifier with the following

```swift
.task {
    do {
        let publicSample = try await ArcGISCredential.publicSample
        ArcGISEnvironment.authenticationManager.arcGISCredentialStore.add(publicSample)
    } catch {
        print("\(error.localizedDescription)")
    }
    print("finished")
}
```

You'll see that the error is a `Swift.CancellationError` and "finished" gets printed twice.

It seems some view in the hierarchy gets computed twice very quickly, which caused the task to run twice, thus the first task is immediately cancelled. I tinker with it a bit but couldn't find which one is causing this. 🤔

dfeinzimer (commented):
Since it's a legitimate force-unwrap crash and the examples aim to demonstrate best practices my preference would be to keep that task where it is, catch and emit the error, and note the known issue:

```swift
.task {
    do {
        let publicSample = try await ArcGISCredential.publicSample
        ArcGISEnvironment.authenticationManager.arcGISCredentialStore.add(publicSample)
    } catch {
        // See Also: https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/1129
        print("Error creating credential:", error.localizedDescription)
    }
}
```

@yo1995, @des12437, thoughts?

yo1995:
I like this approach 👍

des12437:
> catch and emit the error, and note the known issue

This prevents the credential crash from occurring when the error is caught:
```
Error creating credential: The operation couldn’t be completed. (Swift.CancellationError error 1.)
```
 Just to note, the example still works when the error is caught because the task runs twice as described in the related issue.

yo1995 (dismissed):

dfeinzimer (commented):
Thanks @des12437, can you please apply the changes to `UtilityNetworkTraceStep2.swift`, `UtilityNetworkTraceStep3.swift`, and `UtilityNetworkTraceStep4.swift` too please?

You can probably leave out the known issue reference since that is Examples app dependent.

dfeinzimer (approved):
Thank you!

yo1995 (approved):

PR #1144, opened 2025-04-10
offline: Reset status when preplanned job is canceled

yo1995:
## Description

While I'm writing design doc, I noticed a difference between `PreplannedMapModel` and `OnDemandMapModel`: on-demand shows a `downloadCancelled` status whereas preplanned doesn't. It might be a small oversight for parity.

In this PR I added the following:

- ~~a `downloadCancelled` status to [`PreplannedMapModel.Status`](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/main/Sources/ArcGISToolkit/Components/Offline/Preplanned/PreplannedMapModel.swift#L235) to match its counterpart in [`OnDemandMapModel.Status`](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/main/Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandMapModel.swift#L337)~~
- Logics to reset the status to `packaged` when a preplanned download is cancelled
- the status to various convenience Booleans
    - a difference between preplanned and on-demand is that, if an on-demand job is cancelled, it has to be removed and added again; whereas a preplanned job can start again after being cancelled. This is reflected in the `allowsDownload` Boolean.
- the unit tests for both statuses that were missing

## Questions

- ✔️  Was `PreplannedMapModel.Status.downloadCancelled` left out intentionally? If so I can revert the changes and only add the missing unit tests.
- ❌  Should a cancelled on-demand area job allow being downloaded again?

### DIFF
--- a/Sources/ArcGISToolkit/Components/Offline/Preplanned/PreplannedListItemView.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/Preplanned/PreplannedListItemView.swift
@@ -50,7 +50,7 @@ struct PreplannedListItemView: View {
             EmptyView()
         case .loading:
             ProgressView()
-        case .packaged, .downloadFailure, .downloadCancelled:
+        case .packaged, .downloadFailure:
             DownloadOfflineMapAreaButton(model: model)
         case .downloading:
             OfflineJobProgressView(model: model)
@@ -123,13 +123,12 @@ extension PreplannedMapModel: OfflineMapAreaListItemInfo {
         case .downloading: .downloading
         case .downloaded: .downloaded
         case .downloadFailure: .downloadFailed
-        case .downloadCancelled: .cancelled
         }
     }
     
     var statusSystemImage: String {
         switch status {
-        case .notLoaded, .loading, .packaged, .downloaded, .downloading, .downloadCancelled:
+        case .notLoaded, .loading, .packaged, .downloaded, .downloading:
             ""
         case .loadFailure, .mmpkLoadFailure, .downloadFailure:
             "exclamationmark.circle"
@@ -165,14 +164,6 @@ private extension LocalizedStringResource {
             "Packaging failed",
             bundle: .toolkit,
             comment: "The status text when a map area packaging failed."
-        )
-    }
-    
-    static var cancelled: Self {
-        .init(
-            "Cancelled",
-            bundle: .toolkit,
-            comment: "The status text when a map area download is cancelled."
         )
     }
 }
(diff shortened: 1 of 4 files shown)


Sources/ArcGISToolkit/Components/Offline/Preplanned/PreplannedMapModel.swift
```diff
@@ -215,7 +215,13 @@ class PreplannedMapModel: ObservableObject, Identifiable {
         case .success:
             status = .downloaded
         case .failure(let error):
-            status = .downloadFailure(error)
+            if error is CancellationError {
```
yo1995, line 218: This is modeled after https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/main/Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandMapModel.swift#L266

Sources/ArcGISToolkit/Components/Offline/Preplanned/PreplannedListItemView.swift
```diff
@@ -166,4 +167,12 @@ private extension LocalizedStringResource {
             comment: "The status text when a map area packaging failed."
         )
     }
+    
+    static var cancelled: Self {
```
yo1995, line 171: I'll move this extension to `LocalizedStringResource` as it is used by both models, if we agree to add the `.downloadCancelled` case.

des12437 (approved):
Changes look good.

The `downloadCancelled` case should be used by the preplanned model so that when a download is cancelled the status text "Cancelled" is displayed in the UI rather than "Download failed".

rolson:
If you cancel a preplanned map area it should just go back to "Ready to download" not "Download failed" or "Download cancelled".

It should look like this after canceling:

image


Not like this:
image

rolson (commented):
I don't think we need download cancelled for preplanned. It should just go back to the ready state.

rolson:
> Should a cancelled on-demand area job allow being downloaded again?

I think it's worth considering as an enhancement. Maybe create an issue for tracking and asking if it should be done. If it should be done, then it would only be possible to do (1) with some sort of UI change because we would need to add a button to re-start the download and we need to keep the "x" button, and (2) it could only be done while the area was still listed. Because after you cancel an area and then close and re-open the form, those cancelled/failed areas are cleaned up. This is by design so that failed areas don't build up and we don't need to serialize the information for the failed areas.

So it wouldn't be trivial to do allow it to be downloaded again and I'm not sure it's something people need to be able to do.

yo1995:
OK. I'll log the on-demand cancel-retry idea in the potential future improvements doc, instead of creating an issue, as it seems unlikely to be a feature someone wants.

des12437 (approved):

rolson (approved):

rolson:
Good fix 👍

PR #1145, opened 2025-04-11
Remove iOS 16 support

dfeinzimer:
Swift 6777


dfeinzimer:
@philium, @mhdostal please feel free to merge when ready.

Sources/ArcGISToolkit/Components/Augmented Reality/FlyoverSceneView.swift
```diff
@@ -155,10 +155,10 @@ public struct FlyoverSceneView: View {
                     )
                 }
 #endif
-                .onChange(initialCamera) { initialCamera in
+                .onChange(of: initialCamera) { _, initialCamera in
```
mhdostal, line 158: @philium - is it necessary to include the new value in this line or can the code block just access the `@State` value directly, like below. I noticed this is what you're doing in the samples [PR](https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/571).
```suggestion
                .onChange(of: initialCamera) {
```

mhdostal: The same question would apply to most of the `.onChange(of: ...)` changes.

philium: According to the doc, values captured by the closure will have their new value, so yes, the state value can be used directly (i.e. we can use the `onChange` variant whose `action` function does not have any parameters.

mhdostal (approved):
Thanks @philium

yo1995 (approved):
LGTM

PR #1067, opened 2025-02-12
Offline: Try to prevent sheet from moving when interacting with the MapView.

rolson:
This seems to help with the pinching problem.

PR #1148, opened 2025-04-17
offline: Reorder the rotate gesture priorities on config view

yo1995:
- On iOS 17 simulator, with the `RotateGesture`, it would cause the sheet to move when panning the map view. Reordering the priority gestures fixes the problem.
- On iOS 18 simulator, with or without the change, the sheet doesn't move.

This was introduced in #1067. ~~However, I couldn't figure out what the pinching problem was referring to. Let me know if there are other considerations.~~

test map: 3da658f2492f4cfd8494970ef489d2c5

https://github.com/user-attachments/assets/cf94e8e9-b491-4004-a2d9-1f7979127844

### DIFF
--- a/Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandConfigurationView.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandConfigurationView.swift
@@ -164,7 +164,6 @@ struct OnDemandConfigurationView: View {
                 }
             }
             .highPriorityGesture(DragGesture())
-            .highPriorityGesture(RotateGesture())
             .interactiveDismissDisabled()
     }
     
--- a/Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandConfigurationView.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandConfigurationView.swift
@@ -163,6 +163,7 @@ struct OnDemandConfigurationView: View {
                     mapIsReady = true
                 }
             }
+            .highPriorityGesture(RotateGesture())
             .highPriorityGesture(DragGesture())
             .interactiveDismissDisabled()
     }


rolson:
> However, I couldn't figure out what the pinching problem was referring to.

I would recommend asking about things like this in the future. Because I could have told you what to test specifically.

I went back and tested this and I think you're right it doesn't need to be there. The behavior is strange. If you pinch first and then two finger drag, it doesn't pull the sheet down. If you don't pinch first and just do a 2 finger drag it pulls the sheet down. The difference of behavior was enough for me to think that the `.highPriorityGesture(RotateGesture())` had alleviated the issue with the sheet pulling down. That line can probably be removed.

rolson (dismissed):

des12437 (dismissed):
Change looks good. I tested this on iPhone and iPad.

yo1995:
I've done a bit more testing, here are the results…

1. Original code

```swift
            .highPriorityGesture(DragGesture())
            .highPriorityGesture(RotateGesture())
            .interactiveDismissDisabled()
```

iOS 17:

- one finger drag: buggy
- two finger drag: buggy
- rotate then drag: OK

iOS 18:

- one finger drag: OK
- two finger drag: buggy
- rotate then drag: OK

2. Proposed fix

```swift
            .highPriorityGesture(DragGesture())
            .interactiveDismissDisabled()
```

iOS 17:

- one finger drag: OK
- two finger drag: buggy
- rotate then drag: buggy

iOS 18:

- one finger drag: OK
- two finger drag: buggy
- rotate then drag: buggy

3. Reverse order

```swift
            .highPriorityGesture(RotateGesture())
            .highPriorityGesture(DragGesture())
            .interactiveDismissDisabled()
```

iOS 17:

- one finger drag: OK
- two finger drag: OK
- rotate then drag: OK

iOS 18:

- one finger drag: OK
- two finger drag: buggy
- rotate then drag: OK

rolson (approved):
Tested on iPhone with iOS 18, I see no difference in behavior between original code and the code in this PR.

des12437 (approved):

PR #1150, opened 2025-04-18
hotfix: 200.7.1 Release

yo1995:
# Description

#1149 fixes a bug that stops the on-demand workflow from downloading an area without moving the extent selector. Please test with an API key and the breweries web map 3da658f2492f4cfd8494970ef489d2c5 .

After these cherry-picked changes are reviewed, I'll figure out what version numbers need to be updated.
- edit: no example app version needs to be bumped as there aren't changes in them. They will pick up the latest toolkit once this patch is tagged.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandConfigurationView.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/OnDemand/OnDemandConfigurationView.swift
@@ -32,14 +32,8 @@ struct OnDemandConfigurationView: View {
     /// The max scale of the map to take offline.
     @State private var maxScale: CacheScale = .street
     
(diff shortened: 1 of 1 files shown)


rolson (approved):

yo1995:
I realized that the version numbers for example apps don't need to be bumped as there aren't changes in them.

Also, all the example apps are configured with `.upToNextMinor(from: "200.7.0")` so the change will be automatically picked up once we tag the 200.7.1 patch.

mhdostal (approved):

des12437 (approved):

yo1995:
Thanks. After chatting with folks it seems the impact is minimal, so I'll just tag the release.

PR #1151, opened 2025-04-22
Fix: Refactor localized strings

des12437:
Related to `swift/6790`

- Change some localized strings from type String to type LocalizedStringResource
- Use localized Button.cancel, Button.cancel, Button.ok where applicable
- Remove redundant localized string for "Cancel"

### DIFF
--- a/Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
+++ b/Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
@@ -143,7 +143,7 @@ struct SiteAndFacilitySelector: View {
             .background(.quinary)
             .clipShape(.rect(cornerRadius: 10))
             if textFieldIsFocused {
-                Button(String.cancel) {
+                Button.cancel {
                     query.removeAll()
                     textFieldIsFocused = false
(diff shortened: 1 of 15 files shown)


Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
```diff
@@ -143,7 +143,7 @@ struct SiteAndFacilitySelector: View {
             .background(.quinary)
             .clipShape(.rect(cornerRadius: 10))
             if textFieldIsFocused {
-                Button(String.cancel) {
+                Button.cancel {
```
des12437, line 146: This change applies the `.cancel` role to the button, which was not previously applied.  @dfeinzimer Should this change be reverted?

philium (approved):

Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
dfeinzimer: I don't see any issue with the change.

dfeinzimer (approved):

PR #1163, opened 2025-05-10
Add `FloatingPanelDetent.Preference`

dfeinzimer:
In certain situations, a view presented in a floating panel needs the map view to be revealed for user interaction:

<p align="center">
  
</p>

Typically, this would require the presented view to have a binding to the floating panel's detent; however, the view might not be presented in a floating panel. By providing this internal preference, Toolkit components can communicate a preferred detent to the Floating Panel without developers needing to write _any_ code. [Example usage](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/1161/commits/55d6a7d8b99cb2e6efac0aab49bf50f546ec5a01).

---

 - [ ] Testing
 - [ ] Can [`UtilityNetworkTrace.floatingPanelDetent(_:)`](https://developers.arcgis.com/swift/toolkit-api-reference/documentation/arcgistoolkit/utilitynetworktrace/floatingpaneldetent(_:)) be deprecated?
 - [ ] Should this be a public type?
 - [ ] Does this take too much control away from the developer?


mhdostal (dismissed):
this looks good and is a nice way to override the detentions when a component needs to.

dfeinzimer:
A reservation I have with this approach is that supporting the same behavior for sheets seems problematic. Sheets (surprisingly) don't animate nicely when nested views with the `presentationDetents(_:)` modifier applied appear and disappear:

<p align="center">
  
</p>

Should we forego this magic approach and instead have developers manage their detents by adding more cases to `FeatureFormView.EditingEvent` like `mapInteractionStarted`/`mapInteractionEnded`?

mhdostal:
> Should we forego this magic approach and instead have developers manage their detents by adding more cases to `FeatureFormView.EditingEvent` like `mapInteractionStarted`/`mapInteractionEnded`?

Will this alleviate the problems with sheets?

dfeinzimer:
> Will this alleviate the problems with sheets?

It doesn't seem to. There's some discussion about this issue [here](https://stackoverflow.com/questions/79447807/animating-content-height-when-changing-swiftui-sheet-detent-programmatically), but I haven't yet been able to make the suggestions work correctly with our views.

Regardless, it might be best overall to take the sheet/panel agnostic approach and just report what the user is doing via the new cases [suggested above](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/1163#:~:text=appear%20and%20disappear%3A-,Should%20we%20forego%20this%20magic%20approach%20and%20instead%20have%20developers%20manage%20their%20detents%20by%20adding%20more%20cases%20to%20FeatureFormView.EditingEvent%20like%20mapInteractionStarted/mapInteractionEnded%3F,-Member) on `FeatureFormView.EditingEvent` or another new enum altogether.

mhdostal:
> Regardless, it might be best overall to take the sheet/panel agnostic approach and just report what the user is doing via the new cases [suggested above](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/1163#:~:text=appear%20and%20disappear%3A-,Should%20we%20forego%20this%20magic%20approach%20and%20instead%20have%20developers%20manage%20their%20detents%20by%20adding%20more%20cases%20to%20FeatureFormView.EditingEvent%20like%20mapInteractionStarted/mapInteractionEnded%3F,-Member) on `FeatureFormView.EditingEvent` or another new enum altogether.

Ok. Let's try and make it the default/predominant use case work with little or not extra code.

PR #1165, opened 2025-05-20
Add new usage specific initializers for AttachmentsFeatureElementView

dfeinzimer:
This resolves a crash introduced in #1160 by ensuring that an `InternalFeatureFormViewModel` is present when needed.

Unlike with an `ObservableObject`, if an `@Observable` object is not in the environment, such as is the case when using `AttachmentsFeatureElementView` in a popup, a fatal error is encountered:

> SwiftUICore/Environment+Objects.swift:34: Fatal error: No Observable object of type InternalFeatureFormViewModel found. A View.environmentObject(_:) for InternalFeatureFormViewModel may be missing as an ancestor of this view.

Note that `FormViewModel` was renamed to `InternalFeatureFormViewModel` after #1160 in #1162.

Thank you to @CalebRas for catching this new crash.


mhdostal (approved):

CalebRas (approved):
Looks good, thanks for fixing this!

PR #1168, opened 2025-05-22
`FeatureFormView` - General refactoring and warning resolution

dfeinzimer:
- `InputWrapper` is generalized to `FormElementWrapper` to wrap `FormElement` instead of  `FieldFormElement`
-  `InputHeader` is generalized to `FormElementHeader` to wrap `FormElement` instead of  `FieldFormElement`
- `InputFooter` is generalized to `FormElementFooter` to wrap `FormElement` instead of  `FieldFormElement`
---
- `makeElement(_:)` is simplified to call `internalMakeElement(_:)` in more cases
- `makeUtilityAssociationsFormElement(_:)` is simplified to use `FormElementWrapper`
---
- `FieldFormElementView` is added
- `GroupView` is renamed to `GroupFormElementView`
---
| Before | After |
|---|---|
|  |  |


Examples/Examples/FeatureFormExampleView.swift
```diff
@@ -172,7 +172,7 @@ extension FeatureFormExampleView {
             submitButtonVisibility = .hidden
             return
         }
-        let databases = editedTables.compactMap { $0.serviceGeodatabase }
+        let databases = editedTables.compactMap(\.serviceGeodatabase)
```
dfeinzimer, line 175: Resolves SwiftLint warning

> Prefer Key Path Violation: Use a key path argument instead of a closure with property access (prefer_key_path)

mhdostal (approved):

PR #1171, opened 2025-05-29
`PopupView` - Support utility network associations

CalebRas:
## Description

This PR adds `UtilityAssociationsPopupElement` support to the `PopupView` following the `UtilityAssociationsPopupElementView.md` design. Tests will be added in a separate PR (`swift/issues/6934`).

This is my first large Toolkit PR, so I would appreciate an in-depth review. Thanks in advance.

#### Public API changes:

🟢  Additions:
- `PopupView.closeButton(_:)`
- `PopupView.onPopupChanged(perform:)`

🟡 Deprecations:
- `PopupView.header(_:)`

## Linked Issue(s)

- `swift/issues/6686`
- Feature Form PR: ~#1088~ #1176

## How To Test

- Modify `PopupExampleView.map` to use the test data found in the design. 
  - The URL can be found in the `common-toolkit/blob/main/designs/Popup/UtilityAssociationsPopupElementView.md#developer-pattern` section.
  - This will require authentication. Let me know if you have issues. 
- Tap on features and navigate through their associations.

## Screenshots

Simulator Screenshot - 1 - iPhone 16 Pro - 2025-05-29 at 13 32 16    Simulator Screenshot - 1 - iPhone 16 Pro - 2025-05-29 at 13 32 32
Simulator Screenshot - 1 - iPhone 16 Pro - 2025-05-29 at 13 31 39    Simulator Screenshot - 1 - iPhone 16 Pro - 2025-05-29 at 13 32 01

## To Discuss

Below is a list of the places this implementation deviates from the design. Feel free comment on any of these points.

- In the design, there is an extra back button, which clears the navigation stack all the way back to the root popup. This was not implemented since that isn't a common pattern on iOS. 
  - See `UtilityAssociationsPopupElementView.md#navigation`.
- In the design, a move-up arrow is shown in place of a chevron when a `UtilityAssociationGroupResult.associatedFeature` is the same as the root feature. Tapping on this takes you to the root popup and clears the navigation stack. This was not implemented for the same reason as above.
  - See `UtilityAssociationsPopupElementView.md#utilityassociationresult`.
- I didn't use Calcite icons where I thought an SFSymbol would do, such as with `list.bullet`.
- The `resultCount` is shown next to each `UtilityAssociationsFilterResult`. This was done to match the feature form implementation and is not in the design/ArcGIS Online.
  - See `UtilityAssociationsPopupElementView.md#utilityassociationspopupelement`.

I wrote `UtilityAssociationResultLink` and `SearchBar` with the intent that they could be shared with the feature form implementation. Let me know if you think they won't be used, and I will move the files out of `Sources/ArcGISToolkit/Common/`.

## Credits

Thanks @dfeinzimer for all your help and letting me look off your code.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Popups/UtilityAssociationsPopupElementView.swift
+++ b/Sources/ArcGISToolkit/Components/Popups/UtilityAssociationsPopupElementView.swift
@@ -62,9 +62,12 @@ struct UtilityAssociationsPopupElementView: View {
                 ProgressView()
                     .frame(maxWidth: .infinity)
                     .task {
-                        filterResultsResult = await Result {
+                        let filterResultsResult = await Result {
                             try await popupElement.associationsFilterResults
                                 .filter { $0.resultCount > 0 }
+                        }
+                        withAnimation {
+                            self.filterResultsResult = filterResultsResult
                         }
                     }
             }


Sources/ArcGISToolkit/Components/Popups/UtilityAssociationsPopupElementView.swift
```diff
@@ -0,0 +1,267 @@
...
+            comment: "A label in reference to utility associations elements contained within a popup."
+        ) : title
+    }
+}
+
+private extension UtilityAssociationResult {
+    /// A title for the result.
+    var displayTitle: String {
+        let popup = associatedFeature.toPopup()
+        return popup.title.isEmpty ? "Unknown" : popup.title
```
CalebRas, line 265: I just found out that `UtilityAssociationResult.title` was added in 200.8, which can be used here instead. Based on `nautilus/issues/2584`, though, it doesn't look like it is quite working yet, so my vote is to leave it as is for now and fix it later in the release. (Partly because then `v.next` would require `swift-daily` to build.) Up to you guys, though.

mhdostal: @CalebRas Yes, that sounds fine to me.

Sources/ArcGISToolkit/Components/Popups/PopupView.swift
```diff
@@ -35,179 +35,127 @@ import ArcGIS
...
-    private var isPresented: Binding<Bool>?
+    ///
+    /// This property can be removed once the deprecated
+    /// `init(popup:isPresented:)` initializer is removed.
+    var isPresented: Binding<Bool>?
+    
+    /// Creates a `PopupView` to display a given popup.
+    /// - Parameters:
+    ///   - popup: A binding to the popup that view displays. When `popup` is non-`nil`, the `PopupView` is displayed.
+    /// - Since: 200.8
```
dfeinzimer, line 80:
```suggestion
    /// Creates a `PopupView` to display a given popup.
    /// - Parameter popup: A binding to the popup that view displays. When `popup` is non-`nil`, the `PopupView` is displayed.
    /// - Since: 200.8
```

Sources/ArcGISToolkit/Common/NavigationLayer/NavigationLayer.swift
```diff
@@ -77,7 +77,7 @@ struct NavigationLayer<Content: View>: View {
                     root(model)
                         .transition(model.transition)
                 } else if let presented = model.presented?.view {
-                    AnyView(presented())
+                    AnyView(presented)
```
dfeinzimer, line 80: Thank you for making this change.

It might be good to move this and the associated changes in `NavigationLayerModel` into a dedicated PR.

Sources/ArcGISToolkit/Common/SearchBar.swift
```diff
@@ -0,0 +1,94 @@
...
+                    .labelStyle(.iconOnly)
+                    .buttonStyle(.plain)
+                    .foregroundColor(.secondary)
+                    .padding(5)
+                }
+            }
+            .animation(.default, value: text.isEmpty)
+            
+            if cancelButtonIsPresented {
+                Button("Cancel") {
```
dfeinzimer, line 77:
```suggestion
                Button.cancel {
```

Sources/ArcGISToolkit/Common/SearchBar.swift
```diff
@@ -0,0 +1,94 @@
...
+    /// that the button will animate when appearing.
+    @State private var cancelButtonIsPresented = false
+    
+    var body: some View {
+        HStack {
+            HStack(spacing: 5) {
+                Image(systemName: "magnifyingglass")
+                    .foregroundStyle(.secondary)
+                
+                let promptView = Text(prompt)
```
dfeinzimer, line 43: I don't think this will localize the prompt.

> Creates a text view that displays a stored string without localization. [Source](https://developer.apple.com/documentation/swiftui/text/init(_:)-9d1g4#:~:text=Creates%20a%20text%20view%20that%20displays%20a%20stored%20string%20without%20localization.)

Instead, change the type of `prompt` to `LocalizedStringKey` or wrap `prompt` with `LocalizedStringKey`.

More info (sourced [here](https://developer.apple.com/documentation/swiftui/text/init(_:tablename:bundle:comment:)#discussion)):

> When you initialize a text view with a string literal, the view triggers this initializer because it assumes you want the string localized, even when you don’t explicitly specify a table, as in the above example. If you haven’t provided localization for a particular string, you still get reasonable behavior, because the initializer displays the key, which typically contains the unlocalized string.

> If you initialize a text view with a string variable rather than a string literal, the view triggers the [init(_:)](https://developer.apple.com/documentation/swiftui/text/init(_:)-9d1g4) initializer instead, because it assumes that you don’t want localization in that case. If you do want to localize the value stored in a string variable, you can choose to call the init(_:tableName:bundle:comment:) initializer by first creating a [LocalizedStringKey](https://developer.apple.com/documentation/swiftui/localizedstringkey) instance from the string variable:

> ```Text(LocalizedStringKey(someString)) // Localizes the contents of `someString`.```

dfeinzimer (commented):
A few things to begin with.

CalebRas:
In response to this [feedback](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/1171#discussion_r2127168711) (which is now hidden), I moved the `NavigationLayer` changes into a separate [PR](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/1174) and based this PR of that branch to hide the changes from the diff. I will change it back once it has been merged.

Sources/ArcGISToolkit/Common/UtilityAssociationResultLink.swift
```diff
@@ -0,0 +1,113 @@
...
+    }
+    
+    /// A Boolean value indicating whether the `associatedFeature` global ID
+    /// matches the `toElement` global ID.
+    private var associatedFeatureIsToElement: Bool {
+        associatedFeature.globalID == association.toElement.globalID
+    }
+    
+    /// The description of the result's association.
+    var description: String? {
```
mhdostal, line 83: I'd like this to change to `details`. `description` is used in a bunch of other places, primarily as descriptive text authored in the web map. I spent time looking for the "description" property in the `UtilityAssociationResult` YAML before realizing it was in the extension.

cc: @dfeinzimer Thoughts?

mhdostal (changes requested):
Just the one comment

Sources/ArcGISToolkit/Common/UtilityAssociationResultLink.swift
dfeinzimer: Yes, `details` wouldn't be a bad idea.

Sources/ArcGISToolkit/Components/Popups/UtilityAssociationsPopupElementView.swift
```diff
@@ -0,0 +1,267 @@
...
+                Button {
+                    navigationLayerModel.push {
+                        SearchUtilityAssociationResultsView(results: groupResult.associationResults)
+                            .navigationLayerTitle(groupResult.name, subtitle: popupTitle)
+                    }
+                } label: {
+                    HStack {
+                        VStack(alignment: .leading) {
+                            Text("Show all")
+                            Text("Total: \(groupResult.associationResults.count)")
```
dfeinzimer, line 161: Localization comments needed here too.

Sources/ArcGISToolkit/Components/Popups/UtilityAssociationsPopupElementView.swift
dfeinzimer: - Looks like the PR was merged today, so support is coming soon.

> Partly because then v.next would require swift-daily to build.
- I believe for the daily builds, v.next does build against a daily of the SDK, so this shouldn't be an issue.

dfeinzimer: > Partly because then v.next would require swift-daily to build.

- I believe toolkit dailies build v.next with SDK dailies, so this shouldn't be an issue when the API is ready, which looks like it just went into verification today.

rolson:
The title should expand into all the open space instead of truncating so much: 
image

dfeinzimer:
> The title should expand into all the open space instead of truncating so much:

That is a `NavigationLayer` issue which I plan to address as part Apollo # 1234. We'll probably need a [custom layout](https://developer.apple.com/documentation/swiftui/layout) to handle all of the various different states of header content.

rolson:
> I'm getting this warning at the point of use: image

Looks like I need to pass a binding. I think the deprecation message should mention the type of the first parameter (popup), that might be helpful.

rolson:
I'm tapping the close button and it's not dismissing. I'm using the popupview in a sheet.

dfeinzimer (commented):
The latest code changes look good. 

On Catalyst I'm noticing some content is right against the border of the inner lighter background. It appears some padding is needed here.

<p align="center">
  
</p>

CalebRas:
@rolson 

> Looks like I need to pass a binding. I think the deprecation message should mention the type of the first parameter (popup), that might be helpful.

I included the binding type in the message. Let me know if that is not what you had in mind.

> I'm tapping the close button and it's not dismissing. I'm using the popupview in a sheet.

Similar to the feature form implementation, the close button only sets the `Binding<Popup?>` to `nil`. It's on the user to observe that change and close the presentation. (See the updated tutorial for an example.) Is that what you are doing, and it is still not working?

rolson:
No I'm not doing that. I figured it would use a dismiss action from the environment to close the sheet.

dfeinzimer:
> Are you using the latest `swift-daily`?

Ah, I was still on Friday's daily. Today's looks much better.

dfeinzimer (commented):
For your search I think you'll want to use `localizedStandardContains` as we do [here](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/1268d3e3df6517ef0cb1ee2bb10ccf8f0fe0ee99/Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift#L239).

Currently things appear to be case sensitive:

dfeinzimer:
> Can we remember the collapsed/expanded state of the groups?

Unfortunately, another current NavigationLayer limitation.

rolson:
> > Can we remember the collapsed/expanded state of the groups?
> 
> Unfortunately, another current NavigationLayer limitation.

Ok - if we don't have an issue please create one

dfeinzimer:
> Ok - if we don't have an issue please create one

Apollo 1126

CalebRas:
> No I'm not doing that. I figured it would use a dismiss action from the environment to close the sheet.

@dfeinzimer and @mhdostal, Is there a reason this wasn't a part of the original implementation? If not, are you okay with me making this change? It would probably require special handling if we wanted to also have it work with the floating panel, but maybe that can be added in another PR.

dfeinzimer:
> Also, using this sheet initializer will get this to work without a `dismiss` call:
> 
> ```swift
> .sheet(item: $popup) { popup in // (Note that Popup identifiable conformance would be a requirement for this)
>     PopupView(popup: $popup)
> }
> ```

Uh oh, looks like there are some issues with this when navigating to a new popup. We'll probably need to look into this:

Simulator Screen Recording - iPhone 16 - 2025-06-10 at 11 06 06

Sources/ArcGISToolkit/Components/Popups/InternalPopupView.swift
```diff
@@ -0,0 +1,140 @@
...
+                    default:
+                        EmptyView()
+                    }
+                }
+#if targetEnvironment(macCatalyst)
+                .listRowInsets(.toolkitDefault)
+#endif
+            }
+            .listStyle(.plain)
+            .phoneBottomPadding()
```
rolson, line 115: Do you recall why this is needed here?

Sources/ArcGISToolkit/Components/Popups/InternalPopupView.swift
```diff
@@ -0,0 +1,140 @@
...
+        }
+    }
+    
+    /// An object used to hold the result of evaluating the expressions of a popup.
+    private final class Evaluation {
```
rolson, line 120: would a struct work here? If so you wouldn't need the init

Sources/ArcGISToolkit/Components/Popups/InternalPopupView.swift
```diff
@@ -0,0 +1,140 @@
...
+        init(elements: [PopupElement] = [], error: Error? = nil) {
+            self.elements = elements
+            self.error = error
+        }
+    }
+}
+
+extension EnvironmentValues {
+    /// The title of the popup associated with the view.
+    @Entry var popupTitle = ""
```
rolson, line 139: Does this work back to Xcode 16?

Sources/ArcGISToolkit/Components/Popups/PopupView.swift
```diff
@@ -35,179 +35,126 @@ import ArcGIS
...
-                            .fontWeight(.bold)
-                    }
-                    Spacer()
-                    if (showCloseButtonDeprecatedModifierIsApplied && showCloseButton)
-                        || (!showCloseButtonDeprecatedModifierIsApplied && isPresented != nil) {
-                        XButton(.dismiss) {
-                            isPresented?.wrappedValue = false
+        VStack(spacing: 0) {
+            if let popup = presentedPopups.first {
+                NavigationLayer { model in
```
rolson, line 87:
```suggestion
                EmbeddedNavigationStack { model in
```
The NavigationLayer name doesn't need to change in this PR, but I do think we should come up with a different name. This is just a possible suggestion. "Embedded" in the sense that it works within an existing system level navigation stack.

Sources/ArcGISToolkit/Components/Popups/UtilityAssociationsPopupElementView.swift
```diff
@@ -0,0 +1,278 @@
...
+    /// A Boolean value indicating whether the associations filter results are being loaded.
+    @State private var isLoading = true
+    
+    var body: some View {
+        DisclosureGroup(isExpanded: $isExpanded) {
+            if isLoading {
+                ProgressView()
+                    .frame(maxWidth: .infinity)
+                    .task {
+                        let filterResults = try? await popupElement.associationsFilterResults
```
rolson, line 39: Do we want to ignore the error here?

Sources/ArcGISToolkit/Common/UtilityAssociationResultButton.swift
```diff
@@ -0,0 +1,106 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ArcGIS
+import SwiftUI
+
+/// A view that displays a `UtilityAssociationResult` and performs an action
+/// when it is tapped.
+struct UtilityAssociationResultButton: View {
```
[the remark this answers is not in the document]
dfeinzimer: I intend to update the original forms implementation when this PR is merged to use this new shared version.

Sources/ArcGISToolkit/Components/Popups/PopupView.swift
dfeinzimer: Added this suggestion to Apollo 1235.

Examples/Examples/PopupExampleView.swift
```diff
@@ -68,10 +66,16 @@ struct PopupExampleView: View {
...
                     horizontalAlignment: .leading,
                     isPresented: $showPopup
-                ) { [popup] in
-                    PopupView(popup: popup!, isPresented: $showPopup)
-                        .padding()
+                ) {
+                    PopupView(popup: $popup)
+                }
+                .onChange(of: popup == nil) {
+                    showPopup = popup != nil
```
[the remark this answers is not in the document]
dfeinzimer: Created #1178

CalebRas:
> Uh oh, looks like there are some issues with this when navigating to a new popup. We'll probably need to look into this:

@dfeinzimer That is likely because the popup binding gets updated to the new popup when you navigate to another `InternalPopupView`, which then causes the sheet to re-present. If we wanted to be able to use the popup binding as an "item" to present a sheet or floating panel, we will probably have to remove that functionality, which means the user will not have access to subsequent as they navigate through the associations. Is that the route you would like me to take?

dfeinzimer:
> Is that the route you would like me to take?

No, I think the binding should remain. I was expecting that a sheet would remain presented if the bound item changed, without going `nil` between values, but from my testing, I'm seeing that is not the case.

I suppose this mean's we'd want to consider `sheet(item:onDismiss:content:)` not supported when using popups with UNA.

Sources/ArcGISToolkit/Components/Popups/InternalPopupView.swift
rolson: default args work with the implicit initializer if you make the properties `var` and set the values accordingly.

> Would you still like me to change it to a struct?

If it works the same as a struct, then yes as structs should be favored over classes by default where possible.

Examples/Examples/PopupExampleView.swift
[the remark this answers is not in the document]
rolson: No, I don't prefer the custom binding. It's just unfortunate that we need both a bool and an item in this case. It would be really good if we could make this simpler in some way. It doesn't need to be part of this PR. Should the PopupView then take a binding to a Bool isPresented so that it can set that to false when the close button is tapped? 

The fact that it doesn't work with the `sheet(item:...)` is confusing, especially in light of this comment:

```swift
- Parameter popup: A binding to the popup that view displays. When `popup` is non-`nil`, the `PopupView` is displayed.
```

We need to step back here and think of a clear strategy. We want to support FloatingPanel - fine. But I think we need to support Sheet first and foremost because that is a system SwiftUI View.

Sources/ArcGISToolkit/Components/Popups/InternalPopupView.swift
CalebRas: Since I was going to be modifying this code anyway, I replaced `Evaluation` with a `Result`, as that seemed more appropriate.

CalebRas:
As we discussed, I went back to using the old `PopupView` initializer and updated the PR description accordingly. I also reverted most of the example and tutorial changes since they were no longer applicable. This is ready for review again.

dfeinzimer (dismissed):
Looks good & working well across Catalyst, iOS and Vision.

rolson:
Looks like you will need to filter cancellation error and/or retry:

image

rolson:
I think it would be good if the dismiss button stayed in the same spot and didn't move up and down as the size of the header changes.

Simulator Screen Recording - iPhone 16 Pro - 2025-06-13 at 14 38 25

rolson:
I don't think this requires an exclamation mark unless every UN feature is supposed to have associations?

image

rolson:
> > Looks like you will need to filter cancellation error and/or retry:
> 
> Can you share how you reproduced this please?

I think what I did was when the popup first showed, I scrolled down past fields to see associations then scrolled back up to the top, which I think cancelled something. Then collapsed the fields section and it showed that error.

CalebRas:
> I think it would be good if the dismiss button stayed in the same spot and didn't move up and down as the size of the header changes.

I am fairly certain this would have to be fixed in the `NavigationLayer`, so I added it to `apollo/issues/1234`.

CalebRas:
> I don't think this requires an exclamation mark unless every UN feature is supposed to have associations?

I don't think every feature is supposed to have associations, but my vote would be to still keep it. Similar to `ContentUnavailableView`, it seems like a symbol should go with the text, and the exclamation mark is the one the design and Online uses. (I didn't use `ContentUnavailableView` becuase it takes up a lot of room.) It's not a big deal, though, so if you still aren't a fan, let me know, and I'll remove it. 

Screenshot 2025-06-13 at 2 08 50 PM

dfeinzimer (approved):

rolson:
> > I don't think this requires an exclamation mark unless every UN feature is supposed to have associations?
> 
> I don't think every feature is supposed to have associations, but my vote would be to still keep it. Similar to `ContentUnavailableView`, it seems like a symbol should go with the text, and the exclamation mark is the one the design and Online uses. (I didn't use `ContentUnavailableView` becuase it takes up a lot of room.) It's not a big deal, though, so if you still aren't a fan, let me know, and I'll remove it.
> 
> Screenshot 2025-06-13 at 2 08 50 PM

Unless that signifies there was an error fetching associations, I think that's a bad design. A red exclamation means error, an exclamation of any color to me signifies something un-expected. Reach out to @PeggyDu09 and get her thoughts, if she thinks it's unnecessary to show that icon then we can remove it in a subsequent PR.

rolson (approved):
Really nice work @CalebRas !

mhdostal (approved):

PR #1185, opened 2025-06-14
Implement secrets setup

dfeinzimer:
Apollo 1263


README.md
```diff
@@ -52,6 +52,30 @@ Use of ArcGIS location services, such as basemap styles, geocoding, and routing
...
 
+> [!IMPORTANT]
+> To run the example apps and access ArcGIS Location Services, follow these steps to obtain an **API key** access token and store it in a secrets file stored in the project file's directory, `$(SRCROOT)/.secrets`.
+
+1. Go to the [Create an API key](https://developers.arcgis.com/documentation/security-and-authentication/api-key-authentication/tutorials/create-an-api-key/) tutorial to obtain the API key access token. Ensure that the following privileges are enabled:
+
+* Location services > Basemaps
+* Location services > Routing
+
+2. Create a hidden secrets file in the project file's directory.
```
yo1995, line 63:
```suggestion
2. Create a hidden secrets file in the toolkit's root directory.
```

Per current config it would be the root directory of this repo.

Shared/AppSecrets.swift.masque
```diff
@@ -15,6 +15,6 @@
 import ArcGIS
 
 extension APIKey {
-    /// The default API key.
-    static let `default` = APIKey("<#API Key#>")
+    /// The API key for toolkit examples and testing.
+    static let iOS = APIKey("{{ ARCGIS_API_KEY_IOS }}")
```
philium, line 19: Is this key actually iOS-specific? If not, consider renaming it to something like `toolkit`.

Documentation/ConfigureAppSecrets.md
```diff
@@ -0,0 +1,25 @@
+# Configure App Secrets
+
+As a best-practice principle, our project conceals app secrets from source code by generating and compiling an `AppSecrets.swift` source code file at build time using a custom build rule. The same pattern can be applied to your app development as well.
+
+This build rule looks for a secrets file stored in the toolkit's root directory, `/.secrets`.
+
+Note: License strings are not required for development. Without licensing or licensing with invalid keys do not throw an exception, but simply fail to license the app, falling back to Developer Mode (which will display a watermark on the map and scene views). Apply the license strings when your app is ready for deployment.
```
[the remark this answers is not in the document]
dfeinzimer: I had copied this word for word from samples. I've revised it now.

Shared/AppSecrets.swift.masque
dfeinzimer: I had gone with `iOS` as it goes with the `_IOS` suffix in `ARCGIS_API_KEY_IOS` which is the key that the Samples project is also using. If we match the samples project people can just copy their existing secrets file into the Toolkit repository. Would you be open to reconsidering keeping `iOS`?

philium: I don't know. It is easy enough to change after copying. Instead of perpetuating a poor naming choice, it might be better to submit a PR to the samples app to change the name there and then use that same name here.

A little background might be helpful. We originally thought that the Samples app might  have iOS or macOS specific samples that might need to use separate keys, so we had both a iOS and macOS key. At some point the macOS key was done away with, but the iOS key wasn't renamed.

Also, the file was put in the Shared directory in the Samples repo to mimic the directory structure used by the multiplatform app template (at least at the time). That doesn't really make sense in this repo because the root of the repo doesn't contain a project. You could put the file at the root of the repo.

dfeinzimer: Okay, thank you for the background. I was curious about that naming. I'm happy to change course here and go with a better name and carry that work into the Samples repo.

I could see the same naming issue with `toolkit` as `iOS`, it's not necessarily toolkit-specific. Would `developer` work well? The point of use would be:

```swift
ArcGISEnvironment.apiKey = .developer
```

philium: I suppose either of those will do if you want to try and use the same name across various different projects. Or, to acknowledge the covert nature of the key, you could use `secret`.

dfeinzimer: I landed on `.development` for now. Folks/orgs with multiple keys might have one for development, one for staging and one for production.

philium: Things in the root of the repo are shared by convention. For example, the gitinore file and the SwiftLint config can apply to everything in the repo. You can put it where you want, but I think the root of the repo would be just fine.

Shared/AppSecrets.swift.masque
```diff
@@ -15,6 +15,6 @@
 import ArcGIS
 
 extension APIKey {
-    /// The default API key.
-    static let `default` = APIKey("<#API Key#>")
+    /// The API key for toolkit examples and testing.
```
philium, line 18: If you want this to be project agnostic, you'll want to change this comment. Perhaps something like "The API key for development of packages/projects in this repository."

philium (dismissed):

Documentation/ConfigureAppSecrets.md
```diff
@@ -15,7 +15,7 @@ Note: License strings are not required for development. Without licensing or lic
 2. Add your **API Key** access token to the secrets file.
 
   ```sh
-  echo ARCGIS_API_KEY_IOS=your-api-key >> .secrets
+  echo ARCGIS_API_KEY_DEVELOPMENT=your-api-key >> .secrets
   ```
 
   > Replace 'your-license-key', 'your-extension-license-key' and 'your-api-key' with your keys.
```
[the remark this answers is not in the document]
dfeinzimer: Nice catch.

Test Runner/Test Runner.xcodeproj/project.pbxproj
```diff
@@ -327,11 +351,34 @@
...
+			);
+			inputFileListPaths = (
+			);
+			inputPaths = (
+			);
+			name = "Create .secrets File If It Does Not Exist";
+			outputFileListPaths = (
+			);
+			outputPaths = (
+			);
```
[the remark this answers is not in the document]
yo1995: Scrub the screenshot… 🤦 that was misleading. `$(SRCROOT)/../.secrets` as you figured out!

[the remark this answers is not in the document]
dfeinzimer: Thank you for catching this! Fixed now.

yo1995 (approved):

PR #1191, opened 2025-06-23
`PopupView` - Utility network associations enhancements

CalebRas:
## Description

This PR makes a couple of updates to the `PopupView`  utility network associations support (added in #1171):
- Replaced `NavigationLayer` with `NavigationStack`.
  - This required using a sheet instead of a floating panel in the example.
  - This also allowed for a few more native APIs to be used, such as `searchable` instead of the custom `SearchBar`.
- Enhanced `UtilityAssociationsPopupElementView` to fetch the associations filter results in the background.
- Removes the exclamation mark icon from `UtilityAssociationsPopupElementView` "No Associations" label (see [comment](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/1171#issuecomment-2984911320)).
- Fixed a few small bugs:
   - An issue where the `UtilityAssociationsPopupElementView` progress view wouldn't reappear after scrolling away from it in the list.
   - A `ForEach` undefined results runtime error that would be thrown when the `UtilityAssociationsFilterResultLink`'s filter result had multiple group results with the same name.

## Linked Issue(s)

- `swift/issues/6686`

### Example

|Before|After|
|:-:|:-:|
|    Simulator Screenshot - 1 - iPhone 16 Pro - 2025-06-23 at 11 30 33    |     Simulator Screenshot - 1 - iPhone 16 Pro - 2025-06-23 at 11 28 31    |

### Popup UNA

|Before|After|
|:-:|:-:|
|    Simulator Screenshot - 1 - iPhone 16 Pro - 2025-05-29 at 13 32 16    |     Simulator Screenshot - 1 - iPhone 16 Pro - 2025-06-23 at 11 25 22    |
|    Simulator Screenshot - 1 - iPhone 16 Pro - 2025-05-29 at 13 31 39    |     Simulator Screenshot - 1 - iPhone 16 Pro - 2025-06-23 at 11 22 18     |

## How To Test

- Modify `PopupExampleView.map` to use the test data found in the design. 
  - The URL can be found in the `common-toolkit/blob/main/designs/Popup/UtilityAssociationsPopupElementView.md#developer-pattern` section.
  - This will require authentication. Let me know if you have issues. 
- Tap on features and navigate through their associations.


Sources/ArcGISToolkit/Components/Popups/PopupViewToolbar.swift
```diff
@@ -0,0 +1,69 @@
...
+                        Text(title)
+                            .fontWeight(.semibold)
+                            .font(subtitle != nil ? .subheadline : .headline)
+                        
+                        if let subtitle {
+                            Text(subtitle)
+                                .font(.caption2)
+                                .foregroundStyle(.secondary)
+                        }
+                    }
```
CalebRas, line 55: Since [navigationSubtitle(_:)](https://developer.apple.com/documentation/swiftui/view/navigationsubtitle(_:)-2h20t) is getting added to iOS and iPadOS in `26.0` I thought it would be okay to match the design using a custom implementation. I'm not sure this is the best path forward, though, so I would like to hear other peoples opinions.

Screenshots for reference:

|My Implementation on `18.5`|navigationSubtitle on `26.0`|
|:-:|:-:|
|     Simulator Screenshot - iPhone 16 Pro - 2025-06-23 at 14 44 45    |     Simulator Screenshot - iPhone 16 Pro - 2025-06-23 at 14 49 49    |

Examples/Examples/PopupExampleView.swift
```diff
@@ -63,12 +54,7 @@ struct PopupExampleView: View {
...
                 }
-                .floatingPanel(
-                    attributionBarHeight: attributionBarHeight,
-                    selectedDetent: $floatingPanelDetent,
-                    horizontalAlignment: .leading,
-                    isPresented: $showPopup
-                ) { [popup] in
+                .sheet(isPresented: $showPopup) { [popup] in
                     PopupView(popup: popup!, isPresented: $showPopup)
                 }
```
dfeinzimer, line 59: As part of this PR, should we add some documentation to note the changes in expectations where a `PopupView` can be used?

I'm not sure if an entire DocC article would be warranted, but perhaps an added blurb to the existing doc at the top of PopupView.swift?

Sources/ArcGISToolkit/Components/Popups/EmbeddedPopupView.swift
```diff
@@ -53,7 +53,8 @@ struct EmbeddedPopupView: View {
                 .frame(maxWidth: .infinity)
             }
         }
-        .navigationLayerTitle(popup.title)
+        .preference(key: PresentedPopupPreferenceKey.self, value: .init(popup: popup))
```
dfeinzimer, line 56: Just curious - any reason to go with a preference? `onAppear` seemed to suffice for the FeatureFormView implementation. Did you find it cleaner than doing something like passing down a `popupChangedAction`?

Sources/ArcGISToolkit/Components/Popups/PopupViewToolbar.swift
dfeinzimer: [I updated](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/1190/commits/63d119838a2275a8293fe45c7390d8618008f0d5) the FeatureFormView implementation this morning to use `View.navigationSubtitle(_:)` where available. I'd also be okay with doing this as a fallback if that would be everyone's preference.

Sources/ArcGISToolkit/Components/Popups/EmbeddedPopupView.swift
CalebRas: Since in this implementation there are two tail-end views of a navigable popup (`UtilityAssociationGroupResultView` and `SearchUtilityAssociationResultsView`) instead of one, like with the feature form, it made it pretty complicated to get the correct behavior using `onAppear`. So, yea, it was much simpler to use a preference key since it gets set automatically when the an `EmbeddedPopupView` gets popped off the navigation stack.

Examples/Examples/PopupExampleView.swift
dfeinzimer: Added similar for the FeatureFormView conversion in https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/1190/commits/3066cc011c488b0dc2c7ade8a905a2da2d98d527.

Sources/ArcGISToolkit/Components/Popups/PopupView.swift
```diff
@@ -44,9 +44,17 @@ import ArcGIS
 ///
 /// **Behavior**
 ///
+/// As of 200.8, PopupView uses a NavigationStack internally to support browsing utility network
+/// associations. As a result, a PopupView now requires a navigation context isolated from any
```
dfeinzimer, line 48: Nitpick. I dropped "now" from the Feature Form version. Take or leave.

```suggestion
/// associations. As a result, a PopupView requires a navigation context isolated from any
```

Sources/ArcGISToolkit/Components/Popups/PopupViewToolbar.swift
mhdostal: The subtitle is what displays the "Fraction along edge" value and the "Terminal: xxx" value, so I think it's necessary for both. I do think the 18.5 version is a bit too compact, but the iOS 26 version is too expanded. But as those seem to be the default behavior, I'm OK with it.

mhdostal (approved):

Sources/ArcGISToolkit/Components/Popups/PopupViewToolbar.swift
CalebRas: > why is the vertical spacing so different between 18.5 and 26?

I think the toolbar is just bigger on 26 since the buttons are also bigger. I am not sure this can be helped, though, because any added padding is ignored on 18.5.

> But my slight preference might be just utilizing .navigationSubtitle when available and not having a subtitle at all for iOS 18.

I'm leaning towards this way as well and will probably remove the custom subtitle if @mhdostal is okay with that.

> The subtitle is what displays the "Fraction along edge" value and the "Terminal: xxx" value, so I think it's necessary for both. I do think the 18.5 version is a bit too compact, but the iOS 26 version is too expanded. But as those seem to be the default behavior, I'm OK with it.

Currently, in the popup, the subtitle is only used to display the popup's title when the user is navigating through its associations. @mhdostal, are you saying that feature form will need a custom subtitle to support additional requirements?

mhdostal (dismissed):

Sources/ArcGISToolkit/Components/Popups/PopupViewToolbar.swift
dfeinzimer: But I will say my preference would be to also just use `navigationSubtitle` where available.

CalebRas: > Unless I'm looking at different content, the FF design includes the additional possible subtitles for "Terminal" and "Fraction Along Edge".

Ah I see. I think we are only talking about the subtitle in the navigation bar. So in the case of your screenshot, the `Busbar - 10000` text below `Electric Distribution Juncti...` would only appear on Mac Catalyst and iOS/iPadOS 26+. Otherwise, just `Electric Distribution Juncti...` display.  The association result descriptions, like `Terminal: High` will still be there on all platforms. Is that okay?

rolson (dismissed):

Sources/ArcGISToolkit/Components/Popups/PopupViewToolbar.swift
rolson: I'm fine with the custom subtitle implementation. But we should have iOS 26 use the new implementation (when we can)

mhdostal (approved):

rolson (approved):

Sources/ArcGISToolkit/Extensions/SwiftUI/View+NavigationTitle.swift
```diff
@@ -0,0 +1,84 @@
...
+#if swift(>=6.2) && !os(visionOS)
+        if #available(iOS 26.0, *) {
+            self.navigationTitle(title)
+                .navigationSubtitle(subtitle)
+        } else {
+            self.modifier(NavigationTitleModifier(title: title, subtitle: subtitle))
+        }
+#elseif targetEnvironment(macCatalyst)
+        self.navigationTitle(title)
+            .navigationSubtitle(subtitle)
```
dfeinzimer, line 72: Looks like this might've been bugged still in the Xcode 16.x series:

<p align="center">
  
</p>

Working okay in 26:

<p align="center">
  
</p>

dfeinzimer (approved):

Sources/ArcGISToolkit/Extensions/SwiftUI/View+NavigationTitle.swift
dfeinzimer: I put an updated version on #1190 to let this PR proceed as approved. The fix will get picked up automatically when that's merged in.

Issue #1166, opened 2025-05-21
OfflineMapAreasView - Rename delete map area button


mmoosbac94:
We noticed that after a user downloads a map area using the OfflineMapAreasView Toolkit component, a button named "Delete Map Area" appears. However, we've received feedback that this naming can be misleading and potentially confusing for users. Because the action doesn't actually delete the map area itself—it removes the downloaded data.

To make this clearer, we suggest renaming the button to something like "Delete Download". The corresponding German translation would be "Download entfernen".

We would really appreciate your consideration of this change.

Thank you,
Marvin

yo1995:
Thanks for the suggestion. We'll discuss and let you know when we make the change.

PR #1192, opened 2025-06-23
Offline: Rename `delete map area` button

des12437:
Related to https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/1166.

Renames the "Delete Map Area" button to "Remove Download" and "Delete Download" for the preplanned and on-demand use-cases respectively.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreaMetadataView.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreaMetadataView.swift
@@ -54,20 +54,7 @@ struct OfflineMapAreaMetadataView<Metadata: OfflineMapAreaMetadata>: View {
                         }
                         model.removeDownloadedArea()
                     } label: {
-                        switch model.mapMode {
-                        case .preplanned, .undetermined:
-                            Text(
-                                "Remove Download",
-                                bundle: .toolkitModule,
-                                comment: "A label for a button to remove a map area download."
-                            )
-                        case .onDemand:
-                            Text(
-                                "Delete Download",
-                                bundle: .toolkitModule,
-                                comment: "A label for a button to delete a map area download."
-                            )
-                        }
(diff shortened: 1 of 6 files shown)


yo1995 (approved):

Sources/ArcGISToolkit/Components/Offline/OfflineMapAreaMetadataView.swift
```diff
@@ -147,7 +156,8 @@ protocol OfflineMapAreaMetadata: ObservableObject {
     var directorySize: Int { get }
     /// A Boolean value indicating whether the metadata view should be dismissed when the map area is deleted.
     var dismissMetadataViewOnDelete: Bool { get }
-    
+    /// The map mode that determines if the map is preplanned or on-demand.
+    var mapMode: OfflineMapViewModel.Mode { get }
```
rolson, line 160: If I look at the pattern established by this protocol, this mode property doesn't really match the existing pattern.

The usage of the protocol in question doesn't look at a mode and make a decision, the protocol defines how to show things. This mode changes that such that the usage looks at the mode and makes a decision. Whereas right now the types implementing the mode make a decision (ie what title to show, whether downloads are allowed).

In this way, I think an approach that seems to fit in better with the existing pattern would be to expose a string property that defines that deleteButtonText.

des12437: > In this way, I think an approach that seems to fit in better with the existing pattern would be to expose a string property that defines that deleteButtonText.

Instead of exposing a String, a LocalizedStringResource can be exposed and used directly.

rolson: Great

rolson (approved):

yo1995 (approved):

PR #1200, opened 2025-07-03
`FeatureFormView` - Suppress updates on discard edits

dfeinzimer:
Apollo 1294


mhdostal (dismissed):

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/ComboBoxInput.swift
```diff
@@ -91,7 +91,7 @@ struct ComboBoxInput: View {
                 XButton(.clear) {
                     model.focusedElement = element
                     defer { model.focusedElement = nil }
-                    updateValue(nil)
+                    model.updateValueAndEvaluateExpressions(element, nil)
```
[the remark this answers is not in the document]
dfeinzimer: I started out with something like you describe - validation directly in each of the input views, but it felt disjoint and things gradually evolved into this centralized method. I like that this centralizes all the calls to `evaluateExpressions()`, reducing the code in each of the various input views to one line. Colocating the `guard`s and `updateValue`s also makes it easy to quickly peak at the criteria for updating each type of form input. 

I also just looked at adding
`FieldFormElement.updateValue(to: CodedValue?)`,
`FieldFormElement.updateValue(to: Date?)`, and
`FieldFormElement.updateValue(to: String?)`
or
`FormInput.updateValue(element:value:CodedValue?)`,
`FormInput.updateValue(element:value:Date?)`, and
`FormInput.updateValue(element:value:String?)`
but I liked neither of these because we'd still need to return whether the update was suppressed or not.

mhdostal (dismissed):

mhdostal (dismissed):

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/ComboBoxInput.swift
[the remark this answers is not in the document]
philium: > reducing the code in each of the various input views to one line.

But that code is just moved, right? It isn't like that code has gone away? And without the switch statement, the code in each of the relevant methods would actually be simpler?

> makes it easy to quickly peak at the criteria for updating each type of form input

I get that, but the criteria is different for each one, right? Wouldn't it make more sense to keep all the code related to a particular input together, in one place? In other words, there are tradeoffs either way. In my opinion, the tradeoffs for putting all the code in a single method are not worth it. I would even go so far as to say that putting type-specific code in a general method only to test for a specific type and then run the type-specific code is a code smell. Just put the type-specific code on the specific type and eliminate the type casting.

dfeinzimer:
@philium this is ready for review again.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/ComboBoxInput.swift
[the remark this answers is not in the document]
philium: You're welcome.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/ComboBoxInput.swift
```diff
@@ -233,9 +233,11 @@ extension ComboBoxInput {
...
     }
     
-    private func updateValue(_ value: (any Sendable)?) {
-        element.updateValue(value)
-        internalFeatureFormViewModel.evaluateExpressions()
+    private func updateValueAndEvaluateExpressions(_ value: CodedValue?) {
+        if element.formattedValue != value?.name {
+            element.updateValue(value?.code)
+            internalFeatureFormViewModel.evaluateExpressions()
+        }
```
philium, line 240: This is a [Yoda condition](https://en.wikipedia.org/wiki/Yoda_conditions), which we try to avoid. Also, it would make sense to use a guard statement here.

```suggestion
        guard value?.name != element.formattedValue else { return }
        element.updateValue(value?.code)
        internalFeatureFormViewModel.evaluateExpressions()
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/DateTimeInput.swift
```diff
@@ -56,9 +56,11 @@ struct DateTimeInput: View {
                 isEditing = internalFeatureFormViewModel.focusedElement == element
             }
             .onChange(of: date) {
-                element.updateValue(date)
+                if element.value as? Date != date {
+                    element.updateValue(date)
+                    internalFeatureFormViewModel.evaluateExpressions()
+                }
                 formattedValue = element.formattedValue
-                internalFeatureFormViewModel.evaluateExpressions()
```
philium, line 61: Again, Yoda condition and if instead of guard:

```swift
guard date != element.value as? Date else { return }
element.updateValue(date)
internalFeatureFormViewModel.evaluateExpressions()
formattedValue = element.formattedValue
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/RadioButtonsInput.swift
```diff
@@ -96,8 +96,10 @@ struct RadioButtonsInput: View {
                 }
             }
             .onChange(of: selectedValue) {
-                element.updateValue(selectedValue?.code)
-                internalFeatureFormViewModel.evaluateExpressions()
+                if element.formattedValue != selectedValue?.name {
+                    element.updateValue(selectedValue?.code)
+                    internalFeatureFormViewModel.evaluateExpressions()
+                }
```
philium, line 102:
```suggestion
                guard selectedValue?.name != element.formattedValue else { return }
                element.updateValue(selectedValue?.code)
                internalFeatureFormViewModel.evaluateExpressions()
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/SwitchInput.swift
```diff
@@ -81,8 +81,11 @@ struct SwitchInput: View {
             // loads, depending if the initial value matches the default value
             // defined for `isOn`.
             .onChange(of: isOn) {
-                element.updateValue(isOn ? input.onValue.code : input.offValue.code)
-                internalFeatureFormViewModel.evaluateExpressions()
+                if (isOn && element.formattedValue == input.offValue.name)
+                    || (!isOn && element.formattedValue == input.onValue.name) {
+                    element.updateValue(isOn ? input.onValue.code : input.offValue.code)
+                    internalFeatureFormViewModel.evaluateExpressions()
+                }
```
philium, line 88:
```suggestion
                guard (isOn && element.formattedValue == input.offValue.name)
                        || (!isOn && element.formattedValue == input.onValue.name) else {
                    return
                }
                element.updateValue(isOn ? input.onValue.code : input.offValue.code)
                internalFeatureFormViewModel.evaluateExpressions()
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/TextInput.swift
```diff
@@ -237,8 +239,10 @@ private extension TextInput {
 #endif
             }
             RepresentedUITextView(initialText: text) { text in
-                element.convertAndUpdateValue(text)
-                internalFeatureFormViewModel.evaluateExpressions()
+                if element.formattedValue != text {
+                    element.convertAndUpdateValue(text)
+                    internalFeatureFormViewModel.evaluateExpressions()
+                }
```
philium, line 245:
```suggestion
                guard text != element.formattedValue else { return }
                element.convertAndUpdateValue(text)
                internalFeatureFormViewModel.evaluateExpressions()
```

philium (commented):
FYI: Using the formatted values from an element means that the user will not get the value they expect if they set the environment's locale to anything but the current locale. And storing the formatted value means that the text on screen may not get updated appropriately if the current locale were to change. If you haven't done so already, consider using SwiftUI and format styles to format the values.

dfeinzimer:
@philium this is ready for review again.

> FYI: Using the formatted values from an element means that the user will not get the value they expect if they set the environment's locale to anything but the current locale. And storing the formatted value means that the text on screen may not get updated appropriately if the current locale were to change. If you haven't done so already, consider using SwiftUI and format styles to format the values.

Given that use of formatted values goes beyond the scope of the bug targeted here, I will handle this feedback in #1208.

philium (approved):

mhdostal (approved):

PR #1211, opened 2025-07-10
`PopupView` - Fix popover background on Mac Catalyst

CalebRas:
### Description

This PR fixes an issue where the `PopupView` background wouldn't display correctly when shown in a `popover` on Mac Catalyst. The appearance of the view using other presentation modes and platforms looks to be unaffected by these changes.

### How To Test

- Replace the `PopupExampleView` sheet with a popover:

```swift
.popover(isPresented: $showPopup, attachmentAnchor: .point(.top)) { [popup] in
    PopupView(root: popup!, isPresented: $showPopup)
        .frame(idealWidth: 320, idealHeight: 500)
}
```

- Run on Mac Catalyst and open a popup.

### Screenshots

|Before|After|
|:-:|:-:|
|    Screenshot 2025-07-09 at 4 43 12 PM    |    Screenshot 2025-07-09 at 4 44 51 PM    |

### DIFF
--- a/Sources/ArcGISToolkit/Components/Popups/EmbeddedPopupView.swift
+++ b/Sources/ArcGISToolkit/Components/Popups/EmbeddedPopupView.swift
@@ -131,7 +131,7 @@ extension EmbeddedPopupView {
                 .listRowInsets(.toolkitDefault)
 #endif
             }
-            .listStyle(.plain)
+            .listStyle(.inset)
         }
     }
 }


CalebRas:
> Dark mode still creates a secondary background. Not sure if there's a way around that but it might be worth looking into.

I wonder if this is intended behavior since using generic views also produces this effect:

```swift
NavigationStack {
    List {
        Text("Placeholder")
        Text("Placeholder")
        Text("Placeholder")
    }
    .listStyle(.inset)
    .navigationTitle("Placeholder")
}
``` 
Screenshot 2025-07-10 at 10 07 57 AM

More interestingly, it only does this when there is a parent `NavigationSplitView`. If the one in the Examples is replaced with a `NavigationStack`, then the list background is black...

@dfeinzimer Would you still like me to work around this?

dfeinzimer:
> Would you still like me to work around this?

I'm okay with proceeding with these changes as-is for now

rolson:
According to the screen shots, this inset the popup contents more than they were before on macOS. I'm not sure the changes look good. Do they affect the popupview when not shown in a popover? Did the same happen for iOS?

CalebRas:
> According to the screen shots, this inset the popup contents more than they were before on macOS. I'm not sure the changes look good. Do they affect the popupview when not shown in a popover? Did the same happen for iOS?

I didn't notice that, thanks for pointing it out. I'll open a new PR to update the insets and add you to it. iOS wasn't affected at all, but I'll retest on all platforms/presentation modes once I update the insets.

rolson:
[like]  Ryan Olson reacted to your message:
________________________________

PR #1215, opened 2025-07-10
`UtilityAssociationResultLabel` - Raise limit on title to two lines

dfeinzimer:
In order to better support usability for UNA data where associations have long titles Apollo is requesting that we bump the title line limit to 2 while setting a common height for all rows, independent of whether they need the space for two lines or not.

@CalebRas, if we want to opt PopupView out of this change I can find a way to do that.

Screenshot 2025-07-10 at 2 33 37 PM
Screenshot 2025-07-10 at 2 33 48 PM
Simulator Screenshot - Apple Vision Pro - 2025-07-10 at 14 36 32
Simulator Screenshot - Apple Vision Pro - 2025-07-10 at 14 36 57
Simulator Screenshot - iPhone 16 - 2025-07-10 at 14 10 07
Simulator Screenshot - iPhone 16 - 2025-07-10 at 14 10 16
Simulator Screenshot - iPhone 16 - 2025-07-10 at 14 40 16
Simulator Screenshot - iPhone 16 - 2025-07-10 at 14 40 22

### DIFF
--- a/Sources/ArcGISToolkit/Common/UtilityAssociationResultLabel.swift
+++ b/Sources/ArcGISToolkit/Common/UtilityAssociationResultLabel.swift
@@ -20,6 +20,12 @@ struct UtilityAssociationResultLabel: View {
     /// The utility association result to display.
     let result: UtilityAssociationResult
     
+    /// A scaled ideal frame height that allows for two lines of title text.
+    ///
+    /// At standard (non-accessible) system sizing, labels with two lines of title text in the body font and
+    /// a single line of detail text in the caption2 font have a height of 55.666.
+    @ScaledMetric(relativeTo: .body) private var frameHeight = 56
+    
     var body: some View {
         HStack {
             result.association.kind.icon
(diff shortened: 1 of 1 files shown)


rolson:
To me, this change doesn't make sense for a few reasons.

- From a design perspective I can see how it would provide some consistency in the feature form case. But it makes the PopupView look really bad. I definitely don't like it for Popup. In terms of feature form, why can't the row just grow to the size of the content? and maybe max it out at a certain number of lines (ie 4 or 5)?

- From a code perspective, this isn't a great solution because it requires a heuristic, hard coding a value, and scaling of the hard-coded value. I don't know if that is really the best choice from a code maintenance perspective. It also doesn't seem resilient to operating system changes.

rolson (changes requested):
I don't approve for PopupView. And for feature form, I think the team can discuss with the designer more. I'd personally lean towards making the row height dynamic.

mhdostal:
> I'd personally lean towards making the row height dynamic.

This was specifically called out by @PeggyDu09 as being undesirable. _...longer titles might take up too much space and make the screen feel cluttered or overwhelming._

The "two rows with middle truncation" option was deemed the best middle ground, along with vertically centering single row text. The other option (aside from fully showing all lines of text, no matter their height), was a single line of text with middle truncation.

rolson (approved):

PeggyDu09:
I'm comfortable with both options. Here’s a quick summary of our discussion:

Option 1: Dynamic height

- Pros: Full visibility of the title
- Cons: Longer titles may take up excessive space, pushing other items out of view


Option 2: Max 2-line title with middle truncation

- Pros: Visual consistency, better scan-ability, and more rows visible on screen
- Cons: Requires hardcoding the row height. Hide some information in the title.

Based on our discussion this morning, we're leaning toward using dynamic height with a maximum of 4 lines. This allows users to access more critical information without overwhelming the layout. And helps prevent long titles from dominating the screen.

mhdostal (approved):
Looks good to me.  @PeggyDu09 Should take a look once implemented and make sure it's OK with her.

CalebRas (approved):

dfeinzimer:
Thank you all!

PR #1220, opened 2025-07-12
`FeatureFormView` - Update example and tutorial for 200.8

dfeinzimer:
Apollo 1313


Examples/Examples/FeatureFormExampleView.swift
```diff
@@ -17,103 +17,85 @@ import ArcGISToolkit
...
 extension FeatureFormExampleView {
+    /// An error encountered while submitting edits.
+    enum SubmissionError: LocalizedError {
+        case applyEdits(any Error)
+        case other(String)
+        
+        var errorDescription: String? {
+            switch self {
+            case .applyEdits(let error):
+                String(reflecting: error)
```
philium, line 78: From the [documentation](https://developer.apple.com/documentation/swift/string/init(reflecting:)): "Creates a string with a detailed representation of the given value, **suitable for debugging.**" (emphasis mine). Using this as a user-facing string isn't modeling best practices.

Examples/Examples/FeatureFormExampleView.swift
```diff
@@ -17,103 +17,85 @@ import ArcGISToolkit
...
-            isApplyingEdits = false
-            updateSubmitButtonVisibility()
-        }
+        editsAreBeingApplied = true
+        defer { editsAreBeingApplied = false }
         
         for table in editedTables {
             guard let database = table.serviceGeodatabase else {
-                alertError = "No geodatabase found."
+                submissionError = .other("No geodatabase found.")
```
philium, line 94: Instead of updating `submissionError` in multiple places in this method, it seems like it would make more sense to make `applyEdits()` throwing and have the caller update `submissionError` with the thrown error. That would also fit more with the spirit of guard, which is to exit the current scope without doing additional work.

Examples/Examples/FeatureFormExampleView.swift
```diff
@@ -126,42 +108,31 @@ extension FeatureFormExampleView {
...
-    /// Identifies features, if any, at the current screen point.
-    /// - Parameter proxy: The proxy to use for identification.
-    /// - Returns: The first identified feature in a layer.
-    private func identifyFeature(with proxy: MapViewProxy) async -> ArcGISFeature? {
-        guard let identifyScreenPoint else { return nil }
-        let identifyLayerResults = try? await proxy.identifyLayers(
+    /// Opens a form for the first feature found at the point on the map.
+    /// - Parameter point: The point to run identify at on the map.
+    /// - Parameter map: The map to identify on.
+    private func makeFeatureForm(point: CGPoint, map: MapViewProxy) async {
```
philium, line 124: The name of the second parameter should be `mapView`. Also, the doc should be reformatted to use the `Parameters` delimiter.

Examples/Examples/FeatureFormExampleView.swift
```diff
@@ -126,42 +108,31 @@ extension FeatureFormExampleView {
...
+        guard let identifyScreenPoint else { return }
+        let identifyLayerResults = try? await map.identifyLayers(
             screenPoint: identifyScreenPoint,
             tolerance: 10
         )
-        return identifyLayerResults?.compactMap { result in
+        if let feature = identifyLayerResults?.compactMap({ result in
             result.geoElements.compactMap { element in
```
philium, line 131: Using `compactMap` and then getting the first value means that an intermediate array is built up and then immediately thrown away. There are multiple other ways of doing this that are more efficient.

Examples/Examples/FeatureFormExampleView.swift
```diff
@@ -170,24 +141,63 @@ extension FeatureFormExampleView {
     /// A Boolean value indicating whether general form workflow errors are presented.
     private var alertIsPresented: Binding<Bool> {
```
philium, line 142: This should just be a state property. Which an error is presented and which error is presented are two different properties of the view's state.

Examples/Examples/FeatureFormExampleView.swift
```diff
@@ -170,24 +141,63 @@ extension FeatureFormExampleView {
...
+                if case .savedEdits = editingEvent,
+                   let table = featureForm?.feature.table as? ServiceFeatureTable,
+                   !editedTables.contains(where: { $0 === table }) {
+                    editedTables.append(table)
+                }
+            }
+    }
+    
     /// A Boolean value indicating whether the form is presented.
     private var featureFormViewIsPresented: Binding<Bool> {
```
philium, line 165: This too should just be state.

Examples/Examples/FeatureFormExampleView.swift
```diff
@@ -170,24 +141,63 @@ extension FeatureFormExampleView {
...
+    /// The button used to apply edits made in forms.
+    @ViewBuilder private var submitButton: some View {
+        let databases = editedTables.compactMap(\.serviceGeodatabase)
+        let localEditsExist = databases.contains(where: \.hasLocalEdits)
+        if !featureFormViewIsPresented.wrappedValue, localEditsExist {
+            Button("Submit") {
+                Task {
+                    await applyEdits()
+                }
+            }
```
philium, line 184: As far as I can tell, this button could be spammed should it be disabled when edits are being applied (`.disabled(editsAreBeingApplied)`)?

Examples/Examples/FeatureFormExampleView.swift
```diff
@@ -170,24 +141,63 @@ extension FeatureFormExampleView {
...
+        }
+    }
+    
+    /// Overlay content that indicates the form is being submitted to the user.
+    @ViewBuilder private var submittingOverlay: some View {
+        if editsAreBeingApplied {
+            HStack(spacing: 5) {
+                ProgressView()
+                    .progressViewStyle(.circular)
+                Text("Submitting")
```
philium, line 194: `ProgressView` has an initializer for this:

```suggestion
                ProgressView("Submitting")
```

Sources/ArcGISToolkit/Documentation.docc/Resources/FeatureFormView/FeatureFormViewTutorialStep11.swift
```diff
@@ -0,0 +1,7 @@
+NavigationStack {
+    // App-level navigation
+    …
```
philium, line 3: The ellipsis character is good for localization, but it isn't great as a code placeholder when displayed with a monospaced font. `...` is easier to read and understand than `…`.

Same elsewhere.

philium (commented):
Also, and this is not a big deal, but still it can improve readability: there is no need for a blank line in between state properties. Views are just structs and we typically don't separate fields of a struct with blank lines. Instead, use blank lines for grouping properties, e.g. separating state properties from non-state properties.

Examples/Examples/FeatureFormExampleView.swift
yo1995: The error handling in sample is not a best practice, but since the `errorAlert` is also a modifier that we "hide" from the user, it is less likely to propagate into user's codebase.
(At least for me, when I copy paste from a sample into a different project and see the missing `errorAlert` compile error, I remove the modifier - thus the reflection won't be there)

Examples/Examples/FeatureFormExampleView.swift
dfeinzimer: @mhdostal, any preference between the two approaches? The fewer lines of code is nice...

Examples/Examples/FeatureFormExampleView.swift
dfeinzimer: Done in [fe487f6](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/1220/commits/fe487f647cd804c0cc680a9e910709543e99aa98)

philium (approved):

Examples/Examples/FeatureFormExampleView.swift
```diff
@@ -17,104 +17,87 @@ import ArcGISToolkit
...
-                .alert("Error", isPresented: alertIsPresented) {
-                } message: {
-                    if let error = alertError {
-                        Text(error)
-                    }
+                .sheet(isPresented: $featureFormViewIsPresented) {
+                    featureForm = nil
+                } content: {
+                    featureFormView
                 }
```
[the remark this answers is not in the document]
mhdostal: I don't recall. Is there something preventing us from doing it now? Although it might be better to wait until Sprint X, as we want to solidify everything now prior to the release.

Examples/Examples/FeatureFormExampleView.swift
```diff
@@ -126,66 +109,82 @@ extension FeatureFormExampleView {
...
+                if case .savedEdits = editingEvent,
+                   let table = featureForm?.feature.table as? ServiceFeatureTable,
+                   !editedTables.contains(where: { $0 === table }) {
+                    editedTables.append(table)
+                }
             }
+    }
+    
+    /// The button used to dismiss the submission error alert.
+    private var okButton: some View {
```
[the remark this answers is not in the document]
mhdostal: I'm fine with it as is.

mhdostal (approved):

PR #1225, opened 2025-07-16
`FeatureFormView` - Revise `GroupFormElementView.Header` appearance

dfeinzimer:
Apollo 1278

See issue for current and proposed screenshots.

### DIFF
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FormElements/GroupFormElementView.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FormElements/GroupFormElementView.swift
@@ -45,14 +45,9 @@ struct GroupFormElementView<Content>: View where Content: View {
                         .padding(.leading, 16)
                 }
             } label: {
-                VStack {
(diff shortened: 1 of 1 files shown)


rolson:
To me this looks a bit cramped on iOS with the subtitle. @PeggyDu09 can you take a look at the spacing here please?

image

On macOS, the spacing between the down carat and the text seems a bit cramped here:
image

@PeggyDu09 can you investigate this as well please?

rolson:
> This is a vanilla VStack containing two Texts. We could add some padding or spacing, but would have to pick a static or scaled value. If the font size is an issue we can consider bumping it up, but this matches the font used for the details line in a utility association entry.

Maybe just something to follow up on later. Is there a system control that does this in a more standard way? DisclosureGroup? Something else?

PeggyDu09:
I found a similar design for iOS. Here are the measurements and font information from Apple's toolkit:
Screenshot 2025-07-17 at 1 39 21 PM

For Mac, we could adopt a similar design approach. 
Screenshot 2025-07-17 at 2 00 17 PM

PR #1226, opened 2025-07-16
`FeatureFormView`/`PopupView` - Use common model for loading utility associations, improve related UI in FFV

dfeinzimer:
Apollo 1312

- Creates new `UtilityAssociationsElement` protocol (similar to `AttachmentsFeatureElement`)
- Moves `AssociationsFilterResultsModel` into new shared locations and refactors it to use `UtilityAssociationsElement`
- Converts `UtilityAssociationsFormElementView` and `UtilityAssociationsPopupElementView` to use the new shared `AssociationsFilterResultsModel`
- Moves two `Text` labels into a shared location to be used by both `UtilityAssociationsFormElementView` and `UtilityAssociationsPopupElementView`
- Moves `UtilityAssociationResultLabel` into the new shared folder at `Common/UtilityAssociationsElement/`

| Before | After |
|---|---|
|  |  |


Sources/ArcGISToolkit/Common/UtilityAssociationsElement/UtilityAssociationsElement.swift
```diff
@@ -0,0 +1,24 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ArcGIS
+
+/// Common properties for utility associations elements.
+protocol UtilityAssociationsElement: Sendable {
+    var associationsFilterResults: [UtilityAssociationsFilterResult] { get async throws }
```
CalebRas, line 19: Does this need a doc comment? Feel free to ignore if you don't think so.

Sources/ArcGISToolkit/Extensions/SwiftUI/Text.swift
```diff
@@ -55,4 +64,16 @@ extension Text {
             comment: "A label for a button allowing the user to retry an operation."
         )
     }
+    
+    static func makeErrorFetchingFilterResultsMessage(_ error: any Error) -> Self {
```
CalebRas, line 68: Same as above with this.

CalebRas (approved):
Looks good! Nice change

mhdostal (approved):

Sources/ArcGISToolkit/Extensions/SwiftUI/Text.swift
dfeinzimer: Phil R pointed out that since this method is static, it is not a factory method, and therefore it should not start with "make". I've renamed it and added doc.

CalebRas (approved):

mhdostal (approved):

PR #1229, opened 2025-07-18
`FeatureFormView` - Always use custom nav stack back button

dfeinzimer:
I discovered the native NavigationStack back button has a long-press context menu which lets a user jump to an arbitrary item:

<p align="center">
  
</p>

It was Apollo's intention to net yet support navigating to arbitrary items in the navigation path.

This PR changes the system navigation button to be always disabled and replaced by the custom navigation back button.


mhdostal (approved):

dfeinzimer:
> Does it work correctly?
> Why?

Over the course of the release, we had discussed adding a "return-to-home" button, but ultimately left it in an undecided state because we weren't aware that this was a supported pattern in iOS navigation.

We definitely need to override the button regardless when the current has form edits. When then are no edits, the implications are less clear. I think it is likely okay, but opening it up at this point in the release isn't preferable.

If we proceed with this PR, I will open an Apollo issue to revisit this for the next release.

PR #1230, opened 2025-07-18
`FeatureFormView` - Track presented form with preferences

dfeinzimer:
I observed the new `onFeatureFormChanged(perform:)` modifier unexpectedly firing in our test app, immediately after using the close button. I traced it back to `onAppear` firing in `EmbeddedFeatureFormView`.

This switches `FeatureFormView` over from using `onAppear`, to a preference, which does not exhibit the issue and matches `PopupView`'s tracking implementation.

Interestingly, I could only reproduce the issue in the micro-app when using an Inspector and not a Sheet. Nonetheless, I like that this aligns FeatureFormView with PopupView in another area.

### DIFF
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView.swift
@@ -145,7 +145,7 @@ public struct FeatureFormView: View {
                             .navigationTitle(result.filter.title, subtitle: embeddedFeatureFormViewModel.title)
                             .preference(
                                 key: PresentedFeatureFormPreferenceKey.self,
-                                value: .init(featureForm: embeddedFeatureFormViewModel.featureForm)
+                                value: .init(object: embeddedFeatureFormViewModel.featureForm)
                             )
                         case let .utilityAssociationGroupResultView(result, embeddedFeatureFormViewModel):
                             UtilityAssociationGroupResultView(
@@ -269,7 +269,7 @@ public struct FeatureFormView: View {
             .environment(\.validationErrorVisibilityInternal, $validationErrorVisibilityInternal)
             .onPreferenceChange(PresentedFeatureFormPreferenceKey.self) { wrappedFeatureForm in
                 guard let wrappedFeatureForm else { return }
-                formChangedAction(wrappedFeatureForm.featureForm)
+                formChangedAction(wrappedFeatureForm.object)
             }
         }
     }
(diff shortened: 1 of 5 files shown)


Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView.swift
```diff
@@ -154,11 +159,12 @@ public struct FeatureFormView: View {
...
                             .navigationTitle(result.name, subtitle: embeddedFeatureFormViewModel.title)
-                            .onAppear {
-                                formChangedAction(embeddedFeatureFormViewModel.featureForm)
-                            }
                         }
                     }
+                    .preference(
+                        key: PresentedFeatureFormPreferenceKey.self,
+                        value: .init(featureForm: rootFeatureForm)
+                    )
```
CalebRas, line 167: With popup, I was able to get away with just [setting the key once](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/v.next/Sources/ArcGISToolkit/Components/Popups/EmbeddedPopupView.swift#L63) inside of `EmbeddedPopupView`. It seems like these three `preference` modifiers could be replaced with a similar approach, is that not the case?

Sources/ArcGISToolkit/Components/FeatureFormView/PresentedFeatureFormPreferenceKey.swift
```diff
@@ -0,0 +1,37 @@
...
+/// A preference key that specifies the feature form currently presented in the `FeatureFormView` navigation stack.
+struct PresentedFeatureFormPreferenceKey: PreferenceKey {
+    /// A wrapper for making a feature form equatable.
+    struct EquatableFeatureForm: Equatable {
+        let featureForm: FeatureForm
+        
+        static func == (lhs: Self, rhs: Self) -> Bool {
+            lhs.featureForm === rhs.featureForm
+        }
+    }
```
CalebRas, line 27: Consider replacing this and the one in `PopupView` with a shared type, maybe something like `struct EquatableBox<Wrapped: AnyObject & Sendable>: Equatable`. It is probably not a big deal either way, though, so up to you.

Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView.swift
dfeinzimer: Nope that works, my original thinking was it would be cleaner to set them all within the same file, but consolidating two locations into one on EmbeddedFeatureFormView seems nicer.

CalebRas (approved):

mhdostal (approved):

PR #1231, opened 2025-07-18
Fix crash when trying to download an area when there is no internet connection

### LINKED ISSUE #1209: Bug in OfflineMapAreasView - Download area when offline
Hello everyone. We are using the OfflineMapAreasView in our app and have encountered a bug that leads to an app crash. The issue can be reproduced using your OfflineMapAreas example app by following these steps:

1. While online, download a map area.
2. Disconnect from the internet and open the offline map areas screen.
3. Tap on an item to open its detail page.
4. Tap the button to delete the area, then tap the button again to download it (which, of course, is not possible while offline) → App crash.

To avoid this, you could maybe consider hiding or disabling the download button after an area is deleted, especially when there's no internet connection, or at least handle the error gracefully.

rolson:
fixes #1209


des12437 (approved):
Looks good, just one comment.

des12437 (dismissed):

yo1995 (approved):

des12437 (approved):

PR #1232, opened 2025-07-18
Offline: Fix offline mode undetermined for no internet

### LINKED ISSUE #1210: Bug in OfflineMapAreasView - "No Map Areas"
Hello everyone. We are using the OfflineMapAreasView in our app and have encountered another bug. The issue can also be reproduced using your OfflineMapAreas example app by following these steps:

1. While online, open the example app.
2. Without opening the Offline Maps/Areas screen, disconnect from the internet.
3. Then open the Offline Maps screen.

At this point, the message "No Map Areas" is shown along with the option to add an on-demand map area. → Bug. The message is incorrect and misleading because, in this example, map areas are actually available .

We believe it would make more sense to display the message 'No Internet Connection' instead, since the app likely doesn't have access to the map area data while offline. In this case, we would expect the app to attempt a refresh for the available map areas once the user is back online (triggered by the button below the 'No Internet Connection' message). In our opinion, the message "No Map Areas" should only be shown when a map truly has no available areas—not when the app simply cannot retrieve them due to lack of connectivity.

yo1995:
## Description

Close #1210 . 

Currently, if a specific set of steps are followed:

1. While online, open the example app.
2. Without opening the Offline Maps/Areas screen, disconnect from the internet.
3. Then open the Offline Maps screen.

It would show the `onDemandMapAreasView` even if the mode for the offline component is undetermined

https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/472d484b0190ce12aa90f76fef800e73873cb5e8/Sources/ArcGISToolkit/Components/Offline/OfflineMapAreasView.swift#L158-L163

Instead, the user propose to show the `noInternetNoAreasView` when there is no Internet connection.

## Discuss

See some more comment in the code below.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Offline/OfflineMapViewModel.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/OfflineMapViewModel.swift
@@ -152,7 +152,7 @@ class OfflineMapViewModel: ObservableObject {
                 return
             }
             
-            if !hasAnyPreplannedMapAreas, isShowingOnlyOfflineModels {
+            if isShowingOnlyOfflineModels {
                 // In this case there are no preplanned or on-demand map areas and we can
                 // only show offline models (which there are none) because
                 // there is no internet. In that case we set the mode to
--- a/Sources/ArcGISToolkit/Components/Offline/OfflineMapViewModel.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/OfflineMapViewModel.swift
@@ -112,6 +112,15 @@ class OfflineMapViewModel: ObservableObject {
         
         // Note: We don't reset the mode once it is determined.
         
+        // Before loading models, check if the Internet is available.
+        do {
+            _ = try await offlineMapTask.preplannedMapAreas
+        } catch {
+            if error.isNoInternetConnectionError {
+                return
+            }
+        }
+        
         // First load preplanned map models.
         if mode == .undetermined || mode == .preplanned {
             await loadPreplannedMapModels()
@@ -126,9 +135,12 @@ class OfflineMapViewModel: ObservableObject {
         // undetermined.
         if mode == .undetermined || mode == .onDemand {
             await loadOnDemandMapModels()
-            // If there are any on-demand areas at all, and the mode is
-            // undetermined, then set mode to on-demand.
-            if mode == .undetermined, !onDemandMapModels.isEmpty {
+            // If the mode is undetermined, then set mode to on-demand.
+            // If there are on-demand map areas, the mode should be set to
+            // on-demand. If there are no on-demand map areas, we can assume
+            // that the user started with an empty map and want to add on-demand
+            // map areas.
+            if mode == .undetermined {
                 mode = .onDemand
             }
         }
--- a/Sources/ArcGISToolkit/Components/Offline/OfflineMapViewModel.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/OfflineMapViewModel.swift
@@ -146,7 +146,6 @@ class OfflineMapViewModel: ObservableObject {
             // Try to load on-demand map models.
             await loadOnDemandMapModels()
             
-            // Try to load on-demand map models.
             // If there are on-demand areas on device, set mode to on-demand.
             if !onDemandMapModels.isEmpty {
                 mode = .onDemand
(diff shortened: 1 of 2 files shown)


Sources/ArcGISToolkit/Components/Offline/OfflineMapViewModel.swift
```diff
@@ -112,6 +112,15 @@ class OfflineMapViewModel: ObservableObject {
...
         // Note: We don't reset the mode once it is determined.
         
+        // Before loading models, check if the Internet is available.
+        do {
+            _ = try await offlineMapTask.preplannedMapAreas
+        } catch {
+            if error.isNoInternetConnectionError {
+                return
+            }
+        }
```
yo1995, line 122: In order to check the internet connection, I need to make a network request and get the `URLError` to use `isNoInternetConnectionError` extension.

`onlineMap.retryLoad()` only make a network request when the current `loadStatus` isn't `loaded`, so it cannot be used here.

Sources/ArcGISToolkit/Components/Offline/OfflineMapViewModel.swift
```diff
@@ -126,9 +135,12 @@ class OfflineMapViewModel: ObservableObject {
...
             await loadOnDemandMapModels()
-            // If there are any on-demand areas at all, and the mode is
-            // undetermined, then set mode to on-demand.
-            if mode == .undetermined, !onDemandMapModels.isEmpty {
+            // If the mode is undetermined, then set mode to on-demand.
+            // If there are on-demand map areas, the mode should be set to
+            // on-demand. If there are no on-demand map areas, we can assume
+            // that the user started with an empty map and want to add on-demand
+            // map areas.
+            if mode == .undetermined {
```
yo1995, line 143: When there is no Internet connection, the method returns early with `mode == .undetermined`. If the method reaches this part, it means:

- network is connected
- mode is not preplanned, is undetermined

In this case, I think it is safe to assume the mode should be onDemand.

Sources/ArcGISToolkit/Components/Offline/OfflineMapViewModel.swift
```diff
@@ -112,25 +120,51 @@ class OfflineMapViewModel: ObservableObject {
...
+            
+            // Try to load on-demand map models.
+            // If there are on-demand areas on device, set mode to on-demand.
+            if !onDemandMapModels.isEmpty {
                 mode = .onDemand
+                return
+            }
+            
+            if !hasAnyPreplannedMapAreas, isShowingOnlyOfflineModels {
+                // In this case there are no preplanned map areas and we can
```
rolson, line 157:
```suggestion
                // In this case there are no preplanned or on-demand map areas and we can
```

Sources/ArcGISToolkit/Components/Offline/OfflineMapViewModel.swift
yo1995: This is no longer needed as the network connection is checked by `loadPreplannedMapModels()` method.

Sources/ArcGISToolkit/Components/Offline/OfflineMapViewModel.swift
yo1995: I applied Ryan's suggestion to add the `noInternetAvailable` and renamed `undetermined` to `ambiguous`

Sources/ArcGISToolkit/Components/Offline/OfflineMapViewModel.swift
```diff
@@ -112,6 +112,15 @@ class OfflineMapViewModel: ObservableObject {
         
         // Note: We don't reset the mode once it is determined.
         
+        // Before loading models, check if the Internet is available.
+        do {
+            _ = try await offlineMapTask.preplannedMapAreas
+        } catch {
+            if error.isNoInternetConnectionError {
+                return
```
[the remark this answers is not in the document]
yo1995: Mitigated by adding modes.

Sources/ArcGISToolkit/Components/Offline/OfflineMapViewModel.swift
```diff
@@ -112,25 +120,51 @@ class OfflineMapViewModel: ObservableObject {
...
-            if mode == .undetermined, !onDemandMapModels.isEmpty {
+            
+            // Try to load on-demand map models.
+            // If there are on-demand areas on device, set mode to on-demand.
+            if !onDemandMapModels.isEmpty {
                 mode = .onDemand
+                return
+            }
+            
+            if !hasAnyPreplannedMapAreas, isShowingOnlyOfflineModels {
```
rolson, line 155: I guess the `!hasAnyPreplannedMapAreas` isn't necessary since it's checked above.

```suggestion
            if isShowingOnlyOfflineModels {
```

rolson (approved):

des12437 (approved):

yo1995:
I'm doing a little bit more testing… Once it is done I'll merge this PR. Thanks for the patience!

rolson:
> I logged that in #1240 . Do you want me to create a swift issue instead?

I didn't see that, thanks

PR #1244, opened 2025-07-25
Improve Basemap Gallery example

philium:
When I was verifying the BasemapGallery component on Mac Catalyst for the release, I noticed a few things about the example:

- The presented sheet has no title, which is quite odd
- When the gallery is scrolled, there is a lot of vertical padding between the done button and the basemaps
  Screenshot 2025-07-25 at 4 06 25 PM
- The button that presents the sheet is actually a switch, which is a vestige from when the gallery overlayed the map view

This address all those problems. In addition, the padding that was in the example was moved into the component itself. To understand why that was necessary, see what happens when the gallery is scrolled with the padding in the example:

Screen Recording 2025-07-25 at 4 17 05 PM

The padding is not scrolled, which is unexpected. Moving the padding into the component fixes the problem:

Screen Recording 2025-07-25 at 4 15 44 PM


mhdostal (approved):
Thanks @philium !

PR #1170, opened 2025-05-28
[New] Location Button

rolson:
Adds a button to the toolkit that allows a user to show their current location.

The view can be positioned anywhere, with any background. This example uses a "regular material background" in a rounded rectangle.

- Tap once, location display starts, if using a system location datasource, it will request authorization, while starting a progress indicator is shown
- Starts in auto pan mode recenter
- Long press to get a context menu
- Switch to navigation mode, any autopan mode, or turn the datasource off
- If the location display fails to start, an warning signal is shown, tapping again will try to start the datasource again
- Panning turns autopan off, tapping the button again will turn on the last autopan mode that the user had selected (barring the `off` value)

More details
- The context menu is only shown on long press if the count of the auto pan options the user passes in is greater than 1
- The "last selected auto pan mode" starts out with a default value of the first value passed to the initializer in the `autoPanOptions` that matches a value in this list: `.recenter, .compassNavigation, .navigation`

Location Display Starting
Recenter auto pan mode
Long Press Context Menu
Navigation Mode
Location Display failed to start

PR #1253, opened 2025-07-30
location button: fix build errors on visionOS

yo1995:
## Description

This PR fixes the daily build error caused by visionOS not having compass. Related to #1170

### DIFF
--- a/Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
+++ b/Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
@@ -46,11 +46,18 @@ public struct LocationButton: View {
     var buttonIsDisabled: Bool {
         status == .starting || status == .stopping || isPerformingButtonAction
     }
-    
+
+#if os(visionOS)
+    /// The auto-pan modes that are selectable by the user.
+    private(set) var autoPanModes: [LocationDisplay.AutoPanMode] = [
+        .recenter, .off
+    ]
+#else
     /// The auto-pan modes that are selectable by the user.
     private(set) var autoPanModes: [LocationDisplay.AutoPanMode] = [
         .recenter, .compassNavigation, .off
     ]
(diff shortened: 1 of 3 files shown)


Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
```diff
@@ -46,11 +46,18 @@ public struct LocationButton: View {
...
+
+#if os(visionOS)
+    /// The auto-pan modes that are selectable by the user.
+    private(set) var autoPanModes: [LocationDisplay.AutoPanMode] = [
+        .recenter, .off
+    ]
+#else
     /// The auto-pan modes that are selectable by the user.
     private(set) var autoPanModes: [LocationDisplay.AutoPanMode] = [
         .recenter, .compassNavigation, .off
```
yo1995, line 58: Do I need to include `.navigation` in the autopan modes?

Tests/ArcGISToolkitTests/LocationButtonTests.swift
```diff
@@ -51,12 +56,12 @@ struct LocationButtonTests {
         #expect(button.contextMenuAutoPanOptions == [.off, .recenter])
         
         // Test when `off` is last
-        button = button.autoPanModes([.recenter, .compassNavigation, .off])
```
yo1995, line 54: For the testcases below, I simply removed or swapped `.compassNavigation` to let the testcase build on all platforms with small deviation to what the testcase tests originally.

If we want to keep the original testcase, I can add the compiler block instead.

Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
rolson: not for default, no

rolson (approved):
thank you Ting!

PR #1261, opened 2025-07-31
`FeatureFormView` - Fix attachment renaming

dfeinzimer:
Fixes a regression introduced in #1242 which causes the rename attachment alert dialogue to immediately close when typed into.

### DIFF
--- a/Sources/ArcGISToolkit/Common/AttachmentPreview.swift
+++ b/Sources/ArcGISToolkit/Common/AttachmentPreview.swift
@@ -124,14 +124,8 @@ struct AttachmentPreview: View {
             ),
             isPresented: $renameDialogueIsShowing
         ) {
-            TextField(text: $newAttachmentName) {
-                Text(
-                    "New name",
-                    bundle: .toolkitModule,
-                    comment: "A label in reference to the new name of a file, shown in a file rename interface."
-                )
-            }
-            .autocorrectionDisabled()
+            TextField(String.newName, text: $newAttachmentName)
+                .autocorrectionDisabled()
             Button("Cancel", role: .cancel) { }
             Button("OK") {
                 Task {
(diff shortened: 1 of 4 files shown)


Sources/ArcGISToolkit/Utility/Carousel.swift
```diff
@@ -22,7 +22,7 @@ struct Carousel<Content: View>: View {
     @State private var cellSize = CGSize.zero
     
     /// The identifier for the Carousel content.
-    let contentIdentifier = UUID()
+    let contentIdentifier = "content"
```
dfeinzimer, line 25: This is the point of failure. It needs to be a stable identity.

Sources/ArcGISToolkit/Common/AttachmentPreview.swift
```diff
@@ -124,14 +124,8 @@ struct AttachmentPreview: View {
...
-                    "New name",
-                    bundle: .toolkitModule,
-                    comment: "A label in reference to the new name of a file, shown in a file rename interface."
-                )
-            }
-            .autocorrectionDisabled()
+            TextField(String.newName, text: $newAttachmentName)
+                .autocorrectionDisabled()
             Button("Cancel", role: .cancel) { }
             Button("OK") {
```
des12437, line 130: Unrelated to the changes in this PR, the cancel and ok buttons can be localized by using `Button.cancel {}` and `Button.ok {}`.

des12437 (approved):

Sources/ArcGISToolkit/Common/AttachmentPreview.swift
dfeinzimer: Thank you, noted for followup!

mhdostal (approved):

PR #1262, opened 2025-07-31
`FeatureFormView` - Cancel attachment load on form dismissal 

dfeinzimer:
Apollo 1193

### DIFF
--- a/Sources/ArcGISToolkit/Common/AttachmentModel.swift
+++ b/Sources/ArcGISToolkit/Common/AttachmentModel.swift
@@ -69,7 +69,9 @@ internal import os
         self.thumbnailSize = thumbnailSize
         
         if attachment.isLocal {
-            load()
+            initialLoadTask = Task {
+                await load()
+            }
         } else {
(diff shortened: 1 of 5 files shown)


mhdostal (approved):

Sources/ArcGISToolkit/Common/AttachmentModel.swift
```diff
@@ -81,33 +86,39 @@ internal import os
         }
     }
     
+    deinit {
+        initialLoad?.cancel()
```
yo1995, line 90: Probably can use [take()](https://developer.apple.com/documentation/swift/optional/take()) to set it to `nil`.

yo1995 (approved):

Sources/ArcGISToolkit/Common/AttachmentModel.swift
dfeinzimer: Since I don't believe nulling out the task is required here I'll keep the current approach and merge as is. Thank you for the suggestion, though, it is interesting.

PR #1270, opened 2025-08-05
`PopupView` - Remove deprecated members

CalebRas:
### Description

This PR removes all of the deprecated public `PopupView` members since they will no longer be supported in 300:
- `init(popup:isPresented:)`
- `header(_:)`
- `showCloseButton(_:)`

There are no UI changes (when using the support init, `init(root:isPresented:)`).

### How To Test

Ensure the Popup View example and tests still perform as expected.


yo1995:
Changes look good. Do they go into v.next? I think we are waiting for an answer from devops on whether or not to branch release/200.8.1.

yo1995 (approved):

PR #1290, opened 2025-09-03
Compass: Improve accessibility and localization

philium:
This gives the compass an accessibility value, hides it from accessibility when hidden, and localizes the cardinal and intercardinal directions.


dfeinzimer (dismissed):
Looks great, just the one question.

pgruenler (dismissed):

PR #1293, opened 2025-09-04
Update UtilityAssociationResultLabel

dfeinzimer:
Related Apollo 1342, #1288 

Currently, the fractional value is shown in the trailing position. The design spec has been updated to move it into the detail line to accommodate the new association details button when used in FeatureFormViews.

The spec for displaying content visibility was also changed in the re-design.


CalebRas (dismissed):
LGTM!

CalebRas (approved):

PR #1297, opened 2025-09-09
`FeatureFormView` - Improve UI test case selection

dfeinzimer:
Improves the FeatureFormView UI tests by using launch arguments to select the test case. Replaces the current process of picking the desired case from a list.

Saves a little under 1 minute across the 34 FeatureFormView test cases.


dfeinzimer:
@chriswebb09 please review this when you have a moment.

PR #1299, opened 2025-09-17
`FeatureFormView` - UNA removal support follow-up

dfeinzimer:
In accordance with common-toolkit # 300 & 301 to prepare for #1177.

- Utility association form elements now show results for all authored filters, even if empty.
- When the last association is removed from a group result, the user is now only backed out to the filter result page instead of to the form. Test case 12.6 is updated accordingly.
- Association removal confirmation now always uses an alert instead of conditionally switching between a confirmation dialog and alert.

### DIFF
--- a/Sources/ArcGISToolkit/Common/UtilityAssociationsElement/AssociationsFilterResultModel.swift
+++ b/Sources/ArcGISToolkit/Common/UtilityAssociationsElement/AssociationsFilterResultModel.swift
@@ -49,10 +49,10 @@ final class AssociationsFilterResultsModel {
             guard !Task.isCancelled, let self else { return }
             let result = await Result {
                 let allResults = try await element.associationsFilterResults
-                if includeEmptyFilterResults {
-                    return allResults
+                return if includeEmptyFilterResults {
+                    allResults
                 } else {
-                    return allResults.filter { $0.resultCount > 0 }
+                    allResults.filter { $0.resultCount > 0 }
                 }
             }
             withAnimation { self.result = result }


Test Runner/UI Tests/FeatureFormViewTests.swift
```diff
@@ -1717,9 +1717,9 @@ final class FeatureFormViewTests: XCTestCase {
         let transformerButton = app.buttons["Transformer, High"]
         
 #if targetEnvironment(macCatalyst)
-        let deleteButton = app.menuItems["delete"]
+        let inlineRemoveAssociationButton = app.menuItems["remove_association"]
 #else
-        let deleteButton = app.buttons["Delete"]
```
dfeinzimer, line 1722: This change is not directly related to the design spec change. It should've been included in #1288 but I missed it.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FormElements/UtilityAssociationsFormElement/AssociationRemovalConfirmation.swift
```diff
@@ -52,32 +50,19 @@ private struct AssociationRemovalConfirmationDialog: ViewModifier {
...
-            content
-                .confirmationDialog(
-                    confirmationTitle,
-                    isPresented: $isPresented,
-                    titleVisibility: .visible
-                ) {
-                    actions
-                } message: {
-                    confirmationMessage
-                }
```
dfeinzimer, line 65: In iOS 26, confirmation dialogs adopt the popover style and have a misplaced arrow when used on list items.

Simulator Screen Recording - iPhone 17 - 2025-09-17 at 15 05 33

 In iOS 18 this is only an issue on iPadOS. 

Choosing to go with alerts instead for both platforms now as a result.

mhdostal (approved):

dfeinzimer:
@yo1995, please review when you have a moment. Thanks!

Sources/ArcGISToolkit/Common/UtilityAssociationsElement/AssociationsFilterResultModel.swift
```diff
@@ -39,10 +45,15 @@ final class AssociationsFilterResultsModel {
...
+        task = Task { [weak self, includeEmptyFilterResults] in
             guard !Task.isCancelled, let self else { return }
             let result = await Result {
-                try await element.associationsFilterResults.filter { $0.resultCount > 0 }
+                let allResults = try await element.associationsFilterResults
+                if includeEmptyFilterResults {
+                    return allResults
+                } else {
+                    return allResults.filter { $0.resultCount > 0 }
+                }
```
yo1995, line 56:
```swift
                return if includeEmptyFilterResults {
                    allResults
                } else {
                    allResults.filter { $0.resultCount > 0 }
                }
```
A Swift 5.9 [feature](https://www.hackingwithswift.com/swift/5.9/if-switch-expressions) that I saw SDK uses. Up to you.

yo1995 (approved):
Sorry for late response - missed the notification!

yo1995 (approved):

PR #1304, opened 2025-10-01
AttachmentPhotoPicker: Remove Sendable conformance

yo1995:
This type seems to have been annotated as Sendable in SwiftUI/PhotosUI, so remove our conformance to get rid of the warning.

See the type's [doc](https://developer.apple.com/documentation/photosui/photospickeritem#conforms-to).

Screenshot 2025-10-01 at 3 24 04 PM
Screenshot 2025-10-01 at 3 23 16 PM

Tested on Xcode 16.4 and 26.0.1 without warning.


dfeinzimer (approved):
Nice!

PR #1312, opened 2025-10-16
Authenticator: Sheets -> Alerts

rolson:
Authenticator: Sheets -> Alerts

A few reasons for this:
- Aligns closer with what Field Maps is using (they exclusively use Alerts for authenticator)
- Shows standardized UI, which allows us to use standard Apple APIs to allow users to make a choice.
- Adapts to the OS version better (for example with alerts: on iOS 26 the text is left-aligned, on iOS 18 the text is centered)
- Better matches the system look and feel
- Less code, simpler code
- Less instances of sheets showing over sheets

Previous implementation:
- Related [issue](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/381)
- I think it's probably ok to go back to alerts. Apps team originally reported the [related issue](#315), but have also since gone to exclusively using alerts.
- True that alerts cannot display over alerts, which sheets can. However, Alerts are highly transient, so this is not as much of an issue. Also authentication usually happens in spurts, mostly upon login.

UPDATE: "Skip" button replaced with "Ignore" button in UI below.

|   | Sheet | Alert iOS 26 | Alert iOS 18 |
| :--- | :---: | :---: | :---: |
| Trust Host | image | image | image |
| Certificate Required | image | image | image |
| Certificate Import Error | image | image | image |


dfeinzimer (approved):
Tested workflows with the certificate trust and picker - working well.

> please review and also give your thoughts on this approach.

I think the alerts look good! Also tested on Catalyst and visionOS.

njarecha (commented):
I'm okay with moving back to using Alert. However, we should ensure:
1. We won’t run into the same issue that caused us to move from `alerts` to `sheets` in the first place (see PR: https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/381).
2. `CredentialInputModifier` is currently using a sheet. It should also be updated to use Alert.

rolson:
> I'm okay with moving back to using Alert. However, we should ensure:
> 
> 1. We won’t run into the same issue that caused us to move from `alerts` to `sheets` in the first place (see PR: #381).

I don't think that's possible to guarantee. But as mentioned, alerts are much more transient than sheets, so the occurrence will be much less likely to occur as sheet over sheet.

> 2. `CredentialInputModifier` is currently using a sheet. It should also be updated to use Alert.

I've asked field maps for information on how they are doing this and have yet to get a response. But it can happen later, separate from this PR.

njarecha (approved):
Looks good

rolson:
> > I'm okay with moving back to using Alert. However, we should ensure:
> > 
> > 1. We won’t run into the same issue that caused us to move from `alerts` to `sheets` in the first place (see PR: [Authentication Alerts -> Authentication Sheets #381](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/381)).
> 
> I don't think that's possible to guarantee. But as mentioned, alerts are much more transient than sheets, so the occurrence will be much less likely to occur as sheet over sheet.

I'm testing this and finding that alerts are actually queued up. Where I'm seeing issues is interaction between alerts and sheets. Showing an alert with a sheet displayed will dismiss the sheet (when the alert is not a child of the sheet content). Sometimes neither appears depending on timing. So it might be actually better to move to alerts as much as possible to avoid issues like that.

rolson:
> image
> Noticed the “Allow” button looks destructive (red), and this could be confusing since it’s actually the affirmative action. Might be better to use a neutral or default style here?
> 
> [Apple’s HIG](https://developer.apple.com/design/human-interface-guidelines/alerts#:~:text=If%20there%E2%80%99s%20a%20destructive%20action,any%20button%20the%20default%20button.) recommends making both buttons default to encourage people to read an alert carefully.

Interesting, thanks for pointing that out. I will fix this in a follow-up PR.

rolson:
Some terminology from Apple:

image

image

PR #1316, opened 2025-10-21
AR: Fix `ARCoachingOverlay` safe area for iOS 26

des12437:
Related to `swift/7444`

Fixes a UI issue in iOS 26 where the `ARCoachingOverlay` does not extend into the safe area.


CalebRas (approved):

yo1995 (approved):
Don't have an iOS 26 device at hand so assume it works 😬

PR #1318, opened 2025-10-22
CredentialInputView: Sheet -> Alert

rolson:
Follow up to #1312 

Note: I also can confirm that the alerts work with other alerts, but the current sheet does not. It will get dismissed without responding to the challenge and result in a deadlocked challenge queue.

|   | Sheet | Alert iOS 26 | Alert iOS 18 |
| :--- | :---: | :---: | :---: |
| Credential Input | image | image | image |

### DIFF
--- a/Sources/ArcGISToolkit/Utility/CredentialInputView.swift
+++ b/Sources/ArcGISToolkit/Utility/CredentialInputView.swift
@@ -119,7 +119,6 @@ struct CredentialInputModifier: ViewModifier {
                         .padding(.horizontal)
                 }
                 .disabled(!isContinueEnabled)
-                
             } message: {
                 Text(message)
             }


rolson:
NOTE: I don't see why the onAppear or the focus state code is needed with the alert so it was removed.

Sources/ArcGISToolkit/Utility/CredentialInputView.swift
```diff
@@ -83,81 +82,51 @@ struct CredentialInputModifier: ViewModifier {
...
+                
+                Button {
+                    isPresented.wrappedValue = false
+                    continueAction.handler(username, password)
+                } label: {
+                    Text(continueAction.title)
+                        .padding(.horizontal)
+                }
+                .disabled(!isContinueEnabled)
+
```
dfeinzimer:
```suggestion
```

dfeinzimer (dismissed):

PR #1323, opened 2025-10-29
`FeatureFormView` - Page the set of UNA candidates

dfeinzimer:
Apollo 1477

Currently, we only fetch the first page of candidates, with page size being determined by the service.

With this change, we check the result for the existence of "next query parameters", which, if used, will return the next page of candidates. If those parameters are populated, we add a progress view to the bottom of the candidates list. Due to the laziness of a SwiftUI list, once the progress is visible, a task attached it to will (a), run and fetch the next set of candidates, and (b) also refresh the "next query parameters" for any additional pages.


mhdostal (approved):

mhdostal (approved):

dfeinzimer:
Thanks!

PR #1324, opened 2025-10-31
Fix timing issue with navigation path

mhdostal:
There was a crash when closing the feature form after UNA editing. Adding a check to verify the `navigationPath` was not empty fixed it. If you put break statements in while debugging when the `navigationPath` changed, the crash didn't happen; nor did it crash if you manually backed out to the main FF view.

image

Crash happened on line 81 because the `navigationPath` was empty.


mhdostal:
@pgruenler Can you give this a quick review? Thanks.

pgruenler (approved):

mhdostal:
Thanks @pgruenler & @dfeinzimer

PR #1325, opened 2025-11-03
Allow specifying update mode of map areas taken offline

rolson:
issue 7499

cc @mhdostal


des12437 (approved):

yo1995 (approved):
For Epic app's use, we only need the same update mode for the entire app so this should work fine. We can tweak more when other usecases come up.

PR #1332, opened 2025-11-06
Don't show UNA "Locate" button if feature has no geometry

mhdostal:
Apollo #1467

When viewing or adding associated elements, do not show the locate button if the feature (actually the feature's table) doesn't support geometries (i.e., non-spatial data).

Viewing associations:
| Before | After |
|----------|--------|
|image|image|

Adding associations:

| Before | After |
|----------|--------|
|image|image|

### DIFF
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/ShowOnMapButton.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/ShowOnMapButton.swift
@@ -38,8 +38,6 @@ extension FeatureFormView {
                         Image(systemName: "scope")
                     }
                 }
-            } else {
-                EmptyView()
             }
         }
     }


Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/ShowOnMapButton.swift
```diff
@@ -23,18 +23,23 @@ extension FeatureFormView {
...
+                            "Show On Map",
+                            bundle: .toolkitModule,
+                            comment: "A label for a button to show the indicated feature on the map."
+                        )
+                    } icon: {
+                        Image(systemName: "scope")
+                    }
                 }
+            } else {
+                EmptyView()
```
CalebRas:
```suggestion
```

Is using an explicit `EmptyView` here necessary? I believe the `@Viewbuilder` on `View.body` will use one automatically.

mhdostal: Works like a dream, thanks!

dfeinzimer (approved):

CalebRas (approved):

PR #1335, opened 2025-11-12
Use different method to hide the "Locate" button

mhdostal:
Puneet suggesting using a check for `geometry != nil` instead of checking the table's `hasGeometry` property. The thought was this will also catch the case where a spatial feature's geometry was `nil`.

Follow up to Apollo 1467.


dfeinzimer (approved):
Looks good!

PR #1337, opened 2025-11-17
Update translations

dfeinzimer:
Updates strings files with latest translations.

### DIFF
--- a/Sources/ArcGISToolkit/Resources/ar.lproj/Localizable.strings
+++ b/Sources/ArcGISToolkit/Resources/ar.lproj/Localizable.strings
@@ -2,7 +2,7 @@
 utility element (first variable) and a percentage
 (second variable) indicating the relative location along
 an edge where a utility association is (logically) located. */
-"%@, %@" = "بيت_%@, %@_____________لاحقة";
+"%@, %@" = "%@، %@";
 
 /* A label declaring the number of starting points selected for a utility network trace. */
 "%lld selected" = "تم تحديد %lld";
@@ -21,7 +21,7 @@ content on a remote host. The variable is the host that prompted the challenge. 
 "A map area cannot be downloaded at this time." = "يتعذر تنزيل منطقة الخريطة في هذا الوقت.";
 
 /* A label for a button to add a new utility association. */
-"Add" = "بيت_Add_______لاحقة";
+"Add" = "إضافة";
 
 /* A label for a button to add a new utility network association. */
 "Add Association" = "إضافة اقتران";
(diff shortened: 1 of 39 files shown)


Sources/ArcGISToolkit/Resources/ar.lproj/Localizable.strings
```diff
@@ -120,14 +159,27 @@ for the AR experience. */
...
+"compass-button-accessibility-label" = "البوصلة";
+
+/* A label for the "Connectivity" utility association kind. */
+"Connectivity" = "الاتصال";
+
+/* A label for the "Containment" utility association kind. */
+"Containment" = "الاحتواء";
+
+/* A label indicating whether content in the new
+association is visible or not. */
```
njarecha:
```suggestion
/* A label indicating whether content in the new association is visible or not. */
```

Sources/ArcGISToolkit/Resources/ar.lproj/Localizable.strings
```diff
@@ -299,10 +357,22 @@ trace operation. */
 /* A label for a text entry field that allows the user to filter a list of values by name. */
 "Filter" = "عامل تصفية";
 
+/* A label for a search field to filter utility 
+asset types by name. */
```
njarecha:
```suggestion
/* A label for a search field to filter utility asset types by name. */
```

Sources/ArcGISToolkit/Resources/ar.lproj/Localizable.strings
```diff
@@ -299,10 +357,22 @@ trace operation. */
...
+"Filter asset types by name" = "تصفية أنواع الأصول حسب الاسم";
+
+/* A label for a search field to filter utility 
+association candidate features by name. */
```
njarecha:
```suggestion
/* A label for a search field to filter utility association candidate features by name. */
```

njarecha (commented):
Reviewed only few string. There seems to be no consistency when comments are wrapped to next line. There are so many long comments strings in the single line.

dfeinzimer:
> There seems to be no consistency when comments are wrapped to next line. There are so many long comments strings in the single line.

@njarecha Since these files are autogenerated from Swift source, we'd want to standardize the line lengths directly in the places they're defined. From some testing I've done, if we do that now and regenerate each file, the translations for each key are overwritten back to the English default.

It seems the best course of action would be to proceed with this PR as is and open a sprint x issue to standardize the line length of localization comments project-wide. Then, in the next localization round, the changes will be applied to each strings file by the localization team's tooling.

njarecha (approved):

PR #1339, opened 2025-11-18
`AttachmentsFeatureElementView` - Fix cancellation error on initial load

dfeinzimer:
Fixes a cancellation error on initial load of `AttachmentsFeatureElementView`.

Originating report: https://community.esri.com/t5/swift-maps-sdk-questions/popupview-does-not-display-attachments-if-outside/m-p/1666246

PR #1340, opened 2025-11-20
(Cherry-pick) AttachmentsFeatureElementView - Fix cancellation error on initial load

dfeinzimer:
Cherry-picks #1339


dfeinzimer:
> Correct me if I'm wrong, but this will only be picked up if devs select that branch with SPM or if they clone that branch and build it locally.

My understanding is that 200.8.1 will be tagged against `lts.next`.

PR #1283, opened 2025-08-15
`FeatureFormView`: Use List

dfeinzimer:

PR #1343, opened 2025-12-09
`FeatureFormView` - Use `List` when not running UI tests

dfeinzimer:
As part of #1283 we need a strategy to handle the impacts on the `FeatureFormView` UI tests. 

The tests currently rely on the fact that all the form elements are available in the accessibility hierarchy, even if they're "below the fold." By switching to a `List` this will no longer be true, the form elements aren't programmatically accessible until they're visible.

A few of the approaches I've looked into:
- Programmatically scrolling until the element being tested is visible.
  - This has proven to be unreliable, especially on Mac Catalyst, with the distance scrolled being non-deterministic. It also slows the tests way down.
- Programmatically filtering the elements to only include the ones involved in the test.
  - This is promising but adds a fair amount of complexity and may still require scrolling when many form elements are involved.
- Continuing to use `ScrollView` + `VStack` + `ForEach` when running UI tests only.
  - This is the simplest approach and is the one this PR takes. It requires zero changes to the tests and the test view. This essentially maintains the status quo for `FeatureFormView` UI tests.


rolson (commented):
I'm not super comfortable with this approach because we aren't testing the code we are shipping. But it is simple enough, and it still allows us to test some of our code. @philium do you have any thoughts on this approach?

dfeinzimer:
My view is that this keeps us close enough to testing what we’re shipping within a reasonable margin. `List` provides some nice visual formatting and performance optimizations, but it and `VStack`/`ForEach` both serve as fundamental high level containers and if either of them changed much at this point there’d be bigger problems to handle. 

If the tests were doing image comparison, this would be a different story, but these tests primarily verify the presence of elements and that interactions with them work as expected.

> Lastly, did you try swiping instead of scrolling?

Yes, I prototyped that approach on [df/FFV_As_List_Scroll_To_Elements](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/compare/df/FFV_As_List...df/FFV_As_List_Scroll_To_Elements) as part of the prep work for evaluating different approaches and ultimately making this my preference. Here are the two big issues I see with swiping:
1. Under this approach, the set of tests took 5.2 minutes to complete. Under the scrolling approach, completion time is 12 minutes. That’s nearly 7 minutes of test runner time spent on scrolling.
2. There’s quite a big difference in behavior in `swipeUp()` between Mac Catalyst and iOS. On Mac Catalyst, a call to `swipeUp()` scrolls a simple list of numbers all the way to row 83. On iOS, the same call gets us to row 25. As you can imagine, this creates issues with over-scroll when writing tests to run on both platforms. Try it for yourself with this [sample project](https://github.com/user-attachments/files/24112157/Tester.zip) I made. (I reported this to Apple back in August under FB19836397.)

Overall, it’s unfortunate that this dilemma exists. Scrolling wasn’t necessary with `UITableView` (I verified it with [this project](https://github.com/user-attachments/files/24112317/UITesting.zip)). [This radar](https://openradar.appspot.com/FB11764272) actually hits on the issue quite well. Apparently scrolling in lists wasn’t necessary back in early versions of SwiftUI, but that changed in Xcode 14. An Apple Engineer even directly said at the time “You are not supposed to have to scroll manually.”

dfeinzimer:
After more discussion, we will proceed with this testing approach under the assumption that it is a temporary problem. It will allow us to begin using `List` in `FeatureFormView` and limit complicating our current test implementations.

PR #1344, opened 2025-12-10
UI Test Fixes

dfeinzimer:
- Patches a new failure in `AttachmentCameraControllerTests.testOnCameraCaptureModeChanged()` since the test runner upgrade to iOS 26.
- Removes a file entry that was missed in #1275.

### DIFF
--- a/Test Runner/UI Tests/AttachmentCameraControllerTests.swift
+++ b/Test Runner/UI Tests/AttachmentCameraControllerTests.swift
@@ -22,13 +22,9 @@ final class AttachmentCameraControllerTests: XCTestCase {
     
     /// Test `AttachmentCameraController.onCameraCaptureModeChanged(perform:)`
     func testOnCameraCaptureModeChanged() throws {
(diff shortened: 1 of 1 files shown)


yo1995 (dismissed):
Test fails for me on iOS 18 device, on line 54. Does this test need to be run on iOS 26?

yo1995:
In that case, should the project minimum target be bumped to iOS 26 as well?

dfeinzimer:
> In that case, should the project minimum target be bumped to iOS 26 as well?

As far I understand, we keep that tied to the minimum version supported by the overall toolkit, so I wouldn't think so.

I don't personally have an iOS 18 device any longer to test this with, and this specific test requires a hardware device. Is the test failing for you at the step where the VIDEO button is pressed? If so, the problem is with how programmatic interaction with that specific element (not toolkit UI) was more brittle and challenging in iOS 18.

If you think we should keep this specific test passing on iOS 18, I can revert the simplifications made to it in #1324.

PR #1345, opened 2025-12-10
Support BGContinuedProcessingTask in OfflineMapAreasView

rolson:
- Adds support for BGContinuedProcessingTask to the Offline Map toolkit component.
- Add a property `useBGContinuedProcessingTasks` that allows the user to specify if they want to use `BGContinuedProcessingTask` or the `JobManager` for job management.
- Adds a toolkit extension method on `BGContinuedProcessingTask` to easily bind it to an ArcGIS `Job`.
- Update tutorial and example to absorb the changes
- Add a scene modifier to take a configuration closure so that the offline manager can be configured in one place within a user's application.

Note: this doesn't work on the simulator. This must be tested on device.

issue swift/7083

The offline manager will either use the job manager or BGContinuedProcessingTask, it will not try to use both. I don't think a workflow using both makes sense.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
@@ -87,11 +87,15 @@ public class OfflineManager: ObservableObject {
     
     /// A Boolean value indicating if the `JobManager` is managing the jobs.
     fileprivate var isUsingJobManager: Bool {
+#if !os(visionOS) && !targetEnvironment(macCatalyst)
         if #available(iOS 26.0, *) {
             !configuration.useBGContinuedProcessingTasks
         } else {
             true
         }
+#else
+        true
+#endif
     }
     
     /// The configuration settings for the offline manager.
@@ -154,7 +158,7 @@ public class OfflineManager: ObservableObject {
         Task.detached {
             await self.savePendingMapInfo(for: portalItem)
         }
-#if !targetEnvironment(visionOS) && !targetEnvironment(macCatalyst)
+#if !os(visionOS) && !targetEnvironment(macCatalyst)
         if #available(iOS 26.0, *) {
             if configuration.useBGContinuedProcessingTasks {
                 try? startContinuedProcessingTask(for: job, title: title)
@@ -414,14 +418,14 @@ public struct OfflineManagerConfiguration: Sendable {
     /// The update mode of any new preplanned map areas taken offline.
     public var preplannedUpdateMode: DownloadPreplannedOfflineMapParameters.UpdateMode
     
+#if !os(visionOS) && !targetEnvironment(macCatalyst)
     /// Backing variable for `useBGContinuedProcessingTasks`.
     private var _useBGContinuedProcessingTasks = true
     
     /// A Boolean value indicating whether the `OfflineManager` makes use of
     /// `BGContinuedProcessingTask` in-lieu of the `JobManager`.
     /// Changing the value of this property between app runs is not supported.
-    /// Setting this has no effect on platforms other than iOS.
-    /// The default value of this property is `true` on iOS.
+    /// The default value of this property is `true`.
     @available(iOS 26.0, *)
     public var useBGContinuedProcessingTasks: Bool {
         get {
@@ -431,6 +435,7 @@ public struct OfflineManagerConfiguration: Sendable {
             _useBGContinuedProcessingTasks = newValue
         }
     }
+#endif
     
     /// Creates a configuration for the offline manager.
     /// - Parameters:
@@ -448,6 +453,7 @@ public struct OfflineManagerConfiguration: Sendable {
         self.preplannedUpdateMode = preplannedUpdateMode
     }
     
+#if !os(visionOS) && !targetEnvironment(macCatalyst)
     /// Creates a configuration for the offline manager.
     /// - Parameters:
     ///   - useBGContinuedProcessingTasks: Whether the `OfflineManager` makes use of
@@ -468,6 +474,7 @@ public struct OfflineManagerConfiguration: Sendable {
         self.preplannedUpdateMode = preplannedUpdateMode
         _useBGContinuedProcessingTasks = useBGContinuedProcessingTasks
     }
+#endif
 }
 
 extension Logger {
@@ -495,7 +502,7 @@ private extension Dictionary {
     }
 }
 
-#if !targetEnvironment(visionOS) && !targetEnvironment(macCatalyst)
+#if !os(visionOS) && !targetEnvironment(macCatalyst)
 extension OfflineManager {
     /// Starts a background continued processing task for a given job that is
     /// taking a map offline.
--- a/Sources/ArcGISToolkit/Extensions/BackgroundTasks/BGContinuedProcessingTask+Job.swift
+++ b/Sources/ArcGISToolkit/Extensions/BackgroundTasks/BGContinuedProcessingTask+Job.swift
@@ -15,7 +15,7 @@
 import ArcGIS
 import BackgroundTasks
 
-#if !targetEnvironment(visionOS) && !targetEnvironment(macCatalyst)
+#if !os(visionOS) && !targetEnvironment(macCatalyst)
 @available(iOS 26.0, *)
 public extension BGContinuedProcessingTask {
     /// Binds a job to this continued processing task such that the job progress


des12437:
> I wonder if we should basically disable the JobManager for iOS 26 within the OfflineMapAreasView?

Yes the Job Manager should be disabled for iOS 26 for the OfflineMapAreasView because when the app is killed during an active download the `BGContinuedProcessingTask` fails and when the app is re-opened the job manager job resumes and the download succeeds which is unexpected.

Sources/ArcGISToolkit/Extensions/BackgroundTasks/BGContinuedProcessingTask+Job.swift
```diff
@@ -0,0 +1,54 @@
...
+        taskProgress.completedUnitCount = Int64(job.progress.fractionCompleted * 100)
+        
+        // Observe progress on the job and update the task.
+        Task {
+            var observers = Set<NSKeyValueObservation>()
+            let observer = job.progress.observe(\.fractionCompleted, options: .new) { _, value in
+                guard let fractionCompleted = value.newValue else { return }
+                taskProgress.completedUnitCount = Int64(fractionCompleted * 100)
+            }
+            observers.insert(observer)
```
yo1995: It seems the `observers` set isn't referenced elsewhere. Can it be removed and just use `let observer` local variable?

Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
```diff
@@ -360,3 +365,31 @@ private extension Dictionary {
...
+            identifier: cptIdentifier,
+            title: String(
+                localized: "Offline Map",
+                bundle: .toolkitModule,
+                comment: "The default title of the live activity for taking a map offline."
+            ),
+            subtitle: title
+        )
+        request.strategy = .fail
+        try? BGTaskScheduler.shared.submit(request)
```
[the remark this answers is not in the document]
rolson: It could throw, but then the caller would have to handle it or use `try?`. I'm not sure how the caller would handle it any better.

Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
```diff
@@ -45,9 +47,10 @@ import SwiftUI
 /// **Behavior**
 ///
 /// The offline manager is not instantiable, you must use the ``shared`` instance.
-/// Set the `offlineManager(preferredBackgroundStatusCheckSchedule:jobCompletion:)`
-/// modifier at the entry point of your application to add additional setup
-/// required for the component to use the job manager. For example:
+/// Setup and configure the offline manager with the `offlineManager(configure:)`
```
philium:
```suggestion
/// Set up and configure the offline manager with the `SwiftUI.Scene.offlineManager(configure:)`
```

Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
```diff
@@ -73,13 +78,41 @@ public class OfflineManager: ObservableObject {
...
+    /// - Since 300.0 
+    public var onJobCompletion: ((any JobProtocol) -> Void)?
     
-    /// The job manager used by the offline manager.
-    let jobManager = JobManager(uniqueID: "offlineManager")
+    /// Backing variable for the `jobManager` property.
+    let jobManager: JobManager = JobManager(uniqueID: "offlineManager")
+    
+    /// A Boolean value indicating if the `JobManager` is managing the jobs.
+    var isUsingJobManager: Bool {
```
philium: This could be file private.

```suggestion
    fileprivate var isUsingJobManager: Bool {
```

Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
```diff
@@ -102,30 +162,45 @@ public class OfflineManager: ObservableObject {
...
+        // If using the job manager, then perform required setup.
+        if isUsingJobManager {
+            // Observe each job's status.
+            for job in jobManager.jobs {
+                observeJob(job)
+            }
+            
+            // Resume all paused jobs.
+            let count = jobManager.jobs.filter { $0.status == .paused }
+                .count
```
philium:
```suggestion
            let count = jobManager.jobs.count(where: { $0.status == .paused })
```

Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
```diff
@@ -136,16 +211,22 @@ public class OfflineManager: ObservableObject {
...
-            // as possible after removing a job instead of waiting for the app to be backgrounded.
-            jobManager.saveState()
+            if isUsingJobManager {
+                // Remove completed job from JobManager.
+                Logger.offlineManager.debug("Removing completed job from job manager")
+                jobManager.jobs.removeAll { $0 === job }
+                
+                // This isn't strictly required, but it helps to get the state saved as soon
+                // as possible after removing a job instead of waiting for the app to be backgrounded.
+                jobManager.saveState()
```
philium: Perhaps `JobManager.jobs` should be private and it should have methods for modifying its jobs? For example, it could have a `removeJobs` method that saves the state after the jobs are removed.

Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
```diff
@@ -73,13 +78,41 @@ public class OfflineManager: ObservableObject {
...
+        } else {
+            true
+        }
+    }
+    
+    /// The preferred schedule for performing status checks while the application is in the
+    /// background.
+    /// - SeeAlso ``JobManager/preferredBackgroundStatusCheckSchedule``
+    /// - Since 300.0
+    public var preferredBackgroundStatusCheckSchedule: BackgroundStatusCheckSchedule {
```
philium: Could this be a parameter of the `offlineManager` modifiers instead?

Sources/ArcGISToolkit/Extensions/BackgroundTasks/BGContinuedProcessingTask+Job.swift
```diff
@@ -0,0 +1,56 @@
...
+            Task.detached { await job.cancel() }
+        }
+        
+        // Set initial progress on the task.
+        let taskProgress = progress
+        taskProgress.totalUnitCount = 100
+        taskProgress.completedUnitCount = Int64(job.progress.fractionCompleted * 100)
+        
+        // Observe progress on the job and update the task.
+        Task {
```
philium: Should this task be detached as well?

Sources/ArcGISToolkit/Extensions/BackgroundTasks/BGContinuedProcessingTask+Job.swift
```diff
@@ -0,0 +1,56 @@
...
+    func bind(to job: some JobProtocol) {
+        // Cancel job if task expires or is cancelled.
+        expirationHandler = {
+            Task.detached { await job.cancel() }
+        }
+        
+        // Set initial progress on the task.
+        let taskProgress = progress
+        taskProgress.totalUnitCount = 100
+        taskProgress.completedUnitCount = Int64(job.progress.fractionCompleted * 100)
```
philium: Instead of this, you could `initial` as one of the observing options.

Sources/ArcGISToolkit/Extensions/BackgroundTasks/BGContinuedProcessingTask+Job.swift
philium: There's no guarantee that `observers` will stick around. That may work not, but that behavior could change. Instead, just call `invalidate()` on the observer when done:

```swift
let observer = job.progress.observe(\.fractionCompleted, options: [.initial, .new]) { progress, _ in
    taskProgress.completedUnitCount = Int64(progress.fractionCompleted * 100)
}
defer { observer.invalidate() }
```

Sources/ArcGISToolkit/Extensions/BackgroundTasks/BGContinuedProcessingTask+Job.swift
```diff
@@ -0,0 +1,56 @@
...
+            }
+            observers.insert(observer)
+            
+            let result = await job.result
+            switch result {
+            case .success:
+                setTaskCompleted(success: true)
+            case .failure:
+                setTaskCompleted(success: false)
+            }
```
philium: Better to call `setTaskCompleted` in only one place.

```suggestion
            let success = switch result {
            case .success: true
            case .failure: false
            }
            setTaskCompleted(success: success)
```

OfflineMapAreasExample/OfflineMapAreasExample/OfflineMapAreasExampleApp.swift
```diff
@@ -22,17 +22,26 @@ struct OfflineMapAreasExampleApp: App {
...
-        // Use of this scene modifier is required for the offline component to complete map area download jobs when
-        // the app is backgrounded.
-        //
-        // Set the `preferredBackgroundStatusCheckSchedule` to `.regularInterval(interval: 30)` to check the status
-        // of the download job in the background every 30 seconds. Use the `jobCompletionAction` closure to send
-        // a notification once a download job completes.
-        .offlineManager(preferredBackgroundStatusCheckSchedule: .regularInterval(interval: 30)) { job in
-            // Post a local notification that the job is finished.
-            Self.notifyJobCompleted(job: job)
+        // Apply the `.offlineManager(configure:)` scene modifier
```
philium: Leading dot isn't necessary.

```suggestion
        // Apply the `offlineManager(configure:)` scene modifier
```

philium (commented):
I was surprised to see that `JobManager.jobs` is publicly settable. Should it not be?

Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
```diff
@@ -89,6 +122,33 @@ public class OfflineManager: ObservableObject {
...
+    /// A Boolean value indicating whether the `OfflineManager` makes use of
+    /// `BGContinuedProcessingTask` in-lieu of the `JobManager`.
+    /// This property must be set before taking maps offline and should be set
+    /// to the same value the next time the app is run. Changing the value
+    /// of this property between app runs is not supported.
+    /// Setting this has no effect on platforms other than iOS.
+    /// The default value of this property is `true` on iOS.
+    /// - Since 300.0
+    @available(iOS 26.0, *)
+    public var useBGContinuedProcessingTasks: Bool {
```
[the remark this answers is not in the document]
rolson: See answer above

Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
```diff
@@ -73,13 +78,41 @@ public class OfflineManager: ObservableObject {
     public static let shared = OfflineManager()
```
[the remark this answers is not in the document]
rolson: If you think that should be rectified (removing the shared instance) please create an issue.

rolson:
Changed from closure to async sequence.

Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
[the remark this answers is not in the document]
philium: The reason I suggest it now is that now would be an ideal time to make some of those changes. One of the modifiers is new to 300.0 and with my recommendation of moving `useBGContinuedProcessingTasks` to a modifier parameter, that would be a new modifier as well (and in fact the new one, with the `configure` parameter, wouldn't even be necessary at all). The new behavior could be introduced with the new modifier, making it opt in (if you use the new modifier, you get the new behavior).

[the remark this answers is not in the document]
rolson: There are so many properties to configure and more could be added later, that's why I went the configuration closure route. Do you mean the user would create it and set it scene with the modifier?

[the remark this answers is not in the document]
philium: That's not a bad option. Instead of the closure taking an offline manager, the closure could return an offline manager that the user created. I still recommend taking one-time set values as modifier parameters instead of making them readwrite properties (except for the ones where that ship has already sailed, if any).

rolson:
> I was surprised to see that JobManager.jobs is publicly settable. Should it not be?

Yes, that's how the JobManager works, you add/remove jobs to it and it just does basic serialization/deserialization:

```swift
/// The job manager is an `ObservableObject` with a mutable ``jobs`` property. Adding a job to this
/// property will allow the job manager to do the work to make sure that we can rehydrate a job
/// if an app is terminated.
```

rolson:
Ok. I've come up with a solution where it leads the user down the road of setting the properties on initialization and not able to change them later. I like this solution. I think we can still make the OfflineManager not shared in a separate PR, I think that would be still doable if we wanted to go down that route and have the OfflineManager that was created be set on the environment. However, I think that change would be pretty invasive and deserves its own PR. This work doesn't preclude doing that down the road.

OfflineMapAreasExample/OfflineMapAreasExample/OfflineMapAreasExampleApp.swift
```diff
@@ -22,16 +22,19 @@ struct OfflineMapAreasExampleApp: App {
...
+        // at the entry point of the application to setup background download
+        // support for the offline component. Use of this scene modifier is
+        // required for the offline component to complete map area download jobs
+        // when the app is backgrounded. It also gives you a chance to configure
+        // properties of the offline manager.
+        .offlineManager(
+            configuration:
+                OfflineManagerConfiguration(
+                    preferredBackgroundStatusCheckSchedule: .regularInterval(interval: 30)
+                )
```
philium: Not a big deal, but this is typically how we would format this.

```suggestion
            configuration: OfflineManagerConfiguration(
                preferredBackgroundStatusCheckSchedule: .regularInterval(interval: 30)
            )
```

Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
```diff
@@ -57,9 +60,12 @@ import SwiftUI
 ///             ContentView()
 ///         }
 ///         // Setup the offline toolkit components for the app.
-///         .offlineManager(preferredBackgroundStatusCheckSchedule: .regularInterval(interval: 30)) { job in
-///             // Do something after the job completes…
-///         }
+///        .offlineManager(configuration: .init(
+///            preferredBackgroundStatusCheckSchedule: .regularInterval(interval: 30)
+///        )) { job in
```
philium: This is how we would typically format this.

```suggestion
///        .offlineManager(
///            configuration: .init(
///                preferredBackgroundStatusCheckSchedule: .regularInterval(interval: 30)
///            )
///        ) { job in
```

Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
```diff
@@ -99,33 +168,52 @@ public class OfflineManager: ObservableObject {
...
-        jobManager.resumeAllPausedJobs()
     }
     
     /// Starts a job that will be managed by this instance.
     /// - Parameters:
     ///   - job: The job to start.
     ///   - portalItem: The portal item whose map is being taken offline.
-    func start(job: some JobProtocol, portalItem: PortalItem) {
+    ///   - title: The title of the map area being taken offline.
+    func startJob(_ job: some JobProtocol, portalItem: PortalItem, title: String) {
```
philium: Should this method now be throwing since the call to `startContinuedProcessingTask` could throw? Probably better to let the caller decide whether to discard the error rather than making that decision for them.

Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
```diff
@@ -302,25 +397,53 @@ public class OfflineManager: ObservableObject {
...
         offlineMapInfos.removeAll { $0.id == offlineMapInfo.id }
     }
+    
+    fileprivate func apply(configuration: OfflineManagerConfiguration) {
+        preferredBackgroundStatusCheckSchedule = configuration.preferredBackgroundStatusCheckSchedule
+        onDemandUpdateMode = configuration.onDemandUpdateMode
+        preplannedUpdateMode = configuration.preplannedUpdateMode
+        if #available(iOS 26.0, *) {
+            useBGContinuedProcessingTasks = configuration.useBGContinuedProcessingTasks
+        }
```
philium: Could each of the individual properties be replaced with a `configuration` property? Then updating the configuration would be as simple as setting a single property.

Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
```diff
@@ -329,13 +389,87 @@ public extension SwiftUI.Scene {
...
         }
     }
 }
 
+/// The configuration properties for the ``OfflineManager``.
+/// - Since 300.0
+public struct OfflineManagerConfiguration: Sendable {
+    /// The preferred schedule for performing status checks while the application is in the
+    /// background.
+    /// - SeeAlso ``JobManager/preferredBackgroundStatusCheckSchedule``
```
yo1995:
```suggestion
    /// - SeeAlso: ``JobManager/preferredBackgroundStatusCheckSchedule``
```

Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
```diff
@@ -329,13 +389,87 @@ public extension SwiftUI.Scene {
...
+    public var preferredBackgroundStatusCheckSchedule: BackgroundStatusCheckSchedule
+    
+    /// The update mode of any new on-demand map areas taken offline.
+    public var onDemandUpdateMode: GenerateOfflineMapParameters.UpdateMode
+    
+    /// The update mode of any new preplanned map areas taken offline.
+    public var preplannedUpdateMode: DownloadPreplannedOfflineMapParameters.UpdateMode
+    
+    /// Backing variable for `useBGContinuedProcessingTasks`.
+    private var _useBGContinuedProcessingTasks: Bool = true
```
yo1995:
```suggestion
    private var _useBGContinuedProcessingTasks = true
```

yo1995 (commented):
Tested on iOS 26 device and the BGCPT is showing up correctly. 

https://github.com/user-attachments/assets/632887ba-383d-4048-9d79-b2513dc1481d

PR #1349, opened 2026-01-07
`FeatureFormView` - Convert `ComboBoxInput` and `RadioButtonsInput` to `Picker`s

dfeinzimer:
Apollo 1556

As of #1283, `FeatureFormView` uses a SwiftUI `Form`, where it previously used a `VStack`. In a `VStack`, the [`.inline`](https://developer.apple.com/documentation/swiftui/inlinepickerstyle) picker style takes on the appearance of a scroll wheel, which is not what we wanted. In a `Form`, the `.inline` picker style matches the appearance of the current `ComboBoxInput` and `RadioButtonsInput` interface. Therefore, we can now use `Picker` paired with the `.inline` picker style in these two inputs.

| `Picker` in a `VStack` | `Picker` in a `Form` |
|---|---|
|  |  |

## iOS
| Before | After |
|---|---|
|  |  |

## Mac Catalyst
| Before | After |
|---|---|
|  |  |

## visionOS
| Before | After |
|---|---|
|  |  |

### DIFF
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/ComboBoxInput.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/ComboBoxInput.swift
@@ -271,7 +271,7 @@ extension ArcGIS.CodedValue: Swift.Hashable {
     }
 }
 
-private enum ComboBoxValue: Equatable {
+private enum ComboBoxValue: Hashable {
     case coded(CodedValue)
     /// The element's current (but unsupported) value.
     ///
@@ -297,5 +297,3 @@ private enum ComboBoxValue: Equatable {
         }
     }
 }
-
-extension ComboBoxValue: Hashable {}


Test Runner/UI Tests/FeatureFormViewTests.swift
```diff
@@ -2362,3 +2363,10 @@ private extension XCUIApplication {
         buttons["No value Combo Box Option"]
     }
 }
+
+private extension XCUIElement {
+    /// The element's value as a string.
+    var stringValue: String? {
+        self.value as? String
```
CalebRas: Not sure if this a preference thing or not, so feel free to ignore if you would like.

```suggestion
        value as? String
```

CalebRas (commented):
Looks good and tests pass for me. Just some linting suggestions:

Test Runner/UI Tests/FeatureFormViewTests.swift
dfeinzimer: Yep thanks, this was leftover from something I was trying. It isn't needed.

mhdostal (approved):

PR #1353, opened 2026-01-09
Increase timeout on flaky test

mhdostal:
FeatureFormViewTest 13.1 has failed several times in the last few weeks, all with the error `The network source group "Electric Distribution Device" doesn't exist`. This increases the timeout to help fix the flaky test.


dfeinzimer (commented):
I've had my eye on this issue for a while. The three 13.x cases are all impacted. The navigation stack intermittently fails to respond to the last 4 items being removed after association is added here:
https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/1f5daba72fff524e9b87ee49674a1a8233aacdab/Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FormElements/UtilityAssociationsFormElement/UtilityAssociationCreationView.swift#L127
The items are removed but the presented view never changes, so even if we bump the timeout, the expected view won't ever be found.

Occasionally, a purple runtime warning is emitted too, I'm not sure if directly related but I suspect it is:
> Adding '_UIReparentingView' as a subview of UIHostingController.view is not supported and may result in a broken view hierarchy. Add your view above UIHostingController.view in a common superview or insert it into your SwiftUI content in a UIViewRepresentable instead.

I don't think this was an issue in 16.x and cropped up once we moved to 26.x.

CalebRas (approved):

PR #1363, opened 2026-01-22
Floor Filter: iOS 26 look and feel

rolson:
- Uses standard components for the Sites and Facilities picker (NavigationStack, Searchable, proper button roles)
- Adds Liquid Glass look and feel for iOS 26
- Changes look and feel to be similar for iOS 17/18, but uses material instead of liquid glass

iPhone

iPad

image


Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
```diff
@@ -282,3 +189,227 @@ public extension FloorFilter {
...
+                scrollToSelectedLevel(with: scrollView)
+            }
+            .onChange(of: model.selection) { _, _ in
+                guard model.selection?.facility != nil, !isCollapsed else { return }
+                scrollToSelectedLevel(with: scrollView)
+            }
+            .onAppear {
+                guard model.selection?.facility != nil, !isCollapsed else { return }
+                scrollToSelectedLevel(with: scrollView)
+            }
```
CalebRas: These can probably be combined, correct?

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
```diff
@@ -282,3 +189,227 @@ public extension FloorFilter {
...
+    }
+    
+    /// Scrolls the list within the provided proxy to the button representing the selected level.
+    /// - Parameter scrollView: The scroll view proxy containing the scroll view.
+    func scrollToSelectedLevel(with scrollView: ScrollViewProxy) {
+        if let level = model.selection?.level {
+            withAnimation {
+                scrollView.scrollTo(level.id)
+            }
+        }
```
CalebRas: Nitpick, but a guard seems more appropriate here.

Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
```diff
@@ -15,244 +15,195 @@
...
-            headerLowerHalf
+    @State private var searchText = ""
+    
+    /// The filtered sites.
+    var sites: [FloorSite] {
+        guard !searchText.isEmpty else {
+            return model.sites
+        }
+        return model.sites.filter {
+            $0.name.localizedStandardContains(searchText)
```
CalebRas: Should use if instead

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
```diff
@@ -282,3 +189,227 @@ public extension FloorFilter {
...
+                            }
+                        }
+                    }
+                    .modify {
+                        if #available(iOS 18.0, *) {
+                            $0.onScrollGeometryChange(for: CGFloat.self) { geometry in
+                                geometry.contentSize.height
+                            } action: { _, newValue in
+                                contentHeight = newValue
+                            }
```
CalebRas: TODO: Can use key paths?

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
```diff
@@ -282,3 +189,227 @@ public extension FloorFilter {
...
+    
+    @EnvironmentObject private var model: FloorFilterViewModel
+    @State private var isCollapsed = false
+    @State private var contentHeight: CGFloat = .zero
+    
+    var body: some View {
+        VStack(spacing: 0) {
+            if isTopAligned {
+                if !model.sortedLevels.isEmpty {
+                    levelButtons
```
CalebRas: TODO: combine

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
```diff
@@ -282,3 +189,227 @@ public extension FloorFilter {
...
+
+/// The combined UI components for the floor filter.
+private struct FloorFilterBody: View {
+    @EnvironmentObject private var model: FloorFilterViewModel
+    @State private var isSiteSelectorPresented = false
+    
+    let width: CGFloat
+    let isTopAligned: Bool
+    
+    static var buttonShape = RoundedRectangle(cornerRadius: 18, style: .continuous)
```
CalebRas: TODO: let?

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
```diff
@@ -282,3 +189,227 @@ public extension FloorFilter {
...
+    var body: some View {
+        Button {
+            if isCollapsed && model.sortedLevels.count > 1 {
+                isCollapsed = false
+            } else {
+                model.setLevel(level)
+            }
+        }
+        label: {
+            Text(level.shortName)
```
CalebRas: TODO: same line.

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
CalebRas: Should this be `let` instead of `var`? Also, I believe `continuous` is the default argument for `style`.

```suggestion
    static let buttonShape = RoundedRectangle(cornerRadius: 18)
```

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
```diff
@@ -180,79 +124,42 @@ public struct FloorFilter: View {
...
-        }
-        .onChange(of: viewpoint.wrappedValue) {
-            guard isNavigating.wrappedValue else { return }
-            if let newViewpoint = viewpoint.wrappedValue {
-                viewModel.onViewpointChanged(newViewpoint)
+            .onChange(of: viewpoint.wrappedValue) {
+                guard isNavigating.wrappedValue else { return }
+                if let newViewpoint = viewpoint.wrappedValue {
+                    viewModel.onViewpointChanged(newViewpoint)
+                }
```
CalebRas: I know you just copied these lines, but the 2 conditional statements could probably be combined.

```suggestion
                guard isNavigating.wrappedValue, let newViewpoint = viewpoint.wrappedValue else {
                    return
                }
                viewModel.onViewpointChanged(newViewpoint)
```

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
```diff
@@ -282,3 +189,227 @@ public extension FloorFilter {
...
+                        .contentShape(FloorFilterBody.buttonShape)
+                }
+                .buttonStyle(.plain)
+            }
+            
+            if !isTopAligned {
+                if !model.sortedLevels.isEmpty {
+                    levelButtons
+                }
+            }
```
CalebRas: Consider abstracting the button here into a property, so that you can combine some of the conditional logic, similar to what you did with `FloorFilterBody.divider`. I am thinking of something like:

```swift
VStack(spacing: 0) {
    if !model.sortedLevels.isEmpty {
        if isTopAligned {
            levelButtons
            collapseButton
        } else {
            collapseButton
            levelButtons
        }
    }
}
```

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
CalebRas: These 2 modifiers seem to have the same logic, so they can probably be combined using the `onChange`'s `initial` parameter. Also, the unused closure parameters can be removed.

```suggestion
            .onChange(of: model.selection, initial: true) {
                guard model.selection?.facility != nil, !isCollapsed else { return }
                scrollToSelectedLevel(with: scrollView)
            }
```

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
CalebRas: Nitpick, but using a guard seems more appropriate here.

```suggestion
        guard let level = model.selection?.level else { return }
        withAnimation {
            scrollView.scrollTo(level.id)
        }
```

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
```diff
@@ -282,3 +189,227 @@ public extension FloorFilter {
...
+    
+    var body: some View {
+        Button {
+            if isCollapsed && model.sortedLevels.count > 1 {
+                isCollapsed = false
+            } else {
+                model.setLevel(level)
+            }
+        }
+        label: {
```
CalebRas: Should these be on the same line?

```suggestion
        } label: {
```

Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
CalebRas: I would recommend against returning a value (that's not nil) in the guard's else block if possible. Consider using:

```swift
searchText.isEmpty
? model.sites
: model.sites.filter { $0.name.localizedStandardContains(searchText) }
```

Or:

```swift
if searchText.isEmpty {
    model.sites
} else {
    model.sites.filter { $0.name.localizedStandardContains(searchText) }
}
```

Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
```diff
@@ -15,244 +15,195 @@
...
+        let facilities = site?.facilities ?? model.facilities
+        
+        guard !searchText.isEmpty else {
+            return facilities.sorted(using: KeyPathComparator(\.name))
         }
         return facilities
-            .filter { $0.name.localizedStandardContains(query) }
-            .sorted { $0.name < $1.name  }
+            .filter { $0.name.localizedStandardContains(searchText) }
+            .sorted(using: KeyPathComparator(\.name))
```
CalebRas: Same here about avoiding a return value in the guard's else block. Consider using:

```swift
return if searchText.isEmpty {
    facilities.sorted(using: KeyPathComparator(\.name))
} else {
    facilities
        .filter { $0.name.localizedStandardContains(searchText) }
        .sorted(using: KeyPathComparator(\.name))
}
```

(Also, I didn't know `KeyPathComparator` existed. That's cool!)

Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
```diff
@@ -15,244 +15,195 @@
...
-    var matchingSites: [FloorSite] {
-        guard !query.isEmpty else {
-            return viewModel.sites
+    var body: some View {
+        VStack {
+            if facilities.isEmpty {
+                ContentUnavailableView(String.noMatchesFound, systemImage: "building.2")
+            } else {
+                ScrollViewReader { scrollView in
+                    List(facilities, id:\.id) { facility in
```
CalebRas: `FloorFacility` is `Identifiable` so no need to specify the `id` parameter.

```suggestion
                    List(facilities) { facility in
```

Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
```diff
@@ -15,244 +15,195 @@
...
-    @State private var query = ""
+    @State private var navigationPath = NavigationPath()
     
-    /// A Boolean value indicating whether the user pressed the back button in the header.
-    ///
-    /// This allows for browsing the site list while keeping the current selection unmodified.
-    @State private var userDidBackOutToSiteList = false
+    /// A Boolean value indicating if there are multiple sites. If there not
+    /// multiple sites, then we go straight to the facilities list.
+    private var hasMultipleSites: Bool { model.sites.count > 1 }
```
CalebRas: Nitpick, but it looks like this property is only use one place, so you might want to consider removing it. Up to you, though.

Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
```diff
@@ -15,244 +15,195 @@
...
-                    }
-                if textFieldIsFocused && !query.isEmpty {
-                    XButton(.clear) {
-                        query.removeAll()
+    var body: some View {
+        VStack {
+            if sites.isEmpty {
+                ContentUnavailableView(String.noMatchesFound, systemImage: "building.2")
+            } else {
+                List(sites, id:\.id) { site in
```
CalebRas:
```suggestion
                List(sites) { site in
```

`FloorSite` is also `Identifiable`, so the `id` parameter is unneeded.

Examples/Examples/FloorFilterExampleView.swift
dfeinzimer, line 1: We should apply this change to the tutorial too. (`FloorFilterStep5.swift`)

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
```diff
@@ -282,3 +189,227 @@ public extension FloorFilter {
...
+                LevelSelector(isTopAligned: isTopAligned)
+                divider
+                siteSelectorButton
+            }
+        }
+        .frame(width: width)
+        .modify {
+            if #available(iOS 26, *) {
+#if os(visionOS)
+                $0.glassBackgroundEffect(in: FloorFilter26.buttonShape)
```
dfeinzimer: `FloorFilterBody`?

```suggestion
                $0.glassBackgroundEffect(in: FloorFilterBody.buttonShape)
```

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
```diff
@@ -282,3 +189,227 @@ public extension FloorFilter {
...
     }
 }
+
+/// The combined UI components for the floor filter.
+private struct FloorFilterBody: View {
+    @EnvironmentObject private var model: FloorFilterViewModel
+    @State private var isSiteSelectorPresented = false
+    
+    let width: CGFloat
+    let isTopAligned: Bool
```
dfeinzimer:
```suggestion
    let isTopAligned: Bool
    let width: CGFloat
```

⚠️ You'll also need to update where the `FloorFilterBody` initializer is used.

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
```diff
@@ -282,3 +189,227 @@ public extension FloorFilter {
...
+    
+    static var buttonShape = RoundedRectangle(cornerRadius: 18, style: .continuous)
+    static let buttonSize: CGFloat = 56
+    static let padding: CGFloat = 3
+    static let fontSize: CGFloat = 20
```
dfeinzimer:
```suggestion
    static let fontSize: CGFloat = 20
    static let padding: CGFloat = 3
```

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
```diff
@@ -113,65 +110,12 @@ public struct FloorFilter: View {
...
-        } else if viewModel.loadStatus == .loaded {
-            Button {
-                siteAndFacilitySelectorIsPresented.toggle()
-            } label: {
-                Image(systemName: "building.2")
-                    .padding(.toolkitDefault)
-                    .contentShape(.rect(cornerRadius: 5))
-                    .hoverEffect()
-            }
-            .accessibilityIdentifier("Floor Filter button")
```
dfeinzimer, line 138: This removal breaks the first assertion in `FloorFilterTests.testFloorFilter()`.

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilterViewModel.swift
```diff
@@ -54,6 +54,9 @@ final class FloorFilterViewModel: ObservableObject {
     /// The selected site, floor, or level.
     @Published private(set) var selection: FloorFilterSelection?
     
+    /// Whether the selector is showing facilities from all sites.
+    @Published var showingFaciltiesFromAllSites = false
```
dfeinzimer:
```suggestion
    @Published var showingFacilitiesFromAllSites = false
```

Sources/ArcGISToolkit/Components/FloorFilter/LevelSelector.swift
dfeinzimer, line 1: Any downside to keeping this file around? It'll help keep FloorFilter.swift focused and balanced with `SiteAndFacilitySelector` broken out into its own file.

Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
```diff
@@ -15,244 +15,195 @@
 import ArcGIS
 import SwiftUI
 
-/// A view which allows selection of sites and facilities represented in a `FloorManager`.
-///
-/// If the floor aware data contains only one site, the selector opens directly to the facilities list.
```
dfeinzimer, line 20: Is this deletion intentional?

Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
```diff
@@ -15,244 +15,195 @@
...
+            navigationPath.append(facility)
         }
+        .frame(minWidth: 360, minHeight: 500)
     }
+}
+
+private struct SiteList: View {
+    /// Allows the user to toggle the visibility of the site and facility selector.
+    @Binding var isPresented: Bool
+
```
dfeinzimer:
```suggestion
    
```

Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
```diff
@@ -15,244 +15,195 @@
...
-extension SiteAndFacilitySelector {
-    /// A Boolean value indicating whether the back button in the header navigations controls is visible..
-    var backButtonIsVisible: Bool {
-        facilityListIsVisible
-        && multipleSitesAreAvailable
-    }
+private struct FacilityList: View {
+    /// Allows the user to toggle the visibility of the site and facility selector.
+    @Binding var isPresented: Bool
+
```
dfeinzimer:
```suggestion
    
```

Sources/ArcGISToolkit/Extensions/SwiftUI/View.swift
```diff
@@ -108,3 +108,16 @@ extension View {
         }
     }
 }
+
+extension View {
+    /// Modify a view in a closure. This is useful for using an `if` statement
```
dfeinzimer: I think indicative form is more common for SwiftUI modifier doc.
```suggestion
    /// Modifies this view in a closure. This is useful for using an `if` statement
```

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilterViewModel.swift
```diff
@@ -54,6 +54,9 @@ final class FloorFilterViewModel: ObservableObject {
     /// The selected site, floor, or level.
     @Published private(set) var selection: FloorFilterSelection?
     
+    /// Whether the selector is showing facilities from all sites.
```
dfeinzimer:
```suggestion
    /// A Boolean value that indicates whether the selector is showing facilities from all sites.
```

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
dfeinzimer, line 360: Nitpick

```suggestion
            .onAppear {
                guard model.selection?.facility != nil, !isCollapsed else { return }
                scrollToSelectedLevel(with: scrollView)
            }
            .onChange(of: isCollapsed) {
                guard !isCollapsed else { return }
                scrollToSelectedLevel(with: scrollView)
            }
            .onChange(of: model.selection) { _, _ in
                guard model.selection?.facility != nil, !isCollapsed else { return }
                scrollToSelectedLevel(with: scrollView)
            }
```

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
```diff
@@ -282,3 +189,227 @@ public extension FloorFilter {
...
+    }
+}
+
+/// A level button within the level selector.
+private struct LevelButton: View {
+    @EnvironmentObject var model: FloorFilterViewModel
+    
+    let level: FloorLevel
+    let isSelected: Bool
+    @Binding var isCollapsed: Bool
```
dfeinzimer:
```suggestion
    @Binding var isCollapsed: Bool
    let isSelected: Bool
    let level: FloorLevel
```

⚠️ You'll also need to update where the `LevelButton` initializer is used.

Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
```diff
@@ -15,244 +15,195 @@
...
-                                .frame(maxWidth: .infinity, alignment: .leading)
-                        }
-                    }
+        NavigationStack(path: $navigationPath) {
+            VStack {
+                if model.sites.count > 1 {
+                    SiteList(isPresented: $isPresented)
+                } else {
+                    FacilityList(isPresented: $isPresented, site: nil)
                 }
```
CalebRas: If both of the child views (`SiteList` and `FacilityList`) have a `VStack`, is this `VStack` needed? Should it be a `Group` instead? Not sure...

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
```diff
@@ -24,7 +24,7 @@ import SwiftUI
 ///
 /// | iPhone | iPad |
 /// | ------ | ---- |
-/// | ![image](https://user-images.githubusercontent.com/3998072/202811733-dcd640e9-3b27-43a8-8bec-fd9aeb6798c7.png) | ![image](https://user-images.githubusercontent.com/3998072/202811772-bf6009e7-82ec-459f-86ae-6651f519b2ef.png) |
+/// | ![image](https://github.com/user-attachments/assets/f7c8747b-1269-48d8-9dac-b783a63d3f15) | ![image](https://github.com/user-attachments/assets/f2e338f9-f29d-4621-9823-098037e9a135) |
```
CalebRas, line 27: The previous screenshots didn't do this, but consider updating the new ones to follow the "Status Bar" [Marketing Resources and Identity Guidelines](https://developer.apple.com/app-store/marketing/guidelines/). You can set it up by running this command for each simulator you are going to use:
```
xcrun simctl status_bar "iPad (A16)" override --time 9:41 --dataNetwork wifi --wifiMode active --wifiBars 3 --cellularMode active --cellularBars 4 --batteryState charged --batteryLevel 100
```

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
```diff
@@ -113,65 +110,12 @@ public struct FloorFilter: View {
...
-                    levelSelector
-                    Divider()
-                }
-                sitesAndFacilitiesButton
-            }
-        }
-        .frame(width: levelSelectorWidth)
-        .esriBorder()
-        .frame(
-            maxWidth: horizontalSizeClass == .compact ? .infinity : nil,
```
CalebRas, line 169: Due to these changes, it looks like `FloorFilter.horizontalSizeClass` is no longer used anywhere. Could you remove the property?

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
```diff
@@ -180,79 +124,42 @@ public struct FloorFilter: View {
...
-    
     public var body: some View {
-        HStack(alignment: .bottom) {
-            if alignment.horizontal == .trailing {
-                siteAndFacilitySelector
-                levelSelectorContainer
-            } else {
-                levelSelectorContainer
-                siteAndFacilitySelector
+        FloorFilterBody(width: levelSelectorWidth, isTopAligned: isTopAligned)
```
CalebRas: The `FloorFilter.isTopAligned` property is now only used here, so you might want to consider removing it. Up to you, though.

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
CalebRas: You could also use a key path here, but it doesn't make it any shorter, so I am not sure it is an improvement. Up to you.

```suggestion
                            $0.onScrollGeometryChange(
                                for: CGFloat.self,
                                of: \.contentSize.height
                            ) { _, newValue in
                                contentHeight = newValue
                            }
```

Sources/ArcGISToolkit/Extensions/SwiftUI/View.swift
```diff
@@ -108,3 +108,16 @@ extension View {
...
+    /// to determine what modifier, if any, should be applied.
+    @ViewBuilder
+    func modify(@ViewBuilder _ transform: (Self) -> (some View)?) -> some View {
```
CalebRas: It seems a little odd that this is in it's own extension instead of grouped with the other View modifiers in this file. It looks like this isn't the first time this happened, though, so I am not sure what the right answer is here. Just calling it out in case it should be changed.

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
CalebRas: There is a bit of an odd shadow behind the controls on "26" platforms. Is it intended?

Simulator Screenshot - iPad (A16) - 2026-01-26 at 10 10 12  Simulator Screenshot - iPhone 17 Pro - 2026-01-26 at 10 12 48

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
```diff
@@ -282,3 +189,227 @@ public extension FloorFilter {
...
+    @ViewBuilder private var siteSelectorButton: some View {
+        Button {
+            isSiteSelectorPresented.toggle()
+        } label: {
+            Image(systemName: "building.2")
+                .frame(minWidth: FloorFilterBody.buttonSize, maxWidth: .infinity)
+                .frame(height: FloorFilterBody.buttonSize)
+                .font(.system(size: FloorFilterBody.fontSize))
+                .contentShape(FloorFilterBody.buttonShape)
+        }
```
CalebRas, line 245: The Floor Filter buttons appear larger than the toolbar ones on Mac Catalyst. Is that intentional? (I changed the `siteSelectorButton` image in the screenshot to compare it with the sidebar button in the toolbar.)

Screenshot 2026-01-26 at 10 34 35 AM

CalebRas (commented):
UI design looks great overall! Mostly just some "linting" suggestions:

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
rolson: No, that looks different than what I was seeing. Something is broken.

Sources/ArcGISToolkit/Components/FloorFilter/LevelSelector.swift
rolson: I think the way it is now makes more sense. The UI components to the floor filter component that you see on the map are in one file. The sites/facility picker views are in another. Note the button to select the sites is not in that second file. So moving the level selector out makes less sense to me.

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
rolson: I've made these changes, but I'm not necessarily a fan of all changes like this. The order of these determines the order you pass these parameters upon construction of the struct. And that order should not be dictated by the alphabet, but based on importance.

Sources/ArcGISToolkit/Components/FloorFilter/FloorFilter.swift
rolson: what size looks right to you?

rolson: I updated the sizing, let me know thoughts now please

CalebRas (dismissed):
Looks great! Thanks for making those changes.

Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
```diff
@@ -22,237 +22,187 @@ struct SiteAndFacilitySelector: View {
...
-    }
-    
-    /// A Boolean value indicating whether the floor aware data contains more than one site.
-    var multipleSitesAreAvailable: Bool {
-        viewModel.sites.count > 1
+        .searchable(
+            text: $searchText,
+            placement: .navigationBarDrawer,
+            prompt: String.filterFacilities
+        )
```
dfeinzimer: The text in here seems very dimmed, almost to the point of being not visible, until I give it focus. Not much we can do I suppose.

Image

dfeinzimer (commented):
The component's UI test is still failing on simulator and device for me. `FloorFilterTestView` also still has the 400x400 frame that was removed from the tutorial and example code.

CalebRas (dismissed):

dfeinzimer (dismissed):

Sources/ArcGISToolkit/Components/FloorFilter/SiteAndFacilitySelector.swift
rolson: fixed now

rolson:
Please re-review. Fixed attribution bar jump and the issue where search bar was only showing half-way.

Simulator Screen Recording - swift-test-ipad - 2026-01-30 at 10 07 49

dfeinzimer (approved):

CalebRas (approved):

rolson:
thank you both!

PR #1365, opened 2026-01-28
Implement Filter View

mhdostal:
Apollo #1557 - Filter Utility Network Association candidate features based on fields/values.

This is the first set of functionality. Future work will include support for Domains and Date Field types (Apollo #1586 & #1587)

Note: there will need to be some small refactoring in order to make this a true component and not dependent on Utility Network Associations.


Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+            .id("BorderedAddButton")
+            .buttonBorderShape(.automatic)
+            .buttonStyle(.borderedProminent)
+            .shadow(radius: 8)
+        }
+    }
+}
+
+/// A button, sans border, used to add a `FieldFilter` to the list of current `FieldFilters`.
+private struct NoBorderAddbutton: View {
```
dfeinzimer:
```suggestion
private struct NoBorderAddButton: View {
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+                        // Scroll to the last message when messages change
+                        if let lastFieldFilter = model.fieldFilters.last {
+                            withAnimation {
+                                proxy.scrollTo(lastFieldFilter.id, anchor: .bottom)
+                            }
+                        }
+                    }
+                    Spacer()
+                    HStack {
+                        NoBorderAddbutton()
```
dfeinzimer:
```suggestion
                        NoBorderAddButton()
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+                .bold()
+            }
+            .id("NoBorderAddbutton")
+        }
+    }
+}
+
+/// A view representing a single `FieldFilter`, with options to select a field and condition/operation, and to set a value for the operation.
+private struct FieldView: View {
+    /// The mdoel used by the view.
```
dfeinzimer:
```suggestion
    /// The model used by the view.
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+            } else if let value = Double(fieldFilter.value) {
+                fieldFilter.value = String(value * -1)
+            }
+        } label: {
+            Image(systemName: "plus.forwardslash.minus")
+        }
+        .tint(.blue)
+    }
+    
+    /// Determines the conditions to dsiplay for the given `FieldFilter` field type.
```
dfeinzimer:
```suggestion
    /// Determines the conditions to display for the given `FieldFilter` field type.
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+                HStack {
+                    Image(systemName: "plus")
+                        .foregroundColor(.white)
+                        .padding(4)
+                        .background(Circle().fill(Color.blue))
+                    Text("Add Condition")
+                }
+                .bold()
+            }
+            .id("NoBorderAddbutton")
```
dfeinzimer:
```suggestion
            .id("NoBorderAddButton")
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+                        })
+                    }
+                    .onChange(of: model.fieldFilters) {
+                        // Scroll to the last message when messages change
+                        if let lastFieldFilter = model.fieldFilters.last {
+                            withAnimation {
+                                proxy.scrollTo(lastFieldFilter.id, anchor: .bottom)
+                            }
+                        }
+                    }
```
dfeinzimer: Bit of funkiness here if you're scrolled to the top and remove the first filter, this'll scroll you all the way down to the bottom. Maybe we only scroll on an add?

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+                            }
+                        }
+                        .onDelete(perform: { offsets in
+                            withAnimation {
+                                model.fieldFilters.remove(atOffsets: offsets)
+                            }
+                        })
+                    }
+                    .onChange(of: model.fieldFilters) {
+                        // Scroll to the last message when messages change
```
dfeinzimer:
```suggestion
                        // Scroll to the last filter when the set changes
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
dfeinzimer, line 1: Thoughts on moving this file and `FilterViewModel.swift` out of `FeatureFormView/` (1) since there's no direct connection tying the two together and (2) for the eventuality when it becomes its own standalone component or used in other components like a table view?

Maybe a `Common/FilterView/` directory?

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+            ScrollViewReader { proxy in
+                if model.fieldFilters.isEmpty {
+                    Spacer()
+                    Image(systemName: "line.3.horizontal.decrease.circle")
+                        .font(.largeTitle)
+                        .foregroundStyle(.gray)
+                        .padding()
+                    Text("No conditions added")
+                        .font(.title3)
+                        .fontWeight(.bold)
```
dfeinzimer:
```suggestion
                    Text(
                        "No conditions added",
                        bundle: .toolkitModule,
                        comment: "A label indicating no field filtering conditions have been added."
                    )
                    .font(.title3)
                    .fontWeight(.bold)
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+                    Text("Show features that meet all the conditions")
+                        .font(.subheadline)
+                        .foregroundStyle(.secondary)
+                        .padding(.bottom)
```
dfeinzimer:
```suggestion
                    Text(
                        "Show features that meet all the conditions",
                        bundle: .toolkitModule,
                        comment: "A label indicating the intended outcome of using the filter view."
                    )
                    .font(.subheadline)
                    .foregroundStyle(.secondary)
                    .padding(.bottom)
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+                    HStack {
+                        Spacer()
+                        BorderedAddButton()
+                        Spacer()
+                    }
+                    Spacer()
```
dfeinzimer: `ContentUnavailableView` might make sense here. Something like this yields a pretty similar result without any spacers or HStacks:

```swift
ContentUnavailableView {
    Label {
        Text(
            "No conditions added",
            bundle: .toolkitModule,
            comment: "A label indicating no field filtering conditions have been added."
        )
        .font(.title3)
        .fontWeight(.bold)
        Text(
            "Show features that meet all the conditions",
            bundle: .toolkitModule,
            comment: "A label indicating the intended outcome of using the filter view."
        )
        .font(.subheadline)
        .foregroundStyle(.secondary)
        .padding(.bottom)
    } icon: {
        Image(systemName: "line.3.horizontal.decrease.circle")
            .font(.largeTitle)
            .foregroundStyle(.gray)
            .padding()
    }
} actions: {
    BorderedAddButton()
}
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+                                            .imageScale(.large)
+                                    }
+                                }
+                            }
+                        }
+                        .onDelete(perform: { offsets in
+                            withAnimation {
+                                model.fieldFilters.remove(atOffsets: offsets)
+                            }
+                        })
```
dfeinzimer:
```suggestion
                        .onDelete { offsets in
                            withAnimation {
                                model.fieldFilters.remove(atOffsets: offsets)
                            }
                        }
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+                    }
+                    Spacer()
+                } else {
+                    List {
+                        ForEach(model.fieldFilters, id: \.self) { filter in
+                            Section {
+                                FieldView(fieldFilter: filter)
+                            } header: {
+                                HStack {
+                                    Text(filter.name)
```
dfeinzimer: Seems like `filter.name` is always "Condition". Is that intended? If so, maybe we should just use a localized version of "Condition" here?

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+                                    Spacer()
+                                    Menu {
+                                        // Duplicate the current filter
```
dfeinzimer:
```suggestion
                                        // Duplicate the current filter.
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+                            .buttonStyle(.borderless)
+                            .padding()
+                        Spacer()
+                    }
+                }
+            }
+            Spacer()
+                .toolbar {
+                    ToolbarItem(placement: .topBarLeading) {
+                        DismissButton(kind: .cancel){
```
dfeinzimer:
```suggestion
                        DismissButton(kind: .cancel) {
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+                            model.cancel()
+                        }
+                    }
+                    ToolbarItem(placement: .topBarTrailing) {
+                        DismissButton(kind: .confirm){
```
dfeinzimer:
```suggestion
                        DismissButton(kind: .confirm) {
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+    /// - Parameter filter: The `FieldFilter` to delete.
+    /// - Returns: The delete `Button`.
+    private func deleteButton(_ filter: FieldFilter) -> Button<some View> {
+        Button(role: .destructive) {
+            if let index = model.fieldFilters.firstIndex(of: filter) {
+                model.fieldFilters.remove(at: index)
+            }
+        } label: {
+            Text("Delete")
+            Image(systemName: "trash")
```
dfeinzimer:
```suggestion
            Label {
                Text(
                    "Delete",
                    bundle: .toolkitModule,
                    comment: "A label for a button to delete a field filter."
                )
            } icon: {
                Image(systemName: "trash")
            }
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+        Button {
+            if let index = model.fieldFilters.firstIndex(of: filter) {
+                let newFilter = filter.copy()
+                withAnimation {
+                    model.fieldFilters.insert(newFilter, at: index + 1)
+                }
+            }
+        } label: {
+            Text("Duplicate")
+            Image(systemName: "document.on.document")
```
dfeinzimer:
```suggestion
            Label {
                Text(
                    "Duplicate",
                    bundle: .toolkitModule,
                    comment: "A label for a button to duplicate a field filter."
                )
            } icon: {
                Image(systemName: "document.on.document")
            }
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+                        .padding(4)
+                        .background(Circle().fill(Color.blue))
+                    Text("Add Condition")
+                }
+                .bold()
+            }
+            .id("NoBorderAddbutton")
+        }
+    }
+}
```
dfeinzimer: In #1363, which ~~should be merged soon~~ is now merged, Ryan added [a modifier](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/296453e6fa68ecb05c4011e51c61415a40e446d8/Sources/ArcGISToolkit/Extensions/SwiftUI/View.swift#L109-L118) which I think could be used here to merge `BorderedAddButton` and `NoBorderAddButton` into one view.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+                    let newFilter = FieldFilter(field: model.fields.first ?? Field(type: .blob, name: "Empty", alias: "Empty"))
+                    model.fieldFilters.append(newFilter)
+                }
+            } label: {
+                HStack {
+                    Image(systemName: "plus")
+                        .imageScale(.large)
+                        .padding(4)
+                    Text("Add Condition")
+                        .padding(.trailing)
```
dfeinzimer:
```suggestion
                    Text(
                        "Add Condition",
                        bundle: .toolkitModule,
                        comment: "A label for a button to add a new field filtering condition."
                    )
                    .padding(.trailing)
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+                    let newFilter = FieldFilter(field: model.fields.first ?? Field(type: .blob, name: "Empty", alias: "Empty"))
+                    model.fieldFilters.append(newFilter)
+                }
+            } label: {
+                HStack {
+                    Image(systemName: "plus")
+                        .foregroundColor(.white)
+                        .padding(4)
+                        .background(Circle().fill(Color.blue))
+                    Text("Add Condition")
```
dfeinzimer:
```suggestion
                    Text(
                        "Add Condition",
                        bundle: .toolkitModule,
                        comment: "A label for a button to add a new field filtering condition."
                    )
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+                Picker("Condition", selection: $fieldFilter.condition) {
+                    ForEach(conditions, id: \.self) { condition in
+                        Text(condition.displayName)
+                    }
+                }
+                .pickerStyle(MenuPickerStyle())
+            }
+            .onAppear(perform: {
+                conditions = fieldConditions()
+            })
```
dfeinzimer:
```suggestion
            .onAppear {
                conditions = fieldConditions()
            }
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+
```
dfeinzimer, line 295:
```suggestion
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+    init(fieldFilter: FieldFilter) {
+        self.fieldFilter = fieldFilter
+    }
+    
+    var body: some View {
+        Group {
+            // Field
+            if model.fields.isEmpty {
+                HStack {
+                    Text("Field")
```
dfeinzimer:
```suggestion
                    Text.field
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+#endif
+        } else {
+            .default
+        }
+    }
+}
+
+extension Field {
+    /// Determines the display title for the `Field`.
+    /// - Returns: A string representing the display fitle for the `Field`.
```
dfeinzimer:
```suggestion
    /// - Returns: A string representing the display title for the `Field`.
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+                    Spacer()
+                    Text(fieldFilter.field.title())
+                }
+            } else {
+                HStack {
+                    Picker("Fields", selection: $fieldFilter.field) {
+                        ForEach(model.fields, id: \.self) { field in
+                            Text(field.title())
+                        }
+                    }
```
dfeinzimer:
```suggestion
                    Picker(selection: $fieldFilter.field) {
                        ForEach(model.fields, id: \.self) { field in
                            Text(field.title())
                        }
                    } label: {
                        Text.fields
                    }
```


The following could be added under `field` in `Text.swift`:
```swift
    /// Localized text for the word "Fields".
    static var fields: Self {
        .init(
            "Fields",
            bundle: .toolkitModule,
            comment: "Fields in a dataset."
        )
    }
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+            }
+            
+            // Condition
+            HStack {
+                Picker("Condition", selection: $fieldFilter.condition) {
+                    ForEach(conditions, id: \.self) { condition in
+                        Text(condition.displayName)
+                    }
+                }
+                .pickerStyle(MenuPickerStyle())
```
dfeinzimer:
```suggestion
Picker(selection: $fieldFilter.condition) {
                    ForEach(conditions, id: \.self) { condition in
                        Text(condition.displayName)
                    }
                } label: {
                    Text(
                        "Condition",
                        bundle: .toolkitModule,
                        comment: "A label for a control to pick a condition to filter fields in a table against."
                    )
                }
                .pickerStyle(.menu)
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+                    Text(fieldFilter.field.title())
+                }
+            } else {
+                HStack {
+                    Picker("Fields", selection: $fieldFilter.field) {
+                        ForEach(model.fields, id: \.self) { field in
+                            Text(field.title())
+                        }
+                    }
+                    .pickerStyle(MenuPickerStyle())
```
dfeinzimer:
```suggestion
                    .pickerStyle(.menu)
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+            
+            // Value
+            HStack {
+                Text("Value")
+                Spacer()
+                TextField(
+                    text: $fieldFilter.value,
+                    prompt: Text("Enter a value"),
+                    label: {
+                        Label("Value", systemImage: "swift")
```
dfeinzimer: It looks like the Swift icon isn't coming through; is that intended/needed?

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+                    }
+                )
```
dfeinzimer:
```suggestion
                Text.value
                Spacer()
                TextField(
                    text: $fieldFilter.value,
                    prompt: Text(
                        "Enter a value",
                        bundle: .toolkitModule,
                        comment: "A prompt for a text field to enter a value."
                    ),
                    label: {
                        Text.value
                    }
                )
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+        }
+    }
+}
+
+extension Field {
+    /// Determines the display title for the `Field`.
+    /// - Returns: A string representing the display fitle for the `Field`.
+    func title() -> String {
+        alias.isEmpty ? name : alias
+    }
```
dfeinzimer: Should we make this a computed property since it takes no parameters?


```suggestion
    var title: String {
        alias.isEmpty ? name : alias
    }
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,369 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ArcGIS
+import SwiftUI
+
+/// A view allowing the user to assembly a list of `FieldFilters` used to filter a list of features.
+struct FilterView: View {
+    /// The model used by the view.
+    @Environment(FilterViewModel.self) private var model
```
dfeinzimer: Seems like it might be better to make this a bindable property, that way the need for a model is explicitly required by users.


```suggestion
    /// The model used by the view.
    @Bindable var model: FilterViewModel
```

Then, using the initializer looks like this:
```swift
FilterView(model: model)
```
instead of this:
```swift
FilterView()
    .environment(model)
```

Then, `FilterView` can internally put the model into the environment for dependent views, e.g. after the `background` modifier in `body`:
```swift
        .background(Color(.systemGroupedBackground))
        .environment(model)
```

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,297 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ArcGIS
+import Observation
+import SwiftUI
```
dfeinzimer, line 17:
```suggestion
import ArcGIS
import Foundation
```

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,297 @@
...
+        }
+    }
+    /// The list of field filters the user has created.
+    var fieldFilters: [FieldFilter]
+    
+    /// The original list of field filters, used when user cancels changes.
+    private var originalFieldFilters: [FieldFilter]
+    
+    /// Determines if the filter view is presented.
+    var isFilterViewPresented = false
```
dfeinzimer:
```suggestion
    /// A Boolean value indicating whether the filter view is presented.
    var filterViewIsPresented = false
```

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,297 @@
...
+                clause.append(" AND ")
+            }
+            clause.append(fieldFilter.query())
+        }
+        return clause
+    }
+    
+    /// Initializes a filter view model.
+    /// - Parameter featureTable: The feature table for the filter.
+    init(featureTable: ArcGISFeatureTable? = nil, fieldFilters: [FieldFilter] = [FieldFilter]()) {
```
dfeinzimer: If I understand correctly, these are intended to be used independently, correct? E.g. if you pass a table and filters `featureTable`'s `didSet` `Task` will overwrite them. Should we provide two separate initializers, one that takes a table and one that takes a set of fields?

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,297 @@
...
+        didSet {
+            if oldValue !== field {
+                // We're changing the field, so reset the condition
+                condition = firstCondition()
+            }
+        }
+    }
+    
+    /// The name of the filter.
+    var name = "Condition"
```
dfeinzimer: Is there a future use for this? Not sure if I see that it's needed.

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,297 @@
...
+                .notEqual,
+                .isOp,
+                .isNot,
+                .greaterThan,
+                .greaterThanOrEqual,
+                .lessThan,
+                .lessThanOrEqual:
+            return "\(field.name) \(condition.sqlOperator) \(value)"
+        }
+    }
```
dfeinzimer: I feel this is a bit more readable, take or leave it.

```suggestion
    /// Creates the SQL query for this field filter.
    /// - Returns: A string representing a SQL query for the specified `field`, `condition`, and `value`.
    func query() -> String {
        switch condition {
        case .startsWith:
            "\(field.name) \(condition.sqlOperator) '\(value)%'"
        case .endsWith:
            "\(field.name) \(condition.sqlOperator) '%\(value)'"
        case .contains, .doesNotContain:
            "\(field.name) \(condition.sqlOperator) '%\(value)%'"
        case .isBlank, .isNotBlank, .isEmpty, .isNotEmpty:
            "\(field.name) \(condition.sqlOperator)"
        case .equal, .notEqual, .isOp, .isNot, .greaterThan, .greaterThanOrEqual, .lessThan, .lessThanOrEqual:
            "\(field.name) \(condition.sqlOperator) \(value)"
        }
    }
```

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,297 @@
...
+extension FieldFilter: Hashable {
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(id.hashValue)
+    }
+}
+
+extension FilterViewModel {
+    /// Returns the list of fields supported for filtering.
+    /// - Parameter allFields: The list of all candidate fields to filter by.
+    /// - Returns: The final list of supported fileds to filter by.
```
dfeinzimer:
```suggestion
    /// - Returns: The final list of supported fields to filter by.
```

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,297 @@
...
+        allFields.filter { field in
+            (field.type?.isNumeric ?? false) ||
+            field.type == .text ||
+            field.type == FieldType.oid
+        }
+    }
+    
+    /// Returns the list of fields supported for filtering by when used in a Utility Network workflow.
+    /// - Parameter allFields: The list of all candidate fields to filter by.
+    /// - Returns: The final list of supported fileds to filter by. This will automatically filter out the
```
dfeinzimer:
```suggestion
    /// - Returns: The final list of supported fields to filter by. This will automatically filter out the
```

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,297 @@
...
+    /// `ASSETGROUP` and `ASSETTYPE` fields, as those are speciall fields for Utility Networks.
```
dfeinzimer:
```suggestion
    /// `ASSETGROUP` and `ASSETTYPE` fields, as those are special fields for Utility Networks.
```

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,297 @@
...
+    private func unaSupportedFields(_ allFields: [Field]) -> [Field] {
```
dfeinzimer: How about `supportedUNFields`? Fits more with `supportedFields` and I'm not sure the "association"/'A' in "una" piece is relevant in this context.

```suggestion
    private func supportedUNFields(_ allFields: [Field]) -> [Field] {
```

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,297 @@
...
+            .isNotEmpty
+        ]
+        if fieldIsNullable {
+            ops.append(contentsOf: [.isBlank, .isNotBlank])
+        }
+        return ops
+    }
+    
+    /// Returns a list of appropriate operations for numeric fields.
+    /// - Returns: The list of appropriate operations for numeric fields.
```
dfeinzimer: Seems redundant as it adds no additional info. Take or leave.

```suggestion
```

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,297 @@
...
+    case doesNotContain = "does not contain the text"
+    case isBlank = "is blank"
+    case isNotBlank = "is not blank"
+    case isEmpty = "is empty"
+    case isNotEmpty = "is not empty"
+    
+    /// Returns a list of appropriate operations for text fields.
+    /// - Parameter fieldIsNullable: Specifies whether the field is nullable; if `true`, `empty` and `notEmpty` operators
+    /// are added to the list. If `false`, no additional operators are added.
+    /// - Returns: The list of appropriate operations for text fields.
```
dfeinzimer: Seems redundant as it adds no additional info. Take or leave.

```suggestion
```

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
mhdostal: Upon further review, the initializer which takes a set of fields was only really useful for testing, so I've removed it.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
mhdostal: For continuity with this PR, I'll move it out after this is merged; there are two other FilterView issues to implement.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
mhdostal: I refactored it to use the method Kotlin uses, which is "Condition x" where "x" is the index of the element in the list (1-based).

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
mhdostal: I can see how that would work for the modifiers on the Button itself, but inside the `Label`, there are different modifiers on the `plus` image; I can' see how that would work without creating two buttons again. Thoughts?

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
mhdostal: Removed the image.

mhdostal:
With the exception of [this](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/1365#discussion_r2748142948) question, this is ready for re-review.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
dfeinzimer: Sure, this is what I had in mind:

```suggestion
/// A button to add a `FieldFilter` to the list of current `FieldFilters`.
private struct AddButton: View {
    let useBorderedStyle: Bool
    
    init(useBorderedStyle: Bool = false) {
        self.useBorderedStyle = useBorderedStyle
    }
    
    @Environment(FilterViewModel.self) private var model
    var body: some View {
        HStack {
            Button {
                withAnimation {
                    let newFilter = FieldFilter(field: model.fields.first ?? Field(type: .blob, name: "Empty", alias: "Empty"))
                    model.fieldFilters.append(newFilter)
                }
            } label: {
                HStack {
                    Image(systemName: "plus")
                        .modify {
                            if useBorderedStyle {
                                $0.imageScale(.large)
                                .padding(4)
                            } else {
                                $0.foregroundColor(.white)
                                .padding(4)
                                .background(Circle().fill(Color.blue))
                            }
                        }
                    Text(
                        "Add Condition",
                        bundle: .toolkitModule,
                        comment: "A label for a button to add a new field filtering condition."
                    )
                    .modify {
                        if useBorderedStyle {
                            $0.padding(.trailing)
                        }
                    }
                }
                .bold()
            }
            .id("AddButton")
            .modify {
                if useBorderedStyle {
                    $0.buttonBorderShape(.automatic)
                    .buttonStyle(.borderedProminent)
                    .shadow(radius: 8)
                }
            }
        }
    }
}
```

Usage then becomes `AddButton()` for borderless and `AddButton(useBorderedStyle: true)` for bordered.

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,275 @@
...
+    /// `ASSETGROUP` and `ASSETTYPE` fields, as those are special fields for Utility Networks.
+    private func supportedUNFields(_ allFields: [Field]) -> [Field] {
+        allFields.filter { field in
+            ((field.type?.isNumeric ?? false) ||
+             field.type == .text ||
+             field.type == FieldType.oid) &&
+            field.name != "assetgroup" &&
+            field.name != "assettype"
+        }
+    }
```
dfeinzimer, line 197: This method could make use of the other:


```suggestion
    private func supportedUNFields(_ allFields: [Field]) -> [Field] {
        supportedFields(allFields).filter { field in
            field.name != "assetgroup" && field.name != "assettype"
        }
    }
```

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,275 @@
...
+
+@MainActor @Observable
+class FilterViewModel {
+    /// The feature table containing the fields to filter on.
+    public var featureTable: ArcGISFeatureTable? {
+        didSet {
+            if let featureTable {
+                Task {
+                    try? await featureTable.load()
+                    fields = supportedUNFields(featureTable.fields)
```
dfeinzimer, line 27: Should whether this is used or `supportedFields` is used be controllable via the initializer? It seems like for default general usage we wouldn't want to do the extra filtering. Or is this something that's set to be changed soon anyway?

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,275 @@
...
+        }
+        return clause
+    }
+    
+    /// Initializes a filter view model.
+    /// - Parameter featureTable: The feature table for the filter.
+    init(featureTable: ArcGISFeatureTable? = nil) {
+        self.featureTable = featureTable
+    }
+
```
dfeinzimer:
```suggestion
    
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FormElements/UtilityAssociationsFormElement/UtilityAssociationFeatureCandidatesView.swift
```diff
@@ -35,6 +35,10 @@ extension FeatureFormView {
         @State private var queryIsRunning = false
         /// The task for the current query.
         @State private var queryTask: Task<Void, Never>?
+        /// The model for the filter view
+        @State private var filterViewModel = FilterViewModel()
+
```
dfeinzimer:
```suggestion
        /// The attribute expression to query and filter candidates against.
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FormElements/UtilityAssociationsFormElement/UtilityAssociationFeatureCandidatesView.swift
```diff
@@ -35,6 +35,10 @@ extension FeatureFormView {
         @State private var queryIsRunning = false
         /// The task for the current query.
         @State private var queryTask: Task<Void, Never>?
+        /// The model for the filter view
+        @State private var filterViewModel = FilterViewModel()
```
dfeinzimer: Alphabetization

```suggestion
        /// The model for the filter view
        @State private var filterViewModel = FilterViewModel()
        /// The parameters for retrieving the next page of results.
        @State private var nextQueryParameters: QueryParameters?
        /// A Boolean value indicating whether a filter based query is running.
        @State private var queryForFilterPhraseIsRunning = false
        /// A Boolean value indicating if the initial candidate query is complete.
        @State private var queryForFirstPageIsComplete = false
        /// A Boolean value indicating whether a query is running.
        @State private var queryIsRunning = false
        /// The task for the current query.
        @State private var queryTask: Task<Void, Never>?
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FormElements/UtilityAssociationsFormElement/UtilityAssociationFeatureCandidatesView.swift
```diff
@@ -72,6 +76,26 @@ extension FeatureFormView {
...
+                    queryForFirstPageIsComplete = true
+                }
+                
+            }
+            .sheet(isPresented: $filterViewModel.filterViewIsPresented) {
+                FilterView(model: filterViewModel) {
+                    queryForFirstPageIsComplete = false
+                    whereClause = filterViewModel.whereClause()
+                }
+            }
```
dfeinzimer: This could avoid the unstructured task by using a task modifier:


```suggestion
            .sheet(isPresented: $filterViewModel.filterViewIsPresented) {
                FilterView(model: filterViewModel) {
                    queryForFirstPageIsComplete = false
                    whereClause = filterViewModel.whereClause()
                }
            }
            .task(id: whereClause) {
                candidates.removeAll()
                let parameters = QueryParameters()
                parameters.whereClause = filterViewModel.whereClause()
                queryFeatures(parameters: parameters)
                await queryTask?.value
                queryForFirstPageIsComplete = true
            }
```

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,275 @@
...
+    /// Applies the current field filters.
+    func apply() {
+        filterViewIsPresented.toggle()
+        self.originalFieldFilters = fieldFilters
```
dfeinzimer:
```suggestion
       originalFieldFilters = fieldFilters
```

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,275 @@
...
+    var fieldFilters = [FieldFilter]()
+    
+    /// The original list of field filters, used when user cancels changes.
+    private var originalFieldFilters = [FieldFilter]()
+    
+    /// A Boolean value indicating whether the filter view is presented.
+    var filterViewIsPresented = false
+    
+    /// The list of fields generated from the `featureTable`.
+    var fields = [Field]()
```
dfeinzimer: Should this be externally settable?


```suggestion
    private(set) var fields = [Field]()
```

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,275 @@
...
+            return "= ''"
+        case .isNotEmpty:
+            return "<> ''"
+        }
+    }
+    
+    /// The name used to display the operator.
+    var displayName: String {
+        return self.rawValue
+    }
```
dfeinzimer: These unfortunately won't get properly localized. I think we'll need to provide localized labels:

```suggestion
    /// The name used to display the operator.
    var displayName: Text {
        switch self {
        case .isOp:
            Text(
                "is",
                bundle: .toolkitModule,
                comment: "A label for the \"is\" attribute query condition."
            )
        // ... cases for each operator with words ...
        default:
            Text(self.rawValue)
        }
    }
```

Then the point of use in `FieldView` could just be `condition.displayName` instead of `Text(condition.displayName)`.

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
mhdostal: I'll address this with the other refactoring for general usage. Thanks.

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
mhdostal: Perfect, thanks.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FormElements/UtilityAssociationsFormElement/UtilityAssociationFeatureCandidatesView.swift
```diff
@@ -189,9 +223,13 @@ extension FeatureFormView {
...
             queryIsRunning = true
             queryTask = Task {
-                let result = try? await source.queryFeatures(assetType: assetType, parameters: parameters)
-                candidates.append(contentsOf: result?.candidates ?? [])
-                nextQueryParameters = result?.nextQueryParams
+                do {
+                    let result = try? await source.queryFeatures(assetType: assetType, parameters: parameters)
+                    candidates.append(contentsOf: result?.candidates ?? [])
+                    nextQueryParameters = result?.nextQueryParams
+                } catch {
```
dfeinzimer, line 230: New warning introduced here:
> 'catch' block is unreachable because no errors are thrown in 'do' block

dfeinzimer (commented):
Nearly there, just a warning I found to resolve.

One other thing to be done would be to update the translation files. Since we've added the script, I've gotten in the habit of doing it in the same PR that adds/removes strings. All you need to do is run `./update-translations`.

dfeinzimer (approved):

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,386 @@
...
+            }
+        } label: {
+            Image(systemName: "plus.forwardslash.minus")
+        }
+        .tint(.blue)
+    }
+    
+    /// Determines the conditions to display for the given `FieldFilter` field type.
+    /// - Returns: A list of conditions appropriate for the given `FieldFilter` field type.
+    private func fieldConditions() -> [FilterOperator] {
```
dfeinzimer: Some alternative suggestions:
```suggestion
    /// The operators that can be applied for the given `FieldFilter` field type.
    /// - Returns: A list of operators appropriate for the given `FieldFilter` field type.
    private var applicableOperators: [FilterOperator] {
```
or
```suggestion
    /// The operators supported for the given `FieldFilter` field type.
    /// - Returns: A list of operators appropriate for the given `FieldFilter` field type.
    private var supportedOperators: [FilterOperator] {
```

dfeinzimer (dismissed):

mhdostal:
@njarecha This is ready for review.

dfeinzimer (dismissed):

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,386 @@
...
+    public static func == (lhs: ArcGIS.Field, rhs: ArcGIS.Field) -> Bool {
+        lhs.toJSON() == rhs.toJSON()
+    }
+}
+
+extension Field: @retroactive Hashable {
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(toJSON())
+    }
+}
```
philium: We should avoid retroactive conformances of ArcGIS types in the toolkit. If the behavior is essential, then it should be added to ArcGIS itself. In many cases the conformance isn't essential, like here. Either the JSON could be stored (preferred, so as not to potentially re-compute it multiple times) or added as a private `json` property to `Field` and used instead of the field itself. Even the property approach would potentially save re-computing the JSON in case of a hash collision.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,386 @@
...
+            Image(systemName: "plus.forwardslash.minus")
+        }
+        .tint(.blue)
+    }
+    
+    /// The operators supported for the given `FieldFilter` field type.
+    /// - Returns: A list of operators appropriate for the given `FieldFilter` field type.
+    private var supportedOperators: [FilterOperator] {
+        (fieldFilter.field.type?.isNumeric ?? false) ? FilterOperator.numericFilterOperators() : FilterOperator.textFilterOperators(fieldFilter.field.isNullable)
+    }
```
dfeinzimer: I just noticed this logic is duplicated in `FieldFilter.firstCondition()`. It's going to grow a bit more when we add support for date fields. Rather than maintaining it in two places, should we break it out into a property on `FieldFilter`? Then, it could be used both here and in `FieldFilter.firstCondition()`.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,386 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ArcGIS
+import SwiftUI
+
+/// A view allowing the user to assembly a list of `FieldFilters` used to filter a list of features.
```
njarecha:
```suggestion
/// A view allowing the user to assemble a list of `FieldFilters` used to filter a list of features.
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,386 @@
...
+                    )
+                }
+                .pickerStyle(.menu)
+            }
+            .onAppear {
+                conditions = supportedOperators
+            }
+            
+            // Value
+            HStack {
```
dfeinzimer: Should the value field be hidden when a unary operator is selected?

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
njarecha:
```suggestion
/// A view allowing the user to assemble a list of `FieldFilters` used to filter a list of features.
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,386 @@
...
+            } icon: {
+                Image(systemName: "document.on.document")
+            }
+        }
+    }
+}
+
+/// A button to add a `FieldFilter` to the list of current `FieldFilters`.
+private struct AddButton: View {
+    /// Bool value specifying whether to draw the button with a border style.
```
njarecha: Boolean properties are documented this way in the SDK
```suggestion
    /// A Boolean value indicating whether to draw the button with a border style.
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,386 @@
...
+    let useBorderedStyle: Bool
+    
+    /// Creates an `AddButton`, alternately displaying it with a border style.
+    /// - Parameter useBorderedStyle: Bool value specifying whether to draw the button with a border style.
```
njarecha: same comment as above

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
njarecha:
```suggestion
    /// A Boolean value indicating whether to draw the button with a border style.
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
njarecha: Same as above

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,386 @@
...
+                    }
+                )
+                .multilineTextAlignment(.trailing)
+                .keyboardType(keyboardType)
+                .frame(alignment: .trailing)
+#if os(iOS)
+                .toolbar {
+                    ToolbarItemGroup(placement: .keyboard) {
+                        if UIDevice.current.userInterfaceIdiom == .phone, (fieldFilter.field.type?.isNumeric ?? false) {
+                            // Known SwiftUI issue: This button is known to sometimes not appear. (See Apollo #1159)
```
njarecha: Is mentioning `(See Apollo #1159)` necessary in public code? Can we do without it? I'm not sure what practices are used.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,386 @@
...
+                    }
+                }
+#endif
+            }
+        }
+        .id(fieldFilter.id)
+    }
+    
+    /// The button that allows a user to switch the numeric value between positive and negative.
+    var positiveNegativeButton: some View {
```
njarecha, line 333: Should this be named as `invertValueButton`?

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
njarecha: Is this equivalent?
```suggestion
    private var supportedOperators: [FilterOperator] {
        fieldFilter.field.type?.isNumeric == true
        ? FilterOperator.numericFilterOperators()
        : FilterOperator.textFilterOperators(fieldFilter.field.isNullable)
    }
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,386 @@
...
+extension FieldView {
+    /// The keyboard type to use depending on where the input is numeric and decimal.
+    var keyboardType: UIKeyboardType {
+        guard let fieldType = fieldFilter.field.type else { return .default }
+        
+        return if fieldType.isNumeric {
+#if os(visionOS)
+            // The 'positiveNegativeButton' doesn't show on visionOS
+            // so we need to show this keyboard so the user can type
+            // a negative number.
```
njarecha:
```suggestion
            // On visionOS, the `positiveNegativeButton` is not available.
            // Show the keyboard instead so the user can manually enter a negative number.
```

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,335 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ArcGIS
+import Foundation
+import SwiftUI
+
+@MainActor @Observable
```
njarecha, line 19: Is there any linter rule for specifying the property wrapper? In other place saw, `@MainActor @Observable final class CameraRequester`

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,335 @@
...
+class FilterViewModel {
+    /// The feature table containing the fields to filter on.
+    public var featureTable: ArcGISFeatureTable? {
```
njarecha, line 22: If setting the featureTable requires it to load, should it be an asynchronous property? or should it be a precondition that provided feature table a loaded one to get consistent and desired behavior?

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,335 @@
...
+    /// Initializes a filter view model.
+    /// - Parameter featureTable: The feature table for the filter.
+    init(featureTable: ArcGISFeatureTable? = nil) {
+        self.featureTable = featureTable
+    }
+    
+    /// Applies the current field filters.
+    func apply() {
+        filterViewIsPresented.toggle()
+       originalFieldFilters = fieldFilters
```
njarecha:
```suggestion
        filterViewIsPresented.toggle()
        originalFieldFilters = fieldFilters
```

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,335 @@
...
+    
+    /// The operation used specify how the `value` should be applied to the `field`.
+    var condition: FilterOperator = FilterOperator.equal
+    
+    /// The value to filter on.
+    var value = ""
+    
+    init(
+        field: Field,
+    ) {
```
njarecha:
```suggestion
    init(field: Field) {
```

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,335 @@
...
+        }
+    }
+    
+    /// Determines the appropriate set of filter operators for a field, depending on whether the field is numeric or textual.
+    /// It then returns the first available filter operator from that set. If no operators are found, it defaults to .equal.
+    /// - Returns: The first available filter operator from the appropriate set of filters.
+    private func firstCondition() -> FilterOperator {
+        let conditions = (field.type?.isNumeric ?? false) ? FilterOperator.numericFilterOperators() : FilterOperator.textFilterOperators(field.isNullable)
+        return conditions.first ?? FilterOperator.equal
+    }
```
njarecha, line 150:
```suggestion
    private func firstCondition() -> FilterOperator {
        let operators = field.type?.isNumeric == true
        ? FilterOperator.numericFilterOperators()
        : FilterOperator.textFilterOperators(field.isNullable)
        return operators.first ?? .equal
    }
```

njarecha (changes requested):
Some suggestions and comments. Can you please add instructions on how to test these changes?

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
mhdostal: Yup. It will be blank for `.isEmpty`, `.isNotEmpty`, `.isBlank`, and `.isNotBlank`.

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
mhdostal: I don't see any linter errors.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
mhdostal: It's called `positiveNegativeButton` is other parts of the feature form, so I'd prefer to leave it as-is.

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
mhdostal: I refactored this bit to eliminate code duplication with `supportedOperators`.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
mhdostal: Done.

mhdostal: Done.

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
mhdostal: And it's used on its own line for the `EmbeddedFeatureFormViewModel`. If you feel strongly, I can change it.

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
mhdostal: I think in the majority of cases the feature table will be loaded, as it will have to be loaded in order to have already displayed the features. The same goes for a stand-alone table. Because the `fields` property that is set once the table loads is observable, the UI should respond appropriately even if the table isn't loaded initially.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
mhdostal: Done.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
mhdostal: Agreed, we probably shouldn't mention internal issues in public code.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FormElements/UtilityAssociationsFormElement/UtilityAssociationFeatureCandidatesView.swift
```diff
@@ -153,34 +174,47 @@ extension FeatureFormView {
...
-                .disabled(!queryForFirstPageIsComplete)
-                .task(id: filterPhrase) {
-                    try? await Task.sleep(for: .milliseconds(500))
-                    queryForFilterPhraseIsRunning = true
-                    while !Task.isCancelled, filteredCandidates.isEmpty, let nextQueryParameters {
-                        queryFeatures(parameters: nextQueryParameters)
-                        await queryTask?.value
+                    .disabled(!queryForFirstPageIsComplete)
+                    .task(id: filterPhrase) {
+                        try? await Task.sleep(for: .milliseconds(500))
```
njarecha, line 199: Why is this needed? Anything based on delay may not behave the same on all devices and connection quality.

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,353 @@
...
+    static func numericFilterOperators() -> [FilterOperator] { [
+        .equal,
+        .notEqual,
+        .greaterThan,
+        .greaterThanOrEqual,
+        .lessThan,
+        .lessThanOrEqual
+    ] }
+    
+    var isUnary: Bool {
```
dfeinzimer, line 256:
```suggestion
    /// A Boolean value indicating whether the operator functions on a single operand.
    var isUnary: Bool {
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,373 @@
...
+                            }
+                        }
+                    }
+                }
+        }
+        .environment(model)
+        .background(Color(.systemGroupedBackground))
+    }
+    
+    /// Creates a `Button` used to delete a `FieldFilter`.
```
njarecha:
```suggestion
    /// Creates a `Button` that deletes the specified `FieldFilter`.
```

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,353 @@
...
+    func field(for name: String) -> Field? {
+        fields.first { field in
+            field.name == name
+        }
+    }
+}
+
+/// A class representing a single filter operation.
+@Observable
+class FieldFilter {
```
dfeinzimer, line 85: Just wondering if `condition`, `firstCondition()` and `supportedConditions` should be `operator`, `firstOperator` (computed property), and `suppportedOperators` respectively.

dfeinzimer (dismissed):
Looks good. Just leaving a few suggestions.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
njarecha:
```suggestion
    /// Creates a `Button` that deletes the specified `FieldFilter`.
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,373 @@
...
+                    bundle: .toolkitModule,
+                    comment: "A label for a button to delete a field filter."
+                )
+            } icon: {
+                Image(systemName: "trash")
+            }
+        }
+    }
+    
+    /// Creates a `Button` used to duplicate a `FieldFilter`.
```
njarecha:
```suggestion
    /// Creates a `Button` that duplicates the specified `FieldFilter`.
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,373 @@
...
+                    comment: "A label for a button to duplicate a field filter."
+                )
+            } icon: {
+                Image(systemName: "document.on.document")
+            }
+        }
+    }
+}
+
+/// A button to add a `FieldFilter` to the list of current `FieldFilters`.
```
njarecha:
```suggestion
/// A button that adds a `FieldFilter` to the current `FieldFilters`.
```

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -0,0 +1,373 @@
...
+    /// The model used by the view.
+    @Environment(FilterViewModel.self) private var model
+    
+    /// The `FieldFilter` represented by the view.
+    @State private var fieldFilter: FieldFilter
+    
+    /// The list of conditions/operations the user is allowed to choose from.
+    @State private var conditions = [FilterOperator]()
+    
+    @State private var selectedFieldName = ""
```
njarecha, line 244: Please add doc

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,353 @@
...
+        originalFieldFilters = fieldFilters
+    }
+    
+    /// Cancels the current changes to the field filters.
+    func cancel() {
+        fieldFilters = originalFieldFilters
+        filterViewIsPresented.toggle()
+    }
+    
+    func field(for name: String) -> Field? {
```
njarecha: We use this in the SDK code, `func field(named fieldName: String) -> Field?`

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,353 @@
...
+    var field: Field {
+        didSet {
+            if oldValue !== field {
+                // We're changing the field, so reset the condition
+                condition = firstCondition()
+            }
+        }
+    }
+    
+    /// The operation used specify how the `value` should be applied to the `field`.
```
njarecha:
```suggestion
    /// The operation used to specify how the value should be applied to the field.
```

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,353 @@
...
+        case .equal, .notEqual, .isOp, .isNot, .greaterThan, .greaterThanOrEqual, .lessThan, .lessThanOrEqual:
+            "\(field.name) \(condition.sqlOperator) \(value)"
+        }
+    }
+    
+    /// Determines the appropriate set of filter operators for a field, depending on whether the field is numeric or textual.
+    /// It then returns the first available filter operator from that set. If no operators are found, it defaults to .equal.
+    /// - Returns: The first available filter operator from the appropriate set of filters.
+    private func firstCondition() -> FilterOperator {
+        return supportedConditions.first ?? FilterOperator.equal
```
njarecha:
```suggestion
        return supportedConditions.first ?? .equal
```

njarecha (changes requested):
Some more suggestions

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -0,0 +1,353 @@
...
+    }
+    
+    /// The operators supported for the given `FieldFilter` field type.
+    /// - Returns: A list of operators appropriate for the given `FieldFilter` field type.
+    var supportedConditions: [FilterOperator] {
+        field.type?.isNumeric == true
+        ? FilterOperator.numericFilterOperators()
+        : FilterOperator.textFilterOperators(field.isNullable)
```
dfeinzimer: Numeric fields can be nullable too. I know the Kotlin design didn't have this, but should `numericFilterOperators()` have a `fieldIsNullable` parameter as well to support using `isBlank` and `isNotBlank` too?

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FormElements/UtilityAssociationsFormElement/UtilityAssociationFeatureCandidatesView.swift
dfeinzimer: I added this in https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/1334, looks like it's just getting moved in this PR. IIRC, this functions as a debounce when the filter phrase changes as a result of the user typing, before the next query starts.

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
mhdostal: I wanted to do that originally, but `operator` is a keyword in Swift:

`Keyword 'operator' cannot be used as an identifier here`

njarecha (approved):

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
dfeinzimer: Ah yes. If you use backticks, you can get around that. Like so:

```swift
/// The operator used to specify how the `value` should be applied to the `field`.
var `operator`: FilterOperator = FilterOperator.equal
```

The same can be done with `FilterOperator.isOp`:
```swift
case `is` = "is"
```

Up to you...

dfeinzimer (approved):

mhdostal:
@njarecha @dfeinzimer @philium Thank you all for your feedback!!

PR #1366, opened 2026-01-28
`PopupView` - edit summary

des12437:
Related to `swift/7692`.

Integrates the `Popup.editSummary` property into the PopupView.

- Refactors EmbeddedPopupView popupElements list to use ForEach so the edit summary text can be added to the bottom of the List
- Adds optional Content closure for editSummary view
- Passes `nil` to closure when the edit summary text is empty

edit summary
Popup edit summary

### DIFF
--- a/Sources/ArcGISToolkit/Components/Popups/EmbeddedPopupView.swift
+++ b/Sources/ArcGISToolkit/Components/Popups/EmbeddedPopupView.swift
@@ -31,8 +31,10 @@ struct EmbeddedPopupView: View {
         VStack(spacing: 0) {
             switch evaluationResult {
             case .success(let evaluatedElements):
-                PopupElementList(popupElements: evaluatedElements)
-                    .environment(\.popupTitle, popup.title)
+                PopupElementList(popupElements: evaluatedElements) {
+                    !popup.editSummary.isEmpty ? editSummary : nil
+                }
+                .environment(\.popupTitle, popup.title)
             case .failure(let error):
                 Text(
                     "Popup evaluation failed: \(error.localizedDescription)",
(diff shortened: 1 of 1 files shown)


Sources/ArcGISToolkit/Components/Popups/EmbeddedPopupView.swift
```diff
@@ -31,8 +31,10 @@ struct EmbeddedPopupView: View {
         VStack(spacing: 0) {
             switch evaluationResult {
             case .success(let evaluatedElements):
-                PopupElementList(popupElements: evaluatedElements)
-                    .environment(\.popupTitle, popup.title)
+                PopupElementList(popupElements: evaluatedElements) {
+                    !popup.editSummary.isEmpty ? editSummary : nil
+                }
+                .environment(\.popupTitle, popup.title)
```
dfeinzimer: Since the summary isn't a popup element, and this is already in a `VStack`, would a preferable alternative be to simply place the summary here in the success condition? E.g.:

```swift
case .success(let evaluatedElements):
    PopupElementList(popupElements: evaluatedElements)
        .environment(\.popupTitle, popup.title)
    if !popup.editSummary.isEmpty {
        editSummary
    }
```

dfeinzimer: I guess it depends on whether we want it always visible and stuck to the bottom of the view or to the bottom of the last element in the list. Do we know what other platforms do?

des12437: The edit summary should be stuck to the bottom of the last element in the list.

dfeinzimer (dismissed):

Sources/ArcGISToolkit/Components/Popups/EmbeddedPopupView.swift
```diff
@@ -90,31 +92,44 @@ struct EmbeddedPopupView: View {
...
+    }
 }
 
 extension EmbeddedPopupView {
-    private struct PopupElementList: View {
+    private struct PopupElementList<Content>: View where Content: View {
         let popupElements: [PopupElement]
         
+        let editSummary: () -> Content?
+
```
CalebRas: In my opinion, content closures should be used when there are different types of Views that need to be passed down. In this case, the content will always be the same (optional `Text`). It also separates the `editSummary` View definition from its use point, which makes it less readable. Is there a reason we can't/shouldn't do something like:

```swift
private struct PopupElementList: View {
    let popupElements: [PopupElement]
    let editSummary: String
    
    var body: some View {
        List {
            ForEach(popupElements) { popupElement in
                ...
            }
            
            if !editSummary.isEmpty {
                Text(editSummary)
                    .font(.footnote)
                    .foregroundStyle(.secondary)
                    .listSectionSeparator(.hidden, edges: .top)
            }
        }
        .listStyle(.plain)
    }
}
```

Sources/ArcGISToolkit/Components/Popups/EmbeddedPopupView.swift
```diff
@@ -90,31 +92,44 @@ struct EmbeddedPopupView: View {
             return popup.evaluatedElements
         }
     }
+    
+    /// The edit summary of the popup.
+    private var editSummary: some View {
+        Text(popup.editSummary)
```
CalebRas: Is there any data I can use to test this?

Sources/ArcGISToolkit/Components/Popups/EmbeddedPopupView.swift
des12437: This is a great suggestion, this feedback is now applied.

dfeinzimer (approved):

CalebRas (approved):
Looks great!

des12437:
Thank you both! @dfeinzimer @CalebRas

PR #1368, opened 2026-02-04
`FeatureFormView` - Remove "Add Association" menu

dfeinzimer:
Apollo 1599

Previously, we planned to support adding associations via map interaction. This is no longer planned, so the intermediate menu is no longer needed.


mhdostal (approved):
We should make sure the test design is updated as well.

yo1995 (approved):

dfeinzimer:
Thank you both for the quick turnaround! 🚀

PR #1369, opened 2026-02-07
`FilterView` - Support `date` and `dateOnly` field types

dfeinzimer:
Apollo 1587

### DIFF
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
@@ -150,15 +150,25 @@ class FieldFilter {
     func query() -> String {
         switch condition {
         case .startsWith:
-            "\(field.name) \(condition.sqlOperator) '\(value)%'"
+            return "\(field.name) \(condition.sqlOperator) '\(value)%'"
         case .endsWith:
-            "\(field.name) \(condition.sqlOperator) '%\(value)'"
+            return "\(field.name) \(condition.sqlOperator) '%\(value)'"
         case .contains, .doesNotContain:
-            "\(field.name) \(condition.sqlOperator) '%\(value)%'"
+            return "\(field.name) \(condition.sqlOperator) '%\(value)%'"
         case .isBlank, .isNotBlank, .isEmpty, .isNotEmpty:
-            "\(field.name) \(condition.sqlOperator)"
+            return "\(field.name) \(condition.sqlOperator)"
         case .equal, .notEqual, .isOp, .isNot, .greaterThan, .greaterThanOrEqual, .lessThan, .lessThanOrEqual:
-            "\(field.name) \(condition.sqlOperator) \(value)"
+            let formattedValue = switch field.type {
+            case .date:
+                "timestamp '\(value)'"
+            case .dateOnly:
+                "date '\(value)'"
+            case .text:
+                "'\(value)'"
+            default:
+                value
+            }
+            return "\(field.name) \(condition.sqlOperator) \(formattedValue)"
         }
     }
     
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
@@ -113,11 +113,9 @@ class FieldFilter {
     var dateValue: Date {
         didSet {
             guard field.type == .date || field.type == .dateOnly else { return }
-            let dateFormatter = DateFormatter()
-            dateFormatter.locale = Locale(identifier: "en_US")
-            dateFormatter.setLocalizedDateFormatFromTemplate(
-                field.type == .dateOnly ? "MMM dd, yyyy" : "MMM dd, yyyy h:mm a"
-            )
+            let dateFormatter = ISO8601DateFormatter()
+            dateFormatter.timeZone = TimeZone.current
+            dateFormatter.formatOptions = field.type == .dateOnly ? [.withFullDate] : [.withFullDate, .withFullTime]
             value = dateFormatter.string(from: dateValue)
         }
     }
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
@@ -182,8 +182,10 @@ extension FilterViewModel {
     private func supportedFields(_ allFields: [Field]) -> [Field] {
         allFields.filter { field in
             (field.type?.isNumeric ?? false) ||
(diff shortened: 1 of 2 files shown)


Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -368,7 +377,12 @@ private struct FieldView: View {
     /// Determines the conditions to display for the given `FieldFilter` field type.
     /// - Returns: A list of conditions appropriate for the given `FieldFilter` field type.
     private func fieldConditions() -> [FilterOperator] {
-        (fieldFilter.field.type?.isNumeric ?? false) ? FilterOperator.numericFilterOperators() : FilterOperator.textFilterOperators(fieldFilter.field.isNullable)
+        let type = fieldFilter.field.type
+        if type == .date || type == .dateOnly {
+            return FilterOperator.numericFilterOperators() + [.isBlank, .isNotBlank]
```
mhdostal: I refactored this bit, based on your suggestion in the original PR, but shouldn't `.isBlank` and `.isNotBlank` only be added for dates if the field is nullable? That's what Tony seems to suggest [here](https://zentopia.esri.com/workspaces/apollo-646b9f4357b0630019bb6f9e/issues/gh/runtime/apollo/1607#issuecomment-6082591)

dfeinzimer: Yes fixed, thanks!

pgruenler (approved):
Looks good.

mhdostal (approved):

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
```diff
@@ -324,34 +324,43 @@ private struct FieldView: View {
...
 #if os(iOS)
-                    .toolbar {
-                        ToolbarItemGroup(placement: .keyboard) {
-                            if UIDevice.current.userInterfaceIdiom == .phone, (fieldFilter.field.type?.isNumeric ?? false) {
-                                // Known SwiftUI issue: This button is known to sometimes not appear.
-                                positiveNegativeButton
-                                Spacer()
+                        .toolbar {
+                            ToolbarItemGroup(placement: .keyboard) {
+                                if UIDevice.current.userInterfaceIdiom == .phone, (fieldFilter.field.type?.isNumeric ?? false) {
```
philium, line 355: There are often two things often overlooked when performing a check like this:
1. Split screen on iPad
2. Landscape on an iPhone Plus (which is a regular horizontal size class, more like an iPad)

Should size classes be checked here instead? If so, we have a [convenience](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/067de0b6297f4838cc127a5bd064ad6b207ee4b9/Sources/ArcGISToolkit/Extensions/SwiftUI/EnvironmentValues.swift#L21-L25) for that.

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
```diff
@@ -105,18 +105,34 @@ class FieldFilter {
...
     var condition: FilterOperator = FilterOperator.equal
     
+    /// A date value for date pickers to bind to.
+    var dateValue: Date {
+        didSet {
+            guard field.type == .date || field.type == .dateOnly else { return }
+            let dateFormatter = ISO8601DateFormatter()
+            dateFormatter.timeZone = TimeZone.current
+            dateFormatter.formatOptions = field.type == .dateOnly ? [.withFullDate] : [.withFullDate, .withFullTime]
+            value = dateFormatter.string(from: dateValue)
```
philium, line 125: Two questions:

1. Could [`Date.ISO8601FormatStyle `](https://developer.apple.com/documentation/foundation/date/iso8601formatstyle) not be used here? Apple's [recommendation](https://developer.apple.com/documentation/foundation/data-formatting) is to only use format styles in Swift.
2. Could `value` be computed on-demand (i.e. be a computed property)? That is always better for formatted data, for multiple reasons.

philium (commented):
I happened to see this PR come in. Here are a few things I noticed.

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
njarecha: FYI. The `DateOnly`'s [description](https://devtopia.esri.com/runtime/swift/blob/main/ArcGIS/Sources/ArcGIS/Common/DateOnly.swift#L33) already returns ISO 8601 string.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FilterView.swift
dfeinzimer: This code was pre-existing and just re-indented as a result of other changes in this PR. Nonetheless, the intention of this check is to discern between iOS and iPadOS because regardless of window size in iPadOS, the system gives you a full keyboard and ignores keyboard toolbar items. This is why we checked interface idiom and not size class. I suppose we could also just remove the check altogether.

Simulator Screenshot - swift-test-ipad (26 1) - 2026-02-23 at 11 55 20

Sources/ArcGISToolkit/Components/FeatureFormView/FilterViewModel.swift
dfeinzimer: > Could [Date.ISO8601FormatStyle ](https://developer.apple.com/documentation/foundation/date/iso8601formatstyle) not be used here?

Done in https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/1375/changes/547fd403b130c9ba080fc4fde2da01206240e148.

> Could value be computed on-demand (i.e. be a computed property)?

For text type fields, the `TextField` needs a string property to bind to.

philium: Did you try using [this](https://developer.apple.com/documentation/swiftui/textfield/init(value:format:prompt:label:)) initializer, using the `dateValue` as the text field's value instead?

dfeinzimer: Hmm not sure I follow. The `TextField` is bound to `FieldFilter.value: String` for text type fields and the `DatePicker` is bound to `FieldFilter.dateValue: Date` for date and date+time type fields. In #1370 we're set to add `FieldFilter.codedValue: CodedValue?` to bind a `Picker` to for numeric type fields with coded value domains.

For simplicity sake, all values are written back to `FieldFilter.value` so that `FieldFilter.query()` has to only look at one property when building the query.

PR #1377, opened 2026-02-24
Feature Form Add Attachments from File

mhdostal:
This adds the ability to add attachments directly from a file, without loading the file's data into memory. It works for files from the Photo Library or loaded directly from a file on the device.

Apollo #1590 as part of the "large attachment" work.

Note: this is dependent on the Swift API Audit #7823

### DIFF
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentImportMenu.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentImportMenu.swift
@@ -174,7 +174,6 @@ struct AttachmentImportMenu: View {
                     contentType: newAttachmentImportData.contentType,
                     fileURL: url
                 )
-
             } else if let data = newAttachmentImportData.data {
                 newAttachment = element.addAttachment(
                     name: fileName,
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentImportMenu.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentImportMenu.swift
@@ -182,7 +182,7 @@ struct AttachmentImportMenu: View {
                     data: data
                 )
             }
-
+            
             guard let newAttachment else {
                 importState = .errored(.creationFailed)
                 return
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentImportMenu.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentImportMenu.swift
@@ -141,17 +141,19 @@ struct AttachmentImportMenu: View {
         .task(id: importState) {
             guard case let .finalizing(newAttachmentImportData) = importState else { return }
             
-            let attachmentSize = Measurement(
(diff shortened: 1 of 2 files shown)


Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentImportMenu.swift
```diff
@@ -164,11 +166,24 @@ struct AttachmentImportMenu: View {
...
-            ) else {
+            
+            var newAttachment: FeatureAttachment? = nil
+            if let url = newAttachmentImportData.filePath {
+                newAttachment = try? await element.addAttachment(
+                    named: fileName,
+                    contentType: newAttachmentImportData.contentType,
+                    fileURL: url
+                )
+
```
dfeinzimer:
```suggestion
```

Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentImportMenu.swift
```diff
@@ -164,11 +166,24 @@ struct AttachmentImportMenu: View {
...
+            } else if let data = newAttachmentImportData.data {
+                newAttachment = element.addAttachment(
+                    name: fileName,
+                    contentType: newAttachmentImportData.contentType,
+                    data: data
+                )
+            }
+
```
dfeinzimer:
```suggestion
            
```

Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentImportMenu.swift
```diff
@@ -164,11 +166,24 @@ struct AttachmentImportMenu: View {
...
             }
-            guard let newAttachment = element.addAttachment(
-                name: fileName,
-                contentType: newAttachmentImportData.contentType,
-                data: newAttachmentImportData.data
-            ) else {
+            
+            var newAttachment: FeatureAttachment? = nil
+            if let url = newAttachmentImportData.filePath {
+                newAttachment = try? await element.addAttachment(
```
dfeinzimer, line 172: Should we catch any errors here and report them? Possibly through `importState`?

mhdostal: At the end of this `if-else` block there is this:

```
            guard let newAttachment else {
                importState = .errored(.creationFailed)
                return
            }
```

I've asked if there is a list of potential errors that can be thrown, instead of always just returning `.creationFailed`.

mhdostal: From the Core team:

_commonFileNotFound if the file doesn't exist, commonNoData if the file is empty or SQLiteTooBig if the file is larger than 999 MB._

I'm fine keeping the error as `.creationFailed`, but adding more fine-grained errors if warranted.

dfeinzimer: Okay, it seems like we're likely to catch either of those conditions earlier on anyway so we should be all set.

dfeinzimer (approved):

des12437 (approved):

PR #1378, opened 2026-02-26
`FeatureFormView` - Revise asset type row identifiers

dfeinzimer:
Apollo 1638

### DIFF
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FormElements/UtilityAssociationsFormElement/UtilityAssociationAssetTypesView.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FormElements/UtilityAssociationsFormElement/UtilityAssociationAssetTypesView.swift
@@ -56,7 +56,7 @@ extension FeatureFormView {
                     )
                 }
                 Section {
-                    ForEach(filteredTypes, id: \.code) { assetType in
+                    ForEach(filteredTypes, id: \.qualifiedID) { assetType in
                         NavigationLink(
                             assetType.name,
                             value: FeatureFormView.NavigationPathItem.utilityAssociationFeatureCandidatesView(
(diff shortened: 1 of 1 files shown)


yo1995 (dismissed):
Changes are good. 1 suggestion, and 2 things to consider if the suggestion isn't taken

- Maybe consult/confirm with Tony on whether the group code AND type code ensures utility asset uniqueness
- Does the order matter? If there is a precedence in Pro for which code goes before the other, you can follow that. (me personally feel  group code -> type code is more natural, tree-like)

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FormElements/UtilityAssociationsFormElement/UtilityAssociationAssetTypesView.swift
```diff
@@ -86,3 +86,10 @@ extension FeatureFormView {
         }
     }
 }
+
+private extension UtilityAssetType {
+    /// An ID for the type composed of its code and group's code, if present.
+    var qualifiedID: String {
+        [code, group?.code].compactMap { "\($0)" }.joined(separator: "-")
```
yo1995: Maybe consider using the standard hasher

```swift
let hasher = Hasher()
hasher.combine(code)
if let groupCode = group?.code {
    hasher.combine(groupCode)
}
return hasher.finalize()
```

Benefits:

- The returned value is an `Int`, which has smaller memory footprint than the String
- I assume hash is faster to compute? Might depend on Swift implementation, but hashing algorithms mostly happen in hardware nowadays

mhdostal (dismissed):

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FormElements/UtilityAssociationsFormElement/UtilityAssociationAssetTypesView.swift
dfeinzimer: Nice idea

dfeinzimer:
> * Maybe consult/confirm with Tony on whether the group code AND type code ensures utility asset uniqueness

Confirmed

> * Does the order matter? If there is a precedence in Pro for which code goes before the other, you can follow that. (me personally feel  group code -> type code is more natural, tree-like)

Apollo 1635

yo1995 (approved):

mhdostal (approved):

PR #1381, opened 2026-03-10
`AttachmentsFeatureElementView` - Fix potential load cancellation condition

dfeinzimer:
Apollo 1626

Additional information in issue.

PR #1379, opened 2026-02-27
`FeatureFormView` - General refactoring

dfeinzimer:
- Refactors `RangeDomain.displayableNumericMinAndMax` as [suggested](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/1176#discussion_r2183926569) by @yo1995.
- Adds a common `Button.delete` to use across various components.
- Marks a few view model classes as `final`.
- Overhauls `FeatureFormView` UI tests.
  - Uses recently added `XCUIElement` methods and adds a few more.
  - Restores _nearly all_ cases to passing on all supported platforms (save for a couple expected failures).
  - Synchornizes variable naming patterns across all cases.
- Various small improvements.

PR #1382, opened 2026-03-11
`FeatureFormView` - Various bug fixes

dfeinzimer:
- `AttachmentsFeatureElementView`:
  - Adds a test case for #1381, using new members added in #1379.
  - Adds a fallback message for when the element contains no attachments. (Apollo 1626)
- Fixes a class of bugs introduced in #1283:
  - `UtilityAssociationsFormElementView` - Prevent the view's model from being recreated each time the view is scrolled back into visibility. (Apollo 1657)
  - `SwitchInput` - Prevent switches from giving themselves focus each time the view is scrolled back into visibility. (Apollo 1657)
  - `GroupFormElementView` - Prevents groups from rechecking their configured intial collapse state each time the view is scrolled back into visibility. (Apollo 1663)


Sources/ArcGISToolkit/Common/AttachmentsFeatureElementView.swift
```diff
@@ -65,7 +65,16 @@ struct AttachmentsFeatureElementView: View {
...
+                    if models.isEmpty {
+                        Text(
+                            "No attachments",
+                            bundle: .toolkitModule,
+                            comment: """
+                                A label indicating an element
+                                contains no file attachments.
+                                """
+                        )
+                    } else {
```
dfeinzimer, line 77: Adding a view for the empty state prevents the enclosing `Group` from being optimized out, causing the `onAttachmentIsEditableChange` modifier on it to never run.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FormElements/UtilityAssociationsFormElement/UtilityAssociationsFormElementView.swift
```diff
@@ -56,10 +56,12 @@ extension FeatureFormView {
...
-                self.embeddedFeatureFormViewModel.associationsFilterResultsModels[element] = .init(
-                    element: element,
-                    includeEmptyFilterResults: true
-                )
+                if associationsFilterResultsModel == nil {
+                    self.embeddedFeatureFormViewModel.associationsFilterResultsModels[element] = .init(
+                        element: element,
+                        includeEmptyFilterResults: true
+                    )
+                }
```
CalebRas: Nitpick, but maybe a guard here instead?

```suggestion
                guard associationsFilterResultsModel == nil else { return }
                embeddedFeatureFormViewModel.associationsFilterResultsModels[element] = .init(
                    element: element,
                    includeEmptyFilterResults: true
                )
```

CalebRas (dismissed):
Looks good!

mhdostal (dismissed):

CalebRas (approved):

mhdostal (approved):

PR #1383, opened 2026-03-13
Add `BuildingExplorer`

zkline101:
Home-one #4068

I'll send a webscene to the reviewers so you can test the component with a scene that contains multiple BSLs.

A following PR will be made for a step-by-step tutorial and potentially some automated tests.

iOS:
Simulator Screenshot - iPad Pro 13-inch (M5) - 2026-03-18 at 12 37 34

Mac Catalyst:
Screenshot 2026-03-18 at 12 15 49 PM

visionOS:
Screenshot 2026-03-18 at 12 16 45 PM


rolson:
Glad to see this in code review. Can you please add some screen shots for each platform (iOS, macOS, visionOS) to the description? Thanks

zkline101:
Sure, happy to.

zkline101:
I added the screenshots.

rolson:
- Not a big fan of the look and feel of that particular drop-down. Can we look at other options?
- How many construction phases are there? In this case 7, is it always 7?
- Is the zoom to button standard and follow the Hig for buttons? It looks almost like data. Could be a button in the nav bar (ie. leading)
- The common design uses terminology "show overview" instead of "visible", do we need to sync up on that?
- The common design has levels and construction phases part of the top section, do we need to sync up on that?

zkline101:
> Not a big fan of the look and feel of that particular drop-down. Can we look at other options?

Which part don't you like? Are you saying I should explore using a `List` instead of a `Form`?

> How many construction phases are there? In this case 7, is it always 7?

No there could be more. I personally haven't seen more then like 10. But since there is no limit, I have code in the implementation to look out for this. If there is more than 10 phases then we switch to a regular picker and not a segmented one since it could get crowded.

> Is the zoom to button standard and follow the Hig for buttons? It looks almost like data. Could be a button in the nav bar (ie. leading)

Let me look into this. I was following the common design here but I could see why we may want to differ.

> The common design uses terminology "show overview" instead of "visible", do we need to sync up on that?

Two things here:
- I'll be making changes in the common design to change "Show overview" to "Show Full Model" for a few reasons.
- "Show overview" and "visible" are two separate things. "Show overview" is to toggle between the full model and overview (shell of the building) and "visible" is to toggle the visibility of the entire building scene layer. The design will also be updated to add this visibility button.

> The common design has levels and construction phases part of the top section, do we need to sync up on that?

I think the difference you are noticing is the levels and phases are in their own section here where that doesn't look the same in the common design. I was thinking that was a nice change we could add in the Swift Toolkit to separate things and make it more organized. But I should add this to the common design too. I'm not sure what this looks like for others.

rolson:
> Which part don't you like?

The font is small for a top level element and there is a lot of space between that and the first item.

> Are you saying I should explore using a List instead of a Form?

No, nothing to do with that.

>  think the difference you are noticing is the levels and phases are in their own section here where that doesn't look the same in the common design

But the breakdown of the sections isn't obvious to me as to why they are in new sections. It seems that it was more of a way to get a header for a segmented control, and the extra sections seems potentially unnecessary. Unless everything is in it's own section, which we probably don't want.

rolson:
I think sections should have natural breakdowns/categories. If not then maybe there is only one section. Here is an example of sections in an analysis settings view:

image

zkline101:
> The font is small for a top level element and there is a lot of space between that and the first item.

Oh ok I see. I tried changing the font before and SwiftUI wouldn't make the change. I'll look into this again. 
Regarding the spacing, let me see if there is a way to adjust this.

> I think sections should have natural breakdowns/categories.

Hm I see your point. Let me think about this and play around with a few options.

rolson:
There should be some standard options to make things look good. We don't need to do something strange like negative padding... The reason I asked for options is because we should have a target design for inspiration to copy.

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
```diff
@@ -0,0 +1,176 @@
...
+///
+/// ![An image of the BuildingExplorer component](BuildingExplorer)
+///
+/// This component provides a way for users to browse the levels, phases, and
+/// sublayers of a building scene layer. The component can highlight specified levels or phases
+/// and show or hide building features of different categories and subcategories.
+///
+/// **Features**
+/// - Support for multiple building scene layers within a scene.
+/// - Visiblity of the building scene layer can be toggled on and off.
```
dfeinzimer:
```suggestion
/// - Visibility of the building scene layer can be toggled on and off.
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
```diff
@@ -0,0 +1,176 @@
...
+/// - Visbility of the full model can be toggled on and off. If the full model is turned off, the
```
dfeinzimer:
```suggestion
/// - Visibility of the full model can be toggled on and off. If the full model is turned off, the
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
```diff
@@ -0,0 +1,176 @@
...
+/// overview (shell of the building) will show. If the full model is turned on, the full model wil
```
dfeinzimer:
```suggestion
/// overview (shell of the building) will show. If the full model is turned on, the full model will
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
```diff
@@ -0,0 +1,176 @@
...
+/// show and the overview will be turned off.
+/// - A button to zoom into the building.
+/// - Selecting a level of the building to highlight in the building scene layer.
+///     - The selected level and all of the features of the level are rendered normally using
+///     the solid filter mode.
+///     - Levels above are hidden.
+///     - Levels below are rendered using a xray filter mode.
```
dfeinzimer:
```suggestion
///     - Levels below are rendered using a X-ray filter mode.
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
```diff
@@ -0,0 +1,176 @@
...
+/// wasn't set by the explorer then it will be ignored.
+/// - Since: 300.0
+public struct BuildingExplorer: View {
+    /// The scene being used to create items for the explorer.
+    private let scene: ArcGIS.Scene
+    /// The proxy needed to set the viewpoint when pressing the
+    /// "Zoom to building" button.
+    private let localSceneViewProxy: LocalSceneViewProxy?
+    
+    /// An error that occured while setting up the explorer.
```
dfeinzimer:
```suggestion
    /// An error that occurred while setting up the explorer.
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
```diff
@@ -0,0 +1,176 @@
...
+    @Binding private var selection: BuildingExplorerItem?
+    
+    /// Creates the building explorer view.
+    /// - Parameters:
+    ///   - scene: The scene which contains the building scene layers that will
+    ///   be used in the explorer.
+    ///   - items: The items being shown in the building explorer.
+    ///   - selection: The selected item in the building explorer.
+    ///   - localSceneViewProxy: The proxy to provide access to the set viewpoint
+    ///   operation for the "Zoom to buildilng" button. If the proxy is `nil` then that button
```
dfeinzimer:
```suggestion
    ///   operation for the "Zoom to building" button. If the proxy is `nil` then that button
```

Examples/Examples/BuildingExplorerExampleView.swift
```diff
@@ -0,0 +1,47 @@
...
+                                localSceneViewProxy: proxy
+                            )
+                            .frame(idealWidth: 400, idealHeight: 500)
+                            .presentationCompactAdaptation(.popover)
+                        }
+                    }
+                }
+        }
+    }
+}
```
dfeinzimer, line 51: Consider adding a preview.


```suggestion
}

#Preview {
    NavigationStack {
        BuildingExplorerExampleView()
    }
}
```

Sources/ArcGISToolkit/Documentation.docc/Resources/BuildingExplorer/BuildingExplorer.png
dfeinzimer, line 1: We should provide a light mode variant of this image.

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
```diff
@@ -0,0 +1,176 @@
...
+/// sublayers of a building scene layer. The component can highlight specified levels or phases
+/// and show or hide building features of different categories and subcategories.
+///
+/// **Features**
+/// - Support for multiple building scene layers within a scene.
+/// - Visiblity of the building scene layer can be toggled on and off.
+/// - Visbility of the full model can be toggled on and off. If the full model is turned off, the
+/// overview (shell of the building) will show. If the full model is turned on, the full model wil
+/// show and the overview will be turned off.
+/// - A button to zoom into the building.
```
dfeinzimer: "zoom to"?

```suggestion
/// - A button to zoom to the building.
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
```diff
@@ -0,0 +1,176 @@
...
+///     - The selected phase is rendered normally using the solid filter mode.
+/// - Visibility of building feature categories and subcategories can be toggled on and off.
+///
+/// To see it in action, try out the [Examples](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/tree/main/Examples/Examples)
+/// and refer to [BuildingExplorerExampleView.swift](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/main/Examples/Examples/BuildingExplorerExampleView.swift)
+/// in the project.
+///
+/// - Note: The building explorer only supports local scenes and the scene must contain at least
+/// one building scene layer. If there is an active filter on the building scene layer that
+/// wasn't set by the explorer then it will be ignored.
```
dfeinzimer:
```suggestion
/// wasn't set by the explorer, then it will be ignored.
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
```diff
@@ -0,0 +1,176 @@
...
+/// - Since: 300.0
+public struct BuildingExplorer: View {
+    /// The scene being used to create items for the explorer.
+    private let scene: ArcGIS.Scene
+    /// The proxy needed to set the viewpoint when pressing the
+    /// "Zoom to building" button.
+    private let localSceneViewProxy: LocalSceneViewProxy?
+    
+    /// An error that occured while setting up the explorer.
+    @State private var setupError: (any Error)?
```
dfeinzimer: Since "set up" is being used as a verb (two words), I think `setUpError` would be more correct.

```suggestion
    @State private var setUpError: (any Error)?
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
```diff
@@ -0,0 +1,176 @@
...
+    /// A Boolean value indicating if the explorer finished setting up.
+    @State private var setupIsDone = false
```
dfeinzimer: Since "set up" is being used as a verb (two words), I think `setUpIsDone` would be more correct.


```suggestion
    @State private var setUpIsDone = false
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
```diff
@@ -0,0 +1,176 @@
...
+                ProgressView()
+            }
+        }
+        .task {
+            await setup()
+        }
+    }
+    
+    /// Sets up the building explorer.
+    private func setup() async {
```
dfeinzimer: Since "set up" is being used as a verb (two words), I think `setUp` would be more correct.

```suggestion
    private func setUp() async {
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
```diff
@@ -0,0 +1,176 @@
...
+            setupIsDone = true
+        } catch {
+            setupError = error
+        }
+    }
+}
+
+private extension BuildingExplorer {
+    /// An error that can occur when setting up the explorer.
+    enum SetupError: LocalizedError {
```
dfeinzimer: Since "set up" is being used as a verb (two words), I think `SetUpError` would be more correct.


```suggestion
    enum SetUpError: LocalizedError {
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
```diff
@@ -0,0 +1,176 @@
...
+        /// Globe scenes are not supported in the Building Explorer. Only local scenes
```
dfeinzimer:
```suggestion
        /// Global scenes are not supported in the Building Explorer. Only local scenes
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
```diff
@@ -0,0 +1,176 @@
...
+        case noBuildingSceneLayers
+        
+        var errorDescription: String? {
+            switch self {
+            case .globalScenesNotSupported:
+                "Global scenes aren't supported in the Building Explorer."
+            case .noBuildingSceneLayers:
+                "No building scene layers in the scene."
+            }
+        }
```
dfeinzimer, line 180: Since these can be shown to the user and we own the errors we need to generate an entry in the strings files for translation, so we should use `String(localized:bundle:comment:)` or similar. See `LoadImageError` and `CertificatePickerViewModel.CertificateError` as examples.

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
```diff
@@ -0,0 +1,411 @@
...
+    /// discipline and categories.
+    @State private var groupSublayers: [BuildingGroupSublayer] = []
+    
+    /// A Boolean value indicating if the camera should change to zoom into
+    /// the building.
+    @State private var shouldZoomToBuilding = false
+    
+    @State private var layerIsVisible = false
+    
+    // Layer picker properties.
```
dfeinzimer: These should use the MARK syntax, otherwise, they don't show up in the symbol navigator.

```suggestion
    // MARK: Layer picker properties.
```

Image

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
```diff
@@ -0,0 +1,411 @@
...
+    
+    /// The name of the selected layer in the layer picker.
+    @State private var selectedLayerName = ""
+    /// The names of the layers that can be selected in the picker.
+    @State private var layerNames: [String] = []
+    
+    // Level picker properties.
```
dfeinzimer:
```suggestion
    // MARK: Level picker properties.
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
```diff
@@ -0,0 +1,411 @@
...
+    
+    /// The selected level in the level picker.
+    @State private var selectedLevel: String = ""
+    /// The available levels in the level picker.
+    @State private var availableLevels: [String] = []
+    
+    // Phase picker properties.
```
dfeinzimer:
```suggestion
    // MARK: Phase picker properties.
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
```diff
@@ -0,0 +1,411 @@
...
+    
+    /// The selected phase in the phase picker.
+    @State private var selectedPhase = ""
+    /// The available phase in the phase picker.
+    @State private var availablePhases: [String] = []
+    
+    @State private var phasePickerStyle: (any PickerStyle) = .automatic
+    
+    // Full model and sublayer toggle properties.
```
dfeinzimer:
```suggestion
    // MARK: Full model and sublayer toggle properties.
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
```diff
@@ -0,0 +1,411 @@
...
+                    }
+                }
+                
+                // Only show options to filter the full model
+                // if the full model is being shown and the BSL
+                // is visible.
+                if showFullModel && layerIsVisible {
+                    if !availableLevels.isEmpty {
+                        Section("Levels") { levelPicker }
+
```
dfeinzimer:
```suggestion
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
```diff
@@ -0,0 +1,411 @@
...
+                    
+                    Section("Disciplines & Categories") {
+                        ForEach(groupSublayers) { sublayer in
+                            BuildingGroupSublayerToggleView(groupSublayer: sublayer)
+                        }
+                    }
+                }
+                
+            }
+            .onAppear { updateBuildingPicker() }
```
dfeinzimer:
```suggestion
            .onAppear(perform: updateBuildingPicker)
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerItem.swift
```diff
@@ -0,0 +1,40 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ArcGIS
+import Observation
+
+/// An item that can be shown in the ``BuildingExplorer``.
+///- Since: 300.0
+@MainActor
```
dfeinzimer, line 20: Is this needed?

dfeinzimer (commented):
Just an initial partial pass since I know there're more changes in the works.

Every toolkit component also includes a tutorial. Unless the direction is to no longer support or provide them, we should probably include one in this PR for completeness.

zkline101:
>  Unless the direction is to no longer support or provide them, we should probably include one in this PR for completeness.

I was trying to prevent this PR from getting too large. Is it ok if I add it right after this PR?

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerItem.swift
zkline101: I'm not sure that it is "needed" but it seems to make sense. It keeps this type thread safe and it will most likely always be used along side a view so it seemed to just make sense here. `BasemapGalleryItem` is also marked as main actor.

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
zkline101: Yeah that makes sense.

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
zkline101: Thanks for pointing this out. I'll make the change.

Sources/ArcGISToolkit/Documentation.docc/Resources/BuildingExplorer/BuildingExplorer.png
zkline101: Ok I can do that once we get the UI sorted out.

Examples/Examples/BuildingExplorerExampleView.swift
zkline101: Seems to work good. I'll add it.

dfeinzimer: Yep, previews seem to be much more stable lately! 🎉

dfeinzimer:
> I was trying to prevent this PR from getting too large. Is it ok if I add it right after this PR?

Ah yes, sorry, I see that in the description now. Whatever your preference is!

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
```diff
@@ -0,0 +1,410 @@
...
+    /// The proxy which we used to zoom into the building.
+    let localSceneViewProxy: LocalSceneViewProxy?
+    
+    @Environment(\.dismiss) private var dismiss
+    
+    /// The currently selected explorer item.
+    @Binding var selection: BuildingExplorerItem
+    
+    /// The group sublayers which is used to show to build the toggles for the
+    /// discipline and categories.
```
mhdostal:
```suggestion
    /// The group sublayers which is used to show the toggles for the
    /// discipline and categories.
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
```diff
@@ -0,0 +1,410 @@
...
+            ForEach(availableLevels, id: \.self) { level in
+                Text(level)
+            }
+        }
+        .onChange(of: selectedLevel) {
+            if selection.level != selectedLevel {
+                selection.level = selectedLevel
+                
+                // Only change filter if the new selected level
+                // is different. If the user, has a filter
```
mhdostal:
```suggestion
                // is different. If the user has a filter
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
```diff
@@ -0,0 +1,410 @@
...
+            ForEach(availablePhases, id: \.self) { phase in
+                Text(phase)
+            }
+        }
+            .onChange(of: selectedPhase) {
+                if selection.phase != selectedPhase {
+                    selection.phase = selectedPhase
+                    
+                    // Only change filter if the new selected phase
+                    // is different. If the user, has a filter
```
mhdostal:
```suggestion
                    // is different. If the user has a filter
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
```diff
@@ -0,0 +1,410 @@
...
+            // get crowded.
+            picker
+                .pickerStyle(.segmented)
+        } else {
+            picker.pickerStyle(.automatic)
+        }
+    }
+    
+    /// The X button used to close the explorer.
+    @ViewBuilder private var xButton: some View {
```
mhdostal: Can we use [`DismissButton`](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/v.next/Sources/ArcGISToolkit/Common/DismissButton.swift) for part/all of this?

@dfeinzimer - Can the `XButton` be deprecated in favor of the `DismissButton`? There is a lot of overlap in their functionality and I think `DismissButton` is more current.

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
```diff
@@ -0,0 +1,410 @@
...
+            // Create or combine where clauses.
+            if solidWhereClause.isEmpty {
+                solidWhereClause = levelSolidWhereClause
+            } else {
+                solidWhereClause = "\(solidWhereClause) AND \(levelSolidWhereClause)"
+            }
+        }
+        
+        let solidFilterBlock = BuildingFilterBlock(
+            title: "Solid",
```
mhdostal, line 290: Are the titles and description user-facing? Should they be localizable?

mhdostal (changes requested):
This is pretty cool. Nice work!

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
dfeinzimer: > Can the `XButton` be deprecated in favor of the `DismissButton`? There is a lot of overlap in their functionality and I think `DismissButton` is more current.

Possibly so, looks like it's still used in about 6 places. Will look into it in #1385.

mhdostal: @dfeinzimer Excellent, thanks. It would be a good Sprint X task.

zkline101: Ok sounds good. I can use the `DismissButton`

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
zkline101: In this component, they are not user-facing.

zkline101:
I updated the screenshots and pushed the UI changes. This PR is now ready for a full review.

Note: If you want to test out the zoom to building button then you will need to provide a proxy to the `BuildingExplorer` in the example code. I took out the proxy in the example code because it really isn't needed for this scene.

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerItem.swift
```diff
@@ -0,0 +1,40 @@
...
+    /// The filter that was set on the building scene layer by
+    /// the building explorer.
+    public internal(set) var filter: BuildingFilter?
+    
+    init(layer: BuildingSceneLayer) { self.layer = layer }
+}
+
+extension BuildingExplorerItem: @MainActor Equatable {
+    public static func == (lhs: BuildingExplorerItem, rhs: BuildingExplorerItem) -> Bool {
+        lhs.layer.id == rhs.layer.id
```
rolson: might want to do this instead because it wouldn't surprise me for layers to have the same id. or at least across web maps.

```suggestion
        lhs.layer === rhs.layer
```

rolson (dismissed):
FYI I have just skimmed the public api and the UI/UX. David and Mark can give it a thorough review.

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerItem.swift
zkline101: Ok, I'm fine with changing this.

rolson (dismissed):

Examples/Examples/BuildingExplorerExampleView.swift
```diff
@@ -0,0 +1,50 @@
...
+
+import ArcGIS
+import ArcGISToolkit
+import SwiftUI
+
+struct BuildingExplorerExampleView: View {
+    @State private var scene = Scene(url: URL(string: "https://www.arcgis.com/home/item.html?id=b7c387d599a84a50aafaece5ca139d44")!)!
+    @State private var explorerIsVisible = false
+    @State private var items: [BuildingExplorerItem] = []
+    @State private var selection: BuildingExplorerItem?
```
dfeinzimer, line 27: (Nitpick - take or leave) I feel like these should probably have some doc. Examples are good places to explain and teach how to use component initializers.

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
```diff
@@ -0,0 +1,189 @@
...
+                    localized: "Global scenes aren't supported in the Building Explorer.",
+                    bundle: .toolkitModule,
+                    comment: "Description of error thrown when a global scene is being provided to the Building Explorer."
+                )
+            case .noBuildingSceneLayers:
+                String(
+                    localized: "No building scene layers in the scene.",
+                    bundle: .toolkitModule,
+                    comment: "Description of error thrown there are no building scene layers in the scene provided to the Building Explorer."
+                )
```
dfeinzimer, line 178: Please update the strings files for localization with these and any other strings added in this PR by running the [update-translations](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/v.next/update-translations) script.

dfeinzimer (commented):
Just started my second round but found a few things so far.

I'm hitting a crash on Catalyst right when opening the component up. The stack trace doesn't have much useful info in it. I'm using today's swift-daily.

mhdostal (dismissed):

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
```diff
@@ -0,0 +1,402 @@
...
+                    
+                    // These options are only helpful if the
+                    // layer is visible.
+                    if layerIsVisible {
+                        fullModelToggle
+                        
+                        if showFullModel {
+                            if !availableLevels.isEmpty {
+                                levelPicker
+
```
dfeinzimer:
```suggestion
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
```diff
@@ -0,0 +1,402 @@
...
+                            }
+                        }
+                    }
+                }
+                
+                // Only show options to filter the full model
+                // if the full model is being shown and the BSL
+                // is visible.
+                if showFullModel && layerIsVisible {
+                    Section("Disciplines & Categories") {
```
dfeinzimer: Localization needed

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
```diff
@@ -0,0 +1,402 @@
...
+                    overviewSublayer.isVisible = !showFullModel
+                }
+        }
+    }
+    
+    /// The zoom to building button.
+    @ViewBuilder private var zoomToButton: some View {
+        if let proxy = localSceneViewProxy,
+           let layerExtent = selection.layer.fullExtent {
+            Button("Zoom to building", systemImage: "plus.magnifyingglass") {
```
dfeinzimer: Localization needed

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
```diff
@@ -0,0 +1,402 @@
...
+#if os(visionOS)
+                ToolbarItem(placement: .principal) { layerPicker }
+#else
+                ToolbarItem(placement: .principal) { Text(selectedLayerName) }
+#endif
+                ToolbarItem(placement: .topBarTrailing) { DismissButton(kind: .close) }
+            }
+// visionOS puts the title menu above the main window which doesn't
+// make sense in this context so we don't use the title menu
+// on visionOS. Instead we put the layer picker in a toolbar item.
```
dfeinzimer: Xcode auto-indents these lines differently.

```suggestion
            // visionOS puts the title menu above the main window which doesn't
            // make sense in this context so we don't use the title menu
            // on visionOS. Instead we put the layer picker in a toolbar item.
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
```diff
@@ -0,0 +1,402 @@
...
+                
+                // Update explorer contents.
+                await updateForm()
+            }
+        }
+    }
+    
+    /// The building scene layer visibility toggle.
+    private var layerVisibilityToggle: some View {
+        Toggle("Visible", isOn: $layerIsVisible)
```
dfeinzimer: Localization needed

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
```diff
@@ -0,0 +1,402 @@
...
+            .onChange(of: layerIsVisible) {
+                selection.layer.isVisible = layerIsVisible
+            }
+    }
+    
+    /// The layer picker which is used to switch between different building
+    /// scene layers in the scene.
+    private var layerPicker: some View {
+        Picker("Building Scene Layers", selection: $selectedLayerName) {
```
dfeinzimer: Localization needed

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
```diff
@@ -0,0 +1,402 @@
...
+        .pickerStyle(.inline)
+#endif
+    }
+    
+    /// A toggle to switch between the visibility of the full model sublayer.
+    @ViewBuilder private var fullModelToggle: some View {
+        // We need a valid overview and full model for this
+        // toggle to be functional.
+        if let overviewSublayer, let fullModelSublayer {
+            Toggle("Show Full Model", isOn: $showFullModel)
```
dfeinzimer: Localization needed

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
```diff
@@ -0,0 +1,402 @@
...
+                await proxy.setViewpointCamera(camera, duration: 1.5)
+                
+                shouldZoomToBuilding = false
+            }
+        }
+    }
+    
+    /// The level picker to select a level on the building.
+    private var levelPicker: some View {
+        Picker("Level", selection: $selectedLevel) {
```
dfeinzimer: Localization needed

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
```diff
@@ -0,0 +1,402 @@
...
+                // explorer. Only change the filter if they
+                // selected a new level.
+                selection.layer.activeFilter = levelAndPhaseFilter
+            }
+        }
+    }
+    
+    /// The construction phase picker to select a phase on the building.
+    private var phasePicker: some View {
+        Picker("Construction Phase", selection: $selectedPhase) {
```
dfeinzimer: Localization needed

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
```diff
@@ -0,0 +1,402 @@
...
+    /// and phase that the user has selected.
+    private var levelAndPhaseFilter: BuildingFilter? {
+        // If no level or phase is selected then we need show everything
+        // so return 'nil' for the filter.
+        guard selectedLevel != .allLabel || !selectedPhase.isEmpty else { return nil }
+        
+        // Construct the where clauses based on the selection state.
+        
+        var solidWhereClause = ""
+        var xrayWhereClause = ""
```
dfeinzimer: For hyphenated words like X-ray, I believe convention is to treat them as two words for purposes of camel casing.


```suggestion
        var xRayWhereClause = ""
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
```diff
@@ -0,0 +1,402 @@
...
+            }
+        }
+        
+        let solidFilterBlock = BuildingFilterBlock(
+            title: "Solid",
+            whereClause: solidWhereClause,
+            mode: .solid()
+        )
+        
+        let xrayFilterBlock = BuildingFilterBlock(
```
dfeinzimer, line 295:
```suggestion
        let xRayFilterBlock = BuildingFilterBlock(
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingSublayerToggleView.swift
```diff
@@ -0,0 +1,38 @@
...
+    var body: some View {
+        Toggle(sublayer.name, isOn: $isVisible)
+            .onChange(of: isVisible) {
+                sublayer.isVisible = isVisible
+            }
+            .onAppear {
+                // Sets the value of the toggle to the
+                // current visibility of the sublayer.
+                isVisible = sublayer.isVisible
+            }
```
dfeinzimer: (Nitpick - take or leave) Alpha sorted modifiers:

```suggestion
            .onAppear {
                // Sets the value of the toggle to the
                // current visibility of the sublayer.
                isVisible = sublayer.isVisible
            }
            .onChange(of: isVisible) {
                sublayer.isVisible = isVisible
            }
```

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
```diff
@@ -0,0 +1,189 @@
...
+    ///   - items: The items being shown in the building explorer.
+    ///   - selection: The selected item in the building explorer.
+    ///   - localSceneViewProxy: The proxy to provide access to the set viewpoint
+    ///   operation for the "Zoom to building" button. If the proxy is `nil` then that button
+    ///   will not display in the toolbar.
+    public init(
+        scene: ArcGIS.Scene,
+        items: Binding<[BuildingExplorerItem]>,
+        selection: Binding<BuildingExplorerItem?>,
+        localSceneViewProxy: LocalSceneViewProxy? = nil
```
dfeinzimer, line 84: I'm not sure it would make sense to have a BSL in a GlobalSceneView, but if that support is ever added would this component also change? If so, would it likely require a seperate initializer? If not, is there common proxy type we could take instead?

Examples/Examples/BuildingExplorerExampleView.swift
```diff
@@ -0,0 +1,50 @@
...
+                ToolbarItem(placement: .topBarTrailing) {
+                    Button("Building Explorer", systemImage: "building") {
+                        explorerIsVisible = true
+                    }
+                    .popover(isPresented: $explorerIsVisible) {
+                        BuildingExplorer(
+                            scene: scene,
+                            items: $items,
+                            selection: $selection
+                        )
```
dfeinzimer, line 44: Even though there's only item in the example data, seems like it still might be nice to pass in the proxy here to demo the capability. Your call.

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
```diff
@@ -0,0 +1,189 @@
...
+    /// The scene being used to create items for the explorer.
+    private let scene: ArcGIS.Scene
+    /// The proxy needed to set the viewpoint when pressing the
+    /// "Zoom to building" button.
+    private let localSceneViewProxy: LocalSceneViewProxy?
+    
+    /// An error that occurred while setting up the explorer.
+    @State private var setUpError: (any Error)?
+    /// A Boolean value indicating if the explorer finished setting up.
+    @State private var setUpIsDone = false
```
dfeinzimer: I think we should be able to make use of [`Result`](https://developer.apple.com/documentation/swift/result) here to combine these two properties and clean things up a bit.

```suggestion
    /// The result of setting up the explorer.
    @State private var setUpResult: Result<Void, (any Error)>?
```

This would allow `body` to change to something like this:
```swift
Group {
    switch setUpResult {
    case .success(_):
        BuildingExplorerForm(
            items: items,
            selection: $selection,
            localSceneViewProxy: localSceneViewProxy
        )
    case .failure(let failure):
        ContentUnavailableView("\(failure.localizedDescription)", systemImage: "exclamationmark.triangle")
    case nil:
        ProgressView()
    }
}
.task {
    setUpResult = await Result { try await setUp() }
}
```
and `setUp()` to be marked as throwing, allowing you to remove the `do {} catch {}` within it.

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
zkline101: BSL is supported in global scenes, we just don't have that support yet. I have a plan for this, I'll DM you.

Examples/Examples/BuildingExplorerExampleView.swift
zkline101: Yeah, I considered that but in this case the zoom button doesn't add much since the initial viewpoint is at the building already. Thoughts?

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
zkline101: Ok that sounds good to me.

Examples/Examples/BuildingExplorerExampleView.swift
dfeinzimer: Yup that's fine. Just thought it might be useful to get the user back to back origin if they're playing around with the example and happen to zoom way out or something.

zkline101:
This PR is ready for review again, thank you!

By the way, I found a weird bug with popovers in iOS 26 that causes the initial open of a popover to hang for a small amount. I'll work on making an Apple issue for that but it doesn't need to hold up this PR.

dfeinzimer (approved):
This is a cool new component, ship it! 🚢

mhdostal (approved):

zkline101:
Thanks everyone! I'm glad to see this will make it in for 300 🙌

PR #1388, opened 2026-03-19
doc: Fix doc build warnings for OfflineManager

yo1995:
- `OfflineManager/onJobCompletion` is an internal symbol, so I removed the symbol link
- Renamed the parameters for the method docs

### DIFF
--- a/Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
+++ b/Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
@@ -36,7 +36,7 @@ import SwiftUI
 /// - Access map info for web maps that have saved map areas via `OfflineManager.shared.offlineMapInfos`.
 /// - Remove offline map areas from the device.
 /// - Run download jobs while the app is in the background.
-/// - Get notified when the jobs complete via the ``OfflineManager/onJobCompletion`` property.
+/// - Get notified when the jobs complete via the `OfflineManager/onJobCompletion` property.
 ///
 /// The component is useful both for building custom UI with the provided APIs,
 /// and for supporting workflows that require retrieving offline map areas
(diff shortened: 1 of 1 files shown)


Sources/ArcGISToolkit/Components/Offline/OfflineManager.swift
```diff
@@ -36,7 +36,7 @@ import SwiftUI
 /// - Access map info for web maps that have saved map areas via `OfflineManager.shared.offlineMapInfos`.
 /// - Remove offline map areas from the device.
 /// - Run download jobs while the app is in the background.
-/// - Get notified when the jobs complete via the ``OfflineManager/onJobCompletion`` property.
+/// - Get notified when the jobs complete via the `OfflineManager/onJobCompletion` property.
```
dfeinzimer: If `onJobCompletion` is internal, that would make it an implementation detail, wouldn't it? Should this doc be referencing the two public modifiers instead?

`offlineManager(configuration:onJobCompletion:)`
`offlineManager(preferredBackgroundStatusCheckSchedule:jobCompletion:)`

yo1995: I think you are right, though I'm not sure how to mention those two methods as they are on `SwiftUI.Scene` extension. What about 

"Get notified when the jobs complete via the `offlineManager(configuration:onJobCompletion:)` and `offlineManager(preferredBackgroundStatusCheckSchedule:jobCompletion:)` methods on SwiftUI's `Scene`"?

dfeinzimer: No idea why autocomplete doesn't work on it but this builds and links correctly!

```suggestion
/// - Get notified when the jobs complete via ``SwiftUI/Scene/offlineManager(preferredBackgroundStatusCheckSchedule:jobCompletion:)`` or ``SwiftUI/Scene/offlineManager(configuration:onJobCompletion:)``.
```

dfeinzimer (approved):

des12437 (approved):

PR #1391, opened 2026-03-24
`FeatureFormView` - Fix clear button on `TextInput` on visionOS

dfeinzimer:
Apollo 1659

### DIFF
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/TextInput.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/TextInput.swift
@@ -78,12 +78,6 @@ struct TextInput: View {
                 element.convertAndUpdateValue(text)
                 embeddedFeatureFormViewModel.evaluateExpressions()
             }
-            .contentShape(.rect)
-            .onTapGesture {
-                if element.isMultiline {
-                    fullScreenTextInputIsPresented = true
-                }
-            }
 #if !os(visionOS)
             .sheet(isPresented: $scannerIsPresented) {
                 CodeScanner(code: $text, isPresented: $scannerIsPresented)
@@ -115,7 +109,11 @@ private extension TextInput {
                                 .environment(featureFormViewModel)
 #endif
                         }
-                        .frame(minHeight: 100, alignment: .top)
+                        .frame(maxWidth: .infinity, minHeight: 100, alignment: .topLeading)
+                        .contentShape(.rect)
+                        .onTapGesture {
+                            fullScreenTextInputIsPresented = true
+                        }
                 } else {
                     TextField(
                         element.label,
@@ -140,7 +138,6 @@ private extension TextInput {
                     }
                 }
             }
-            .frame(maxWidth: .infinity, alignment: .leading)
 #if os(iOS)
             .toolbar {
                 ToolbarItemGroup(placement: .keyboard) {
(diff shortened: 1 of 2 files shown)


Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/TextInput.swift
```diff
@@ -78,12 +78,6 @@ struct TextInput: View {
                 element.convertAndUpdateValue(text)
                 embeddedFeatureFormViewModel.evaluateExpressions()
             }
-            .contentShape(.rect)
-            .onTapGesture {
-                if element.isMultiline {
-                    fullScreenTextInputIsPresented = true
-                }
-            }
```
dfeinzimer, line 86: I found this to be intersecting taps. It only needs to be on multiline inputs so its been moved accordingly.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/TextInput.swift
```diff
@@ -166,6 +163,8 @@ private extension TextInput {
                     text.removeAll()
                 }
                 .accessibilityIdentifier("\(element.label) Clear Button")
+                // Make the button reliably tappable on visionOS.
+                .hoverEffect(isEnabled: false)
```
dfeinzimer: This increases the button's tappable area without creating a visual highlight effect. The visual effect doesn't look good unless we also add padding and when we do that, there's layout shift when the button appears, which isn't a good look either.

mhdostal (dismissed):

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/TextInput.swift
```diff
@@ -166,6 +160,12 @@ private extension TextInput {
                     text.removeAll()
                 }
                 .accessibilityIdentifier("\(element.label) Clear Button")
+#if os(visionOS)
+                .buttonStyle(.bordered)
+                // Constrain the size of the bordered button on visionOS to
+                // prevent layout shift.
+                .frame(idealWidth: 21, idealHeight: 21)
```
dfeinzimer, line 167: The system didn't respod to any lower size. Higher numbers caused layout shift.

I also tried looking at using `@ScaledMetric` but our test app didn't seem responsive to accessibility font sizes on the Vision Pro simulator.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/TextInput.swift
```diff
@@ -126,9 +124,6 @@ private extension TextInput {
                     .accessibilityIdentifier("\(element.label) Text Input")
                     .focused($isFocused)
                     .keyboardType(keyboardType)
-#if os(visionOS)
-                    .hoverEffectDisabled()
-#endif
```
dfeinzimer, line 131: This restores the hover effect to single line inputs. @zkline101 noticed it was missing and I found it broke when we transitioned to using `Form`.

dfeinzimer:
@zkline101 I applied your feedback, bringing back hover effect both to the button and the input field itself. This is ready for your review once again.

zkline101 (approved):
Thanks David

mhdostal (approved):

PR #1348, opened 2025-12-29
implement fields popup url detection

mmoosbac94:
Regarding this issue: https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/1146, I have implemented a solution to recognize phone numbers in addition to HTTP links. This was very urgent for our client, so we needed a quick solution. Perhaps you can consider it for the next update.

PR #1398, opened 2026-03-28
`PopupView` - Enhance `FieldsPopupElement` link detection

### LINKED ISSUE #1146: Enable phone number recognition in Feature PopupView
Our client requires the ability to add phone numbers as an attribute in the features popup view. These phone numbers should be recognized as clickable links, enabling users to initiate calls directly. Would it be possible to implement it? Or am I missing something?

For web links its working fine, but in the popup, they appear as 'View' elements to tap. Why is that?

dfeinzimer:
Closes #1146 

- Incorporates community contributions (from [#1348](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/1348) by @mmoosbac94) to support hyperlink, phone number _AND_ email address recognition in `FieldsPopupElement`s.
- A single field can contain multiple links in its string and each one will be detected and tappable.
- Single hyperlinks now display the actual link's text instead of "View".

| Before | After |
|---|---|
|  |  |


mhdostal (dismissed):

Sources/ArcGISToolkit/Utility/DataDetector.swift
```diff
@@ -0,0 +1,49 @@
...
+            return nil
+        }
+        
+        let fullRange = NSRange(text.startIndex..., in: text)
+        let matches = detector.matches(in: text, options: [], range: fullRange)
+        
+        return matches.compactMap { match in
+            if let phone = match.phoneNumber {
+                let cleaned = phone
+                    .components(separatedBy: CharacterSet.decimalDigits.inverted)
```
njarecha:
```suggestion
                    .components(separatedBy: .decimalDigits.inverted)
```

Sources/ArcGISToolkit/Components/Popups/FieldsPopupElementView.swift
```diff
@@ -58,37 +58,28 @@ struct FieldsPopupElementView: View {
...
                 }
-#if os(visionOS)
-                    .buttonStyle(.bordered)
-#else
-                    .buttonStyle(.borderless)
-#endif
-            } else {
-                Text(formattedValue)
             }
+            return Text(attributed)
```
njarecha, line 83: Is it expected to return `attributed` string even if no data detectors found?

njarecha (changes requested):
A suggestion and a question

mhdostal (dismissed):

Sources/ArcGISToolkit/Components/Popups/FieldsPopupElementView.swift
dfeinzimer: Addressed in 77893f667fcc99dbee4ab6c3818400d1a6b6d3f9

mhdostal (approved):

dg0yal (approved):

njarecha (approved):

PR #1400, opened 2026-03-30
fix preplanned map model test

rolson:
Fixes the preplanned map model failure. Need to specify in the OfflineManager to use the job manager instead of continued processing tasks.


des12437 (approved):

rolson:
@yo1995 I'd like to get this merged before the daily build if possible. Thank you.

yo1995 (approved):

PR #1402, opened 2026-03-30
Bump versions for 300.0

yo1995:
- Bump the Xcode project "Validate Settings" version to Xcode 26.0. This can be triggered by Xcode > Editor > Validate Settings…
    - Because Xcode 26.0 doesn't run on the latest macOS anymore, if you click it, it would validate using your current version of Xcode (2630 or 2640 for example), and is likely to trigger a few warnings. I only bumped the versions without apply those recommended changes.
Screenshot 2026-03-30 at 4 01 54 PM

- Bump the Xcode project versions to 300.0
- Bump the Swift SDK package dependency version to 300.0
- Cleaned up `#available` macro and `@available` attributes, bump up to iOS 17 and visionOS 2.0


AuthenticationExample/AuthenticationExample.xcodeproj/project.pbxproj
```diff
@@ -8,7 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		88AD13792834355000500B2E /* AuthenticationApp.swift in Sources */ = {isa = PBXBuildFile; fileRef = 88AD13782834355000500B2E /* AuthenticationApp.swift */; };
-		88AD137B2834355000500B2E /* SigninView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 88AD137A2834355000500B2E /* SigninView.swift */; };
+		88AD137B2834355000500B2E /* SignInView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 88AD137A2834355000500B2E /* SignInView.swift */; };
```
yo1995, line 11: This was a fix suggested by Xcode in order to make the project compile.

dfeinzimer (dismissed):

yo1995:
This PR is ready for another review.

dfeinzimer (approved):

rolson (approved):
WE didn't support 16 at 200.8 so I'm not sure why this wasn't done before. Do you know?

yo1995:
It is due to that a target's minimum deployment overrides the project's deployment version, so we won't receive a warning even if the project min ver is iOS 16, when the target's min ver is iOS 17. 

For the changes in this PR, because those example app projects only have 1 target (which is an iOS app), we updated the targets' deployment version correctly in the past releases, but have overlooked their projects' deployment version.

In short, these changes are more of housekeeping than actually affecting the minimum iOS version that the various example apps can build to.

In shorter: should they've been updated? Yes. Did they lead to min target confusion? Probably not.

|Target|Project|
|:-:|:-:|
|Screenshot 2026-04-02 at 12 46 32 PM|Screenshot 2026-04-02 at 12 46 48 PM|

PR #1404, opened 2026-03-31
Use custom flag to to determine how to get the bundle

### LINKED ISSUE #1397: Components can hang on first display
### Issue

Displaying a Toolkit component for the first time can hang and block the main thread for a couple of seconds. This only happens once per app run for the entire Toolkit, e.i, if opening a component hangs, subsequent opens will not, even if it is a different component. This does not happen with all components and in all scenarios (presentation mode, device, etc.). I noticed it with `FeatureFormView` while working on the Feature Editor Demo (`swift/issues/7826`), but it is also present in the Examples app with the `BuildingExplorer`. I am 95% sure it is happening because [`Bundle.init(identifier:)` is getting called](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/v.next/Sources/ArcGISToolkit/Extensions/Foundation/Bundle.swift#L23) on the main thread when the first localized string is created, since [it's doc](https://developer.apple.com/documentation/foundation/bundle/init(identifier:)#Discussion) says: 

> However, if the initial lookup of an already loaded and cached bundle with the specified identifier fails, this method uses potentially time-consuming heuristics to attempt to locate the bundle. As an optimization, you can use the [bundleWithPath:](https://developer.apple.com/documentation/foundation/nsbundle/bundlewithpath:) or [bundleWithURL:](https://developer.apple.com/documentation/foundation/nsbundle/bundlewithurl:) method instead to avoid file system traversal.

### Steps to Reproduce

- Open the "Building Explorer" example on iPhone iOS 26 simulator.
- Tap the building button in the toolbar to open the component and see the hang.

https://github.com/user-attachments/assets/3486e978-8f93-4c37-bdce-33fa8b3d878a

zkline101:
Fixes #1397
cc: @rolson @mhdostal  

The [pitch](https://forums.swift.org/t/pitch-introduce-bundle-macro/79288) for the bundle macro actually discusses the problem we were seeing:

> Or worse yet, they lookup the bundle using its bundle identifier, which while tempting is actually rather inefficient.

So us looking up the bundle by its bundle identifier when using the source code would take a while, hence the problem we were seeing.

In the Bundle [doc](https://developer.apple.com/documentation/foundation/bundle), Apple recommends using this macro:

> In Swift, use the [bundle()](https://developer.apple.com/documentation/foundation/bundle()) macro to insert a bundle instance appropriate to the current execution context, whether an app, app extension, framework, or Swift package.

So we are using that when the source code is used. That macro doesn't work when using the xcframework. The only solution I ended up on was adding a custom flag that we set to false when creating the xcframework and that works. This will require a separate PR in the Swift SDK that I will put up when this is approved.

Before:

https://github.com/user-attachments/assets/89cb7150-3014-46b1-9092-3fe5828c3e2e

After:

https://github.com/user-attachments/assets/d2e6754d-1b99-429a-b434-bb908cc12075

### DIFF
--- a/Sources/ArcGISToolkit/Extensions/Foundation/Bundle.swift
+++ b/Sources/ArcGISToolkit/Extensions/Foundation/Bundle.swift
@@ -20,5 +20,15 @@ extension Bundle {
     
     /// The toolkit module, which is either the resource bundle or the
     /// ArcGISToolkit framework, depending on how the toolkit was built.
-    static let toolkitModule = Bundle(identifier: toolkitIdentifier) ?? .module
+    static let toolkitModule: Bundle = {
+#if USE_BUNDLE_MACRO
+        #bundle
+#else
+        // Only create the bundle using the ID if the toolkit
+        // is being used as a built framework. Otherwise time-consuming
+        // heuristics will be used to locate the bundle and that
+        // can block the main thread.
+        Bundle(identifier: toolkitIdentifier)!
+#endif
+    }()
 }
--- a/Sources/ArcGISToolkit/Extensions/Foundation/Bundle.swift
+++ b/Sources/ArcGISToolkit/Extensions/Foundation/Bundle.swift
@@ -21,14 +21,14 @@ extension Bundle {
     /// The toolkit module, which is either the resource bundle or the
     /// ArcGISToolkit framework, depending on how the toolkit was built.
     static let toolkitModule: Bundle = {
-#if USE_BUNDLE_MACRO
-        #bundle
-#else
+#if USE_BUNDLE_IDENTIFIER
         // Only create the bundle using the ID if the toolkit
         // is being used as a built framework. Otherwise time-consuming
         // heuristics will be used to locate the bundle and that
         // can block the main thread.
         Bundle(identifier: toolkitIdentifier)!
+#else
+        #bundle
 #endif
     }()
 }
--- a/Package.swift
+++ b/Package.swift
@@ -40,9 +40,6 @@ let package = Package(
             dependencies: [
                 .product(name: "ArcGIS", package: "arcgis-maps-sdk-swift"),
                 .product(name: "Markdown", package: "swift-markdown")
-            ],
-            swiftSettings: [
-                .define("USE_BUNDLE_MACRO")
             ]
         ),
         .testTarget(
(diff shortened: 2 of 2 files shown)


zkline101:
Welp `#bundle` doesn't work when the toolkit is being used as a `xcframework` but I have another idea that will fix this.

dfeinzimer:
> Welp #bundle doesn't work when the toolkit is being used as a xcframework but I have another idea that will fix this.

If there's a lightweight method to detect package vs framework to avoid unwrapping `Bundle(identifier: toolkitIdentifier)` and going straight to `Bundle.module`, I think that should get us the same improvements as `#bundle`.

zkline101:
Yep that is the direction I'm looking at.

Sources/ArcGISToolkit/Extensions/Foundation/Bundle.swift
```diff
@@ -20,5 +20,15 @@ extension Bundle {
...
+    static let toolkitModule: Bundle = {
+#if USE_BUNDLE_MACRO
+        #bundle
+#else
+        // Only create the bundle using the ID if the toolkit
+        // is being used as a built framework. Otherwise time-consuming
+        // heuristics will be used to locate the bundle and that
+        // can block the main thread.
+        Bundle(identifier: toolkitIdentifier)!
+#endif
```
dfeinzimer: When we create the framework, is it possible to define new flags so we can do the inverse of this and avoid touching the package?

E.g something like `USE_BUNDLE_IDENTIFIER` and then the above would become:
```swift
#if USE_BUNDLE_IDENTIFIER
        // Only create the bundle using the ID if the toolkit
        // is being used as a built framework. Otherwise time-consuming
        // heuristics will be used to locate the bundle and that
        // can block the main thread.
        Bundle(identifier: toolkitIdentifier)!
#else
        #bundle
#endif
```

dfeinzimer (dismissed):

Sources/ArcGISToolkit/Extensions/Foundation/Bundle.swift
zkline101: Good idea. I think that could work let me try.

rolson: I like that idea as well.

zkline101: Yep this worked. I pushed the changes.

CalebRas (approved):

dfeinzimer (approved):

PR #1407, opened 2026-04-01
Fix construction phase bug in `BuildingExplorer`

zkline101:
I was testing a web scene (that I will send to the reviewers on Slack) and I was using the explorer with it. When I changed the level, I realized the filters weren't behaving as I expected. 

After some debugging, it looks like the problem came down to that there was a single construction phase present and it wasn't a number which threw off the filters being applied. I would consider this an edge case because I've never seen a phase that wasn't a number. But we should still fix the behavior and this PR does that by requiring the construction phase to be a number. The web scene viewer has similar behavior.

### DIFF
--- a/Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
+++ b/Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
@@ -52,9 +52,9 @@ struct BuildingExplorerForm: View {
     // MARK: Phase picker properties.
     
     /// The selected phase in the phase picker.
-    @State private var selectedPhase = ""
+    @State private var selectedPhase: Int?
     /// The available phase in the phase picker.
-    @State private var availablePhases: [String] = []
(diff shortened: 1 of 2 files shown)


Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
```diff
@@ -238,7 +238,8 @@ struct BuildingExplorerForm: View {
     private var phasePicker: some View {
         Picker(selection: $selectedPhase) {
             ForEach(availablePhases, id: \.self) { phase in
-                Text(verbatim: phase)
+                Text(verbatim: "\(phase)")
```
dfeinzimer:
```suggestion
                Text(phase, format: .number)
```

zkline101:
I talked to @kmueller-gis about this solution since he is the author of the BE common design and he agrees with this solution.

dfeinzimer (approved):

mhdostal (approved):

PR #1422, opened 2026-04-07
Extract building scene layers from a `GroupLayer`

zkline101:
We found a situation where a scene could be passed to the BE which contains a group layer and that contains the BSL.
So we check if the scene has any group layers and search those for building scene layers too.

I'll DM the test data to the reviewers.

### DIFF
--- a/Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
+++ b/Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
@@ -120,8 +120,7 @@ public struct BuildingExplorer: View {
             throw SetUpError.globalScenesNotSupported
         }
         
-        let buildingSceneLayers = scene.operationalLayers
-            .compactMap { $0 as? BuildingSceneLayer }
+        let buildingSceneLayers = extractBuildingSceneLayers(from: scene.operationalLayers)
         
         guard !buildingSceneLayers.isEmpty else {
             throw SetUpError.noBuildingSceneLayers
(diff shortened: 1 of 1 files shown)


Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
```diff
@@ -153,6 +152,22 @@ public struct BuildingExplorer: View {
...
+    private func extractBuildingSceneLayers(from layers: [Layer]) -> [BuildingSceneLayer] {
+        var buildingSceneLayers: [BuildingSceneLayer] = []
+        for layer in layers {
+            if let buildingSceneLayer = layer as? BuildingSceneLayer {
+                buildingSceneLayers.append(buildingSceneLayer)
+            } else if let groupLayer = layer as? GroupLayer {
+                buildingSceneLayers.append(contentsOf: extractBuildingSceneLayers(from: groupLayer.layers))
+            }
+        }
+
```
dfeinzimer, line 168:
```suggestion
```

dfeinzimer (approved):

mhdostal (approved):

Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorer.swift
zkline101: I kept this blank line deliberately because I think it helps improve readability but maybe that is just a personal preference. So I would like to keep it unless you feel strongly?

PR #1425, opened 2026-04-08
Update BasemapGallery.swift

mhdostal:
Update screenshots to remove Developer watermark.

Fixes Swift #7880


yo1995 (dismissed):
Changes look good. Hope they show up correctly in the daily build toolkit API ref 🤞

dfeinzimer (commented):
These URLs 404 for me and I'm worried they may continue to not resolve correctly. I think it would be better to refer to local resources, that way we can also support dark mode.

`FeatureFormView` does this [here](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/v.next/Sources/ArcGISToolkit/Components/FeatureFormView/FeatureFormView.swift#L23) and the image are stored [here (light)](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/v.next/Sources/ArcGISToolkit/Documentation.docc/Resources/FeatureFormView/FeatureFormView%403x.png) and [here (dark)](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/v.next/Sources/ArcGISToolkit/Documentation.docc/Resources/FeatureFormView/FeatureFormView~dark%403x.png).

mhdostal:
> These URLs 404 for me and I'm worried they may continue to not resolve correctly. I think it would be better to refer to local resources, that way we can also support dark mode.

Ok, I'll give that a try. The `BasemapGallery` did it differently, but I like the idea of displaying both light/dark mode images. Any special tricks to capturing the screenshots, as relates to image size? The `FeatureForm` one is only iPhone and only shows the `Form`. The existing `BasemapGallery` ones show the whole device (map included) and have both iPhone and iPad versions. @yo1995 @dfeinzimer Any preference to showing both iPhone/iPad or is one enough?

Note the `BasemapGallery` was updated to display in a `sheet`, so maybe both platforms are not necessary.

dfeinzimer:
> Any preference to showing both iPhone/iPad or is one enough?

My feeling is that one would be enough to communicate the general appearance of the component.

yo1995:
I'm also in favor of storing images locally, and have both light/dark variants 👍 

I'm fine with iPhone screenshot alone.

dfeinzimer:
> Any special tricks to capturing the screenshots, as relates to image size?

I believe for `FeatureFormView` I used the built-in capture method in the Simulator app and used the base model of the latest device which at the time was iPhone 16 IIRC.

mhdostal:
I've updated this to only use one screenshot, but with light/dark variants. I used the same naming scheme as the `FeatureFormView` @dfeinzimer referenced.

Sources/ArcGISToolkit/Documentation.docc/Resources/BasemapGallery/BasemapGallery~dark@3x.png
dfeinzimer, line 1: Looks like dark mode isn't working. The filename should be `BasemapGallery~dark@3x.png‎` (tilde replaces the dash, and the `x` needs to be lowercased).

dfeinzimer: There's a Dark Mode control you can add to Control Center for quick testing.

https://github.com/user-attachments/assets/c03d4751-1fc5-45fa-8ec0-8e4a367f749d

yo1995 (approved):

dfeinzimer (approved):

PR #1427, opened 2026-04-13
Update `swift-tools-version` to 6.2

philium:
The [Swift tools version](https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/settingswifttoolsversion/) declares "the minimum version of the Swift compiler required to use the package." Since our declared minimum Xcode version is now 26.0, which shipped with tools version 6.2, increasing our declared tools version to 6.2 makes sense. Doing so will prevent the SDK and toolkit from being used as dependencies of packages that have a swift tools version lower than 6.2, but if we don't change it, we are saying that the package can be used as a dependency of a package built using Swift compiler 6.0 (Xcode 16–16.2) and 6.1 (Xcode 16.3–16.4), which we do not claim to support nor do we build/test with.

Note that starting with Swift 6.0, the tools version no longer needs to be on the first line of the package manifest, so I have moved it to below the preamble.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,5 +1,3 @@
-// swift-tools-version: 6.0
-// The swift-tools-version declares the minimum version of Swift required to build this package.
 // Copyright 2021 Esri
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


yo1995 (commented):
Should it stay on the first line of the manifest file? https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/settingswifttoolsversion/

dfeinzimer (approved):

philium:
> Should it stay on the first line of the manifest file?

That documentation is out of date. See the SPM [changelog](https://github.com/swiftlang/swift-package-manager/blob/main/CHANGELOG.md#swift-60):

> image

yo1995 (approved):

PR #1428, opened 2026-04-14
AR: Replace deprecated SceneKit APIs with RealityKit

des12437:
Related to `swift/7184`.

Removes deprecated SceneKit API usage with RealityKit:

 - Replace SceneKit `ARSCNView` with `RealityKit.ARView`
- Use RealityKit `ARView.raycast(from:allowing:alignment:)` method
- Use RealityKit delegate methods for anchor updates:
     `ARSessionDelegate.session(_ session: ARSession, didAdd anchors: [ARAnchor])`
      `ARSessionDelegate.session(_ session: ARSession, didUpdate anchors: [ARAnchor])`  
- Refactor addPlane and updatePlane methods to use RealityKit to visualize identified planes using a ModelEntity.


Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
```diff
@@ -107,51 +107,53 @@ extension ARSwiftUIView: UIViewRepresentable {
...
 }
 
-extension ARSwiftUIView.Coordinator: ARSCNViewDelegate {
-    func renderer(_ renderer: SCNSceneRenderer, didAdd node: SCNNode, for anchor: ARAnchor) {
-        let sceneParameters = SceneParameters(renderer: renderer, node: node, anchor: anchor)
-        Task { @MainActor [onAddNodeAction] in
-            onAddNodeAction?(sceneParameters)
+extension ARSwiftUIView.Coordinator: ARSessionDelegate {
+    func session(_ session: ARSession, didAdd anchors: [ARAnchor]) {
+        let planeAnchors = anchors.compactMap { $0 as? ARPlaneAnchor }
```
yo1995, line 141: Good to add some explanations on why only the `ARPlaneAnchor`s are handled, while other types of anchors aren't.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
```diff
@@ -27,10 +28,10 @@ struct ARSwiftUIView {
...
     /// The closure to call when the session's frame updates.
     private(set) var onDidUpdateFrameAction: ((ARSession, ARFrame) -> Void)?
-    /// The closure to call when a node corresponding to a new anchor has been added to the view.
-    private(set) var onAddNodeAction: (@MainActor (SceneParameters) -> Void)?
-    /// The closure to call when a node has been updated to match it's corresponding anchor.
-    private(set) var onUpdateNodeAction: (@MainActor (SceneParameters) -> Void)?
+    /// The closure to call when a new plane anchor has been added to the view.
+    private(set) var onAddAnchorAction: (@MainActor (ARPlaneAnchor) -> Void)?
+    /// The closure to call when a plane anchor has been updated.
+    private(set) var onUpdateAnchorAction: (@MainActor (ARPlaneAnchor) -> Void)?
```
yo1995, line 34: For our currently implementation, only plane anchors were used for table top and real world plane detection. A few thoughts:

- For all the other [`ARAnchor`](https://developer.apple.com/documentation/arkit/aranchor) types, I feel like we are unlikely to use them except [`ARGeoAnchor`](https://developer.apple.com/documentation/arkit/argeoanchor). But even if we need other anchor types, we can tweak in the future and it is not a breaking change.
- So I feel OK to simply rename the action to `onAddAnchorAction` instead of a more verbose `onAddPlaneAnchorAction`. Just want to mention it to see if others have comments.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
```diff
@@ -184,6 +178,11 @@ class ARSwiftUIViewProxy: NSObject, @preconcurrency ARSessionProviding {
     @objc dynamic var session: ARSession {
```
yo1995, line 180: I forgot why the `@objc dynamic` keywords are needed. 🤔 

I'll run a tabletop test on real device to see if they make a difference. Please leave a comment if you know that. Was that for some `ARSessionProviding` compatibility issue?

yo1995 (commented):
Some initial comments

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
des12437: Done.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
des12437: I don't recall why this annotation is needed, but I think it was included in the initial implementation.

chriswebb09:
LGTM, I'll run a test on the tabletop.

Sources/ArcGISToolkit/Components/Augmented Reality/TableTopSceneView.swift
yo1995: I've been reading the doc for [ARSCNViewDelegate](https://developer.apple.com/documentation/arkit/arscnviewdelegate) and [ARSessionDelegate](https://developer.apple.com/documentation/arkit/arsessiondelegate). Both have the "didRemove" delegate method, and from the doc it sounds like this method *could be* called automatically, or not called, depending on the session configuration.

In our case, we only run and pause the AR session on the appearance of the view, which is likely to only happen once per session. So I suppose not cleaning up the anchors on view disappear or plane changes *might be* OK. 
Still, I want to pitch this change to ensure each session starts from a clean slate, so the plane anchors are created from scratch. What do you think?

```swift
arViewProxy.session.run(configuration, options: [.removeExistingAnchors])
```

Sources/ArcGISToolkit/Components/Augmented Reality/TableTopSceneView.swift
```diff
@@ -173,65 +176,79 @@ public struct TableTopSceneView: View {
...
         
         // Set help text when plane visualization is updated.
         withAnimation {
             helpText = .planeFound
         }
     }
+    
+    /// Creates a mesh resource for a plane anchor.
+    /// - Parameter anchor: The plane anchor.
+    private func makeMesh(from anchor: ARPlaneAnchor) -> MeshResource {
```
yo1995, line 230: Can you share some doc on this method? I'd like to read a bit more about how the mesh creation works.

Sources/ArcGISToolkit/Components/Augmented Reality/TableTopSceneView.swift
```diff
@@ -173,65 +176,79 @@ public struct TableTopSceneView: View {
...
+        
+        var positions = [SIMD3<Float>]()
+        var triangles = [UInt32]()
+        
+        for vertex in vertices {
+            positions.append([vertex.x, 0, vertex.z])
+        }
+        for index in triangleIndices {
+            triangles.append(UInt32(index))
+        }
```
yo1995: Maybe this and line 234 can be simplified into 

```swift
let triangles = anchor.geometry.triangleIndices.map(UInt32.init)
```

Sources/ArcGISToolkit/Components/Augmented Reality/TableTopSceneView.swift
```diff
@@ -173,65 +176,79 @@ public struct TableTopSceneView: View {
...
+    /// - Parameter anchor: The plane anchor.
+    private func makeMesh(from anchor: ARPlaneAnchor) -> MeshResource {
+        let vertices = anchor.geometry.vertices
+        let triangleIndices = anchor.geometry.triangleIndices
+        
+        var positions = [SIMD3<Float>]()
+        var triangles = [UInt32]()
+        
+        for vertex in vertices {
+            positions.append([vertex.x, 0, vertex.z])
```
yo1995: This line is a bit unclear to me. I'd like to read more to understand why the `y` component is dropped. Can it just be ?

```swift
descriptor.positions = MeshBuffers.Positions(vertices)
```

Sources/ArcGISToolkit/Components/Augmented Reality/TableTopSceneView.swift
```diff
@@ -173,65 +176,79 @@ public struct TableTopSceneView: View {
...
+            isMetallic: true
+        )
         
-        // Add SCNMaterial to plane geometry.
-        let material = SCNMaterial()
-        material.isDoubleSided = true
-        material.diffuse.contents = UIColor.white.withAlphaComponent(0.65)
-        planeGeometry.materials = [material]
+        let planeEntity = ModelEntity(mesh: mesh, materials: [material])
+        planeEntity.position = [0, 0, 0]
```
yo1995:
```suggestion
```
I think the origin position is default to (0, 0, 0)

Sources/ArcGISToolkit/Components/Augmented Reality/TableTopSceneView.swift
```diff
@@ -173,65 +176,79 @@ public struct TableTopSceneView: View {
...
-        guard let planeAnchor = anchor as? ARPlaneAnchor,
-              let planeNode = node.childNodes.first,
-              let planeGeometry = planeNode.geometry as? ARSCNPlaneGeometry
-        else { return }
+        guard let planeEntity = planeEntities[planeAnchor.identifier] else { return }
         
-        // Update extent visualization to the anchor's new geometry.
-        planeGeometry.update(from: planeAnchor.geometry)
+        planeEntity.model?.mesh = makeMesh(from: planeAnchor)
+        planeEntity.position = [0, 0, 0]
```
yo1995:
```suggestion
```
Same as above, I think the position of a model would be default to the (0, 0, 0) origin, but please verify before apply the change.

Sources/ArcGISToolkit/Components/Augmented Reality/TableTopSceneView.swift
```diff
@@ -173,65 +176,79 @@ public struct TableTopSceneView: View {
...
         
-        guard let planeAnchor = anchor as? ARPlaneAnchor,
-              let planeNode = node.childNodes.first,
-              let planeGeometry = planeNode.geometry as? ARSCNPlaneGeometry
-        else { return }
+        guard let planeEntity = planeEntities[planeAnchor.identifier] else { return }
         
-        // Update extent visualization to the anchor's new geometry.
-        planeGeometry.update(from: planeAnchor.geometry)
+        planeEntity.model?.mesh = makeMesh(from: planeAnchor)
```
yo1995, line 220: I'm a bit unclear about this - can you explain or share some info on why this is the equivalent for [`planeGeometry.update(from: planeAnchor.geometry)`](https://developer.apple.com/documentation/arkit/arscnplanegeometry/update(from:))? From my understanding, we need to make a new mesh on every update, because the mesh is immutable for an entity, so we cannot simply update its geometry.

I don't see any performance downgrade or dropped frame when running the app, but wonder if creating a new mesh on each `onUpdateAnchor` call is the best way, as it is called frequently when moving the camera around.

yo1995 (commented):
Thanks for addressing the feedback. A few more questions to better understand the changes…

Sources/ArcGISToolkit/Components/Augmented Reality/TableTopSceneView.swift
des12437: The y values are dropped so that the visualized mesh plane is flat. If the y values are preserved then the mesh plane will match the real-world surface. The previous SceneKit implementation visualized the planes as flat.

Sources/ArcGISToolkit/Components/Augmented Reality/TableTopSceneView.swift
des12437: The position is the origin by default, I removed this line.

Sources/ArcGISToolkit/Components/Augmented Reality/TableTopSceneView.swift
des12437: Great suggestion, I applied this feedback.

Sources/ArcGISToolkit/Components/Augmented Reality/TableTopSceneView.swift
des12437: > From my understanding, we need to make a new mesh on every update, because the mesh is immutable for an entity, so we cannot simply update its geometry.

This is correct. There is not an equivalent way to update an existing plane geometry like there is using SceneKit.

> I don't see any performance downgrade or dropped frame when running the app, but wonder if creating a new mesh on each onUpdateAnchor call is the best way, as it is called frequently when moving the camera around.

In terms of memory, it is not growing since the previous plane entity is removed before updating.

Sources/ArcGISToolkit/Components/Augmented Reality/TableTopSceneView.swift
des12437: Are you suggesting to add more documentation for the method itself and inside the method?

yo1995: Yeah, good to add a short comment on how the anchor's geometry is translated to mesh's descriptor. 

Honestly I still don't quite understand why `y` is dropped, and even if I tried to visualize the `planeEntity`, I couldn't tell the difference between y = $0.y and y = 0. (if you can provide a screenshot comparison I guess I can see it better 🙈)

I'll trust your judgement here.

yo1995 (dismissed):

chriswebb09 (dismissed):
LGTM

Sources/ArcGISToolkit/Components/Augmented Reality/TableTopSceneView.swift
des12437: Added documentation in the method and changed y = 0 to y = $0.y. From testing y = $0.y the planes visualized look correct. Since identified horizontal planes are relatively flat there may not be a noticable visual difference between using  y = 0 and y = $0.y

chriswebb09 (approved):

yo1995 (approved):

des12437:
Thank you both! @yo1995 @chriswebb09

PR #1432, opened 2026-04-22
Remove iOS 17 support

philium:


mhdostal (approved):
Thanks!! It's always nice to remove code conditions based on iOS versions

dfeinzimer (approved):

PR #1433, opened 2026-04-23
Fix hover effect for location button on visionOS

zkline101:
`swift/8083`

Before:

https://github.com/user-attachments/assets/adb0ab6f-353f-4915-9fc4-fe892132cc27

After:

https://github.com/user-attachments/assets/78e7759e-1374-4f39-9fb3-bf20e21af9d3


rolson (approved):
nice fix, thanks

dfeinzimer (approved):

PR #1436, opened 2026-04-24
Fix hover effect for multi-line text input

zkline101:
`swift/8091`

I tried adding hover effects to the `Text` view to fix this but no solution would look as I would expect. This could be due to some default styling being done by `Form`. So by using a `Button` instead of `Text` we get default hover effects that look as expected.

Before there was no hover effects and now it looks like this:

https://github.com/user-attachments/assets/19f0d3e4-9f38-45b4-9846-2b251c0577d3

I checked the other platforms and they look good as well. But I recommend the reviewers test this locally.

### DIFF
--- a/Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/TextInput.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/Inputs/TextInput.swift
@@ -96,24 +96,25 @@ private extension TextInput {
         HStack {
             Group {
                 if element.isMultiline {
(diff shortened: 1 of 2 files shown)


dfeinzimer (commented):
The code changes look good and the button looks good on all platforms. 

Please look into a test failure in `FeatureFormViewTests.testCase_7_1()`.

zkline101:
Good catch, let me check.

dfeinzimer (approved):

mhdostal (approved):

PR #1437, opened 2026-05-01
`FeatureFormView` - `AttachmentsFormElement` Enhancements

dfeinzimer:
Apollo 1721, 1776, 1780

- Uses new 300.1 `AttachmentsFormElement` APIs to add more controls around how attachments are added to attachment elements.
- Test run # 4914
- For manual testing, run the Test Runner with arguments: `-testCase testCase_14_1`.


Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentImportMenu.swift
```diff
@@ -50,6 +48,12 @@ struct AttachmentImportMenu: View {
     /// A Boolean value indicating whether the attachment photo picker is presented.
     @State private var photoPickerIsPresented = false
     
+#warning("""
+Prototype only. Do not merge to main. 
+This will eventually be available on `element`.
+""")
+    @State private var imageInput: _ImageAttachmentsFormInput?
```
mhdostal: From the spec, it doesn't look like this will be nil, so there will potentially be minor code changes to support the final API. I'm fine with this as it gives the ability to switch the input type. Nice work!

mhdostal (approved):
This looks great. I like the ability to switch the input type.

Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentImportMenu.swift
```diff
@@ -102,20 +146,161 @@ struct AttachmentImportMenu: View {
...
+                            takePhotoButton(input: imageFormInput)
+                        case let videoFormInput as _VideoFormInput:
+                            takeVideoButton(input: videoFormInput)
+                        default: EmptyView()
+                        }
+                    }
+                    if inputs.contains(where: {$0 is _ImageFormInput || $0 is _VideoFormInput}) {
+                        chooseFromLibraryButton()
+                    }
+                    chooseFromFilesButton()
```
mhdostal: How does the `input.inputMethod` impact `chooseFromLibraryButton()` and `chooseFromFilesButton()`?

Are we just ignoring the `input.inputMethod` options if there's more than one input? I'm not sure how you would handle it, without changing the menu into a progressive/cascading menu, because audio/video/image inputs have their own `inputMethod`.

- Audio (example: .capture)
    - Capture
- Video (example: .upload)
    -  choose from Library
    - choose from Files
- Image (example: .any)
    - Capture
    - choose from Library
    - choose from Files

dfeinzimer: Yup so that logic would all be here:

https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/6686806053e95ce286fb8dd84bd0a66ca721bee9/Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentImportMenu.swift#L198-L254

Gist of it is: 
1. If there're 2 or more inputs the authoring experience doesn't let you restrict the method between capture or upload so you'll get the capture option for each the respective inputs + choose from library if there's a photo or video input + always choose from files.
2. If there's only a single input then we look at the method configuration to determine what to display.

mhdostal: Ok, thanks for the explanation.

mhdostal (dismissed):

dfeinzimer:
@njarecha, @mhdostal this is still ready for review. I made a slight tweak to not grab filenames for attachments coming from the Photo Library.

Sources/ArcGISToolkit/Extensions/ArcGIS/AttachmentHelpers/AttachmentsFormElement.swift
```diff
@@ -33,10 +34,89 @@ extension AttachmentsFormElement: AttachmentsFeatureElement {
         }
     }
     
+    /// The max duration configured on the either the element's audio or video form input.
+    ///
+    /// If the element has both an audio and video form input, there will be no max duration configured.
+    var maxDuration: Duration? {
+        if let maxDuration = (inputs.first(where: { $0 is AudioFormInput }) as? AudioFormInput)?.maxDuration {
+            return maxDuration
+        } else if let maxDuration = (inputs.first(where: { $0 is VideoFormInput }) as? VideoFormInput)?.maxDuration {
```
zkline101: Naming these `audioMaxDuration` and `videoMaxDuration` would be clearer.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FormElements/FormElementWrapper/FormElementFooter.swift
```diff
@@ -17,53 +17,117 @@ import SwiftUI
...
-    var footerTextForElement: some View {
+    var text: some View {
         switch element {
+        case let element as AttachmentsFormElement:
+            AttachmentsFormElementFooter(element: element, showsErrorsIfPresent: showsErrorsIfPresent)
+                .task { for await _ in element.$validationErrors { id = .init() } }
         case let element as FieldFormElement:
-            FieldFormElementFooter(element: element)
+            FieldFormElementFooter(element: element, showsErrorsIfPresent: showsErrorsIfPresent)
+                .task { for await _ in element.$validationErrors { id = .init() } }
```
zkline101, line 53: I'm not quite sure what is going on with the `id` and these events. Can you explain, please?

Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentImportMenu.swift
```diff
@@ -97,143 +159,243 @@ struct AttachmentImportMenu: View {
...
-            // Show photo/video and library picker.
-#if !os(visionOS)
-            takePhotoOrVideoButton()
-#endif
-            chooseFromLibraryButton()
-            // Always show file picker, no matter the input type.
-            chooseFromFilesButton()
+            if element.inputs.count >= 2 {
+#if os(iOS)
+                if let _ = element.inputs.first(where: { $0 is ImageFormInput }),
```
zkline101:
```suggestion
                if element.inputs.contains(where: { $0 is ImageFormInput }),
```

Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentImportMenu.swift
```diff
@@ -97,143 +159,243 @@ struct AttachmentImportMenu: View {
...
+                   let videoFormInput = element.inputs.first(where: { $0 is VideoFormInput }) as? VideoFormInput {
+                    takePhotoOrVideoButton(videoFormInput: videoFormInput)
+                } else if let imageFormInput = element.inputs.first(where: { $0 is ImageFormInput }) as? ImageFormInput {
+                    takePhotoButton(input: imageFormInput)
+                } else if let videoFormInput = element.inputs.first(where: { $0 is VideoFormInput }) as? VideoFormInput {
+                    takeVideoButton(input: videoFormInput)
+                }
+#endif // os(iOS)
+                if element.inputs.contains(where: {$0 is ImageFormInput || $0 is VideoFormInput}) {
```
zkline101:
```suggestion
                if element.inputs.contains(where: { $0 is ImageFormInput || $0 is VideoFormInput }) {
```

Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentImportMenu.swift
```diff
@@ -70,25 +82,75 @@ struct AttachmentImportMenu: View {
...
+    }
+#endif // os(iOS)
+    
     private func chooseFromLibraryButton() -> Button<some View> {
         Button {
             photoPickerIsPresented = true
         } label: {
-            Text(libraryButtonLabel)
+            Text(
+                "Choose From Library",
```
zkline101: Can we use the same words Apple does for these options? We could match some of the system images too.

You did this for "Take Photo or Video" but we apply the same to the other options too.

Image

Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentImportMenu.swift
```diff
@@ -97,143 +159,243 @@ struct AttachmentImportMenu: View {
...
+        if let min = element.minAttachmentCount, currentAttachmentCount < min {
+            element.minAttachmentCountMessage
+                .font(.caption)
+                .foregroundStyle(.secondary)
+        }
+        if let max = element.maxAttachmentCount, currentAttachmentCount >= max {
+            element.maxAttachmentCountMessage
+                .font(.caption)
+                .foregroundStyle(.secondary)
+        } else {
```
zkline101, line 240: This message seems a bit out of place, can we put in the footer?
Image

It also seems odd to say "No attachments" when a user can probably assume that if there are no visible attachments. But maybe that can be a future enhancement because that isn't apart of this PR.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FormElements/FormElementWrapper/FormElementFooter.swift
dfeinzimer: It's being used a mechanism to refresh the footer anytime the form's validation error visibility configuration changes or the validation errors on a specific element change.

Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentImportMenu.swift
dfeinzimer: @mhdostal, thoughts on moving this down? I know you had put this line here a while back.

Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentImportMenu.swift
dfeinzimer: Done

Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentImportMenu.swift
dfeinzimer: One consideration if we do want to move it down - we'll need to find a way to visually seperate the element's description.

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FormElements/FormElementWrapper/FormElementFooter.swift
zkline101: Interesting ok. I guess I'm not sure I've seen this pattern before. Why not listen to the error events and store the errors in a state variable?

Sources/ArcGISToolkit/Components/FeatureFormView/AttachmentHelpers/AttachmentImportMenu.swift
zkline101: > we'll need to find a way to visually seperate the element's description.

Maybe just an empty line between the two?

mhdostal: I'm OK with it where it is. The idea behind having it (notifying the user what's expected) matches what Kotlin and the MapViewer does, so I'm OK including it. I like it where it is because it's not part of the author-supplied description (so it should be separate from that) and whether it's above or below the button is a matter of opinion. In the end, it's a UI piece, so can be changed if we get pushback.

mhdostal (dismissed):

mhdostal (dismissed):

Sources/ArcGISToolkit/Components/FeatureFormView/Subviews/FormElements/FormElementWrapper/FormElementFooter.swift
dfeinzimer: I think if `AttachmentsFormElementFooter` or `FieldFormElementFooter` only depended on errors that would be cleaner, but since they both also depend on `validationErrorVisibilityExternal` and `validationErrorVisibilityInternal`, refreshing a single identity when any of theses properties change is cleanest.

zkline101 (approved):

mhdostal (approved):

dfeinzimer:
Thank you both!

PR #1439, opened 2026-05-04
[Update] SDK version

chriswebb09:


yo1995:
I would say no need to bump this either. Back in 100.15.1 we didn't bump the toolkit and kept the SDK dependency as up to next minor 100.15.0.

https://github.com/Esri/arcgis-runtime-toolkit-ios/blob/v100.15.1/Package.swift

We shall do the same for 200.8.0. The only difference this time was that 200.8.1 was kind of a "stop-gap" release that got tagged, so I would say let's keep it 200.8.1.

PR #1440, opened 2026-05-05
Add Building Explorer UI Tests

zkline101:
`home-one/4269`


Sources/ArcGISToolkit/Components/BuildingExplorer/BuildingExplorerForm.swift
```diff
@@ -219,19 +245,6 @@ struct BuildingExplorerForm: View {
...
-                
-                // Only change filter if the new selected level
-                // is different. If the user has a filter
-                // already present then we don't want to
-                // change that when they open up the building
-                // explorer. Only change the filter if they
-                // selected a new level.
-                selection.layer.activeFilter = levelAndPhaseFilter
-            }
-        }
```
zkline101, line 234: The tests exposed a bug where the initial level and phase weren't getting updated properly on the selection. So moving this modifier out to the `Form` fixes that. Before the initial phase and level would get updated when the view appears but this modifier wasn't hooked up yet so the initial level wasn't updated on the selection.

Test Runner/Test Runner/TestViews/BuildingExplorerTestViews/BuildingExplorerTestCase1View.swift
```diff
@@ -0,0 +1,45 @@
...
+
+struct BuildingExplorerTestCase1View: View {
+    @Bindable private var viewModel = BuildingExplorerTestViewModel()
+    
+    @State private var layerIsVisible = false
+    
+    var body: some View {
+        LocalSceneView(scene: viewModel.scene)
+            .overlay(alignment: .top) {
+                Text(verbatim: "Layer is visible: \(layerIsVisible)")
```
dfeinzimer: Verbatim helps avoid generating unwanted localized strings for toolkit source, but I don't believe that's a concern in test runner code or the right tool to solve the warning that appears when it's removed.

I think using `layerIsVisible.description` directly might be a better alternative.

Test Runner/UI Tests/BuildingExplorerTests.swift
```diff
@@ -0,0 +1,138 @@
...
+        app.launch()
+        
+        // Tap building explorer button.
+        app.buttons["Building Explorer Tests"].assertExistenceAndTap()
+        
+        // Tap test case one button.
+        app.buttons["Test Case 1"].assertExistenceAndTap()
+        
+        // Tap the visible toggle.
+
```
dfeinzimer:
```suggestion
```

Test Runner/UI Tests/BuildingExplorerTests.swift
```diff
@@ -0,0 +1,138 @@
...
+        app.launch()
+        
+        // Tap building explorer button.
+        app.buttons["Building Explorer Tests"].assertExistenceAndTap()
+        
+        // Tap test case two button.
+        app.buttons["Test Case 2"].assertExistenceAndTap()
+        
+        // Tap the full model toggle.
+
```
dfeinzimer:
```suggestion
```

Test Runner/UI Tests/BuildingExplorerTests.swift
```diff
@@ -0,0 +1,138 @@
...
+    /// Tests the visible toggle.
+    func testCase1() {
+        let app = XCUIApplication()
+        app.launch()
+        
+        // Tap building explorer button.
+        app.buttons["Building Explorer Tests"].assertExistenceAndTap()
+        
+        // Tap test case one button.
+        app.buttons["Test Case 1"].assertExistenceAndTap()
```
dfeinzimer, line 32: You may want to consider a utility function like this to avoid the repeated steps, especially if more cases will be added over time.

https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/24fc14257f2e3bb3d2250cd643605aa6a44c625c/Test%20Runner/UI%20Tests/FeatureFormViewTests.swift#L30-L41

Test Runner/Test Runner/TestViews/BuildingExplorerTestViews/BuildingExplorerTestCase1View.swift
```diff
@@ -0,0 +1,45 @@
...
+                    .banner()
+            }
```
dfeinzimer: The overlay may be obscured by the sheet depending on the window size. 

Even though it doesn't technically need to be visible, how would you feel about applying the overlays directly on the BE component? Might make it easier to debug/add more test cases in the future.

zkline101: Yeah should be ok

Test Runner/UI Tests/BuildingExplorerTests.swift
zkline101: I thought we usually avoid encapsulating test logic in a method because it makes it harder to read and debug when something fails?

dfeinzimer: Your call either way. 

This probably saves us ~250 repetitive lines across all of the FFV tests. I view it as our own set up layer, similar to `XCTestCase.setUp()`, which I think would technically count as encapsulated logic too?

zkline101: Ok I'm just going to keep it as is then.

dfeinzimer (approved):

mhdostal (approved):

Issue #1441, opened 2026-05-05
v.next crashes on iOS 18


CalebRas:
Running an app that is built with the current `v.next` branch crashes with `SIGABRT` on launch when the app is run on iOS 18 simulators. This does not happen on iOS 26 or with the `main` branch. Based on the console output below, this appears to be because of a missing `RealityKit` symbol.

<details><summary>Console Output</summary><p>

```
dyld[55697]: Symbol not found: _$s10RealityKit12AnchorEntityC6anchorACSo8ARAnchorC_tcfC
  Referenced from: <C23B59D0-B2E3-3C9D-A923-8EF4A63E3C61> /Users/cal13336/Library/Developer/CoreSimulator/Devices/1BF0FABD-E839-4091-8CE4-0AE2201762FA/data/Containers/Bundle/Application/93B75ECC-A46D-447B-845A-E123CCC338A3/Toolkit Examples.app/Toolkit Examples.debug.dylib
  Expected in:     <8096B5FF-D155-3D7F-A566-8A3C1C82C373> /Library/Developer/CoreSimulator/Volumes/iOS_22G86/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS 18.6.simruntime/Contents/Resources/RuntimeRoot/System/Library/Frameworks/RealityKit.framework/RealityKit
Symbol not found: _$s10RealityKit12AnchorEntityC6anchorACSo8ARAnchorC_tcfC
  Referenced from: <C23B59D0-B2E3-3C9D-A923-8EF4A63E3C61> /Users/cal13336/Library/Developer/CoreSimulator/Devices/1BF0FABD-E839-4091-8CE4-0AE2201762FA/data/Containers/Bundle/Application/93B75ECC-A46D-447B-845A-E123CCC338A3/Toolkit Examples.app/Toolkit Examples.debug.dylib
  Expected in:     <8096B5FF-D155-3D7F-A566-8A3C1C82C373> /Library/Developer/CoreSimulator/Volumes/iOS_22G86/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS 18.6.simruntime/Contents/Resources/RuntimeRoot/System/Library/Frameworks/RealityKit.framework/RealityKit
[...]

PR #1442, opened 2026-05-05
Fix crash on iOS 18

des12437:
Closes https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/1441.

Removes use of an unavailable initializer for `AnchorEntity(anchor:)` to resolve the crash on iOS 18. Confirmed the crash was resolved on iOS 18 and 26 and tested that plane visualization still works as expected.


CalebRas (approved):
This fix the crash for me. Thanks for looking at this!

philium (approved):
Thanks for the fix!

PR #1444, opened 2026-05-08
`FilterView` - Expose SPI

dfeinzimer:
- Exposes an SPI (System Programming Interface) for `FilterView`.
- Relocates `FilterView` and `FilterViewModel` under `Common/FilterView/` to represent expanded usage.
- Marks necessary members as `public`, adds needed initializers.


rolson:
@philium thoughts on spi vs the underscore "experimental" convention?

philium:
The best approach depends on the situation. Underscored APIs are best for keeping public APIs as hidden (not discoverable) as possible (e.g. implementation details that must be public for some reason). The `_spi` attribute is best for APIs that are designed to be discovered/used publicly, but 1) may be modified/removed in the future, and/or 2) not intended to be used by all users. Swift Testing's [SPI doc](https://github.com/swiftlang/swift-testing/blob/main/Documentation/SPI.md) is helpful in understanding some real world use cases for the `_spi` attribute.

I don't know anything about this new API, but it doesn't look like a public, non-public implementation detail, nor something we want to keep hidden from users. In that case, `_spi` seems like the most appropriate option. Since this appears to be experimental API, you might consider just using `Experimental` as the name.

rolson:
@philium please review in-lieu of me

rolson:
We should create an issue for turning this into a supported component at some point instead of just being available through SPI indefinitely.

dfeinzimer:
> We should create an issue for turning this into a supported component at some point instead of just being available through SPI indefinitely.

Swift 8143

Sources/ArcGISToolkit/Common/FilterView/FilterView.swift
```diff
@@ -29,7 +30,16 @@ struct FilterView: View {
     /// as cancelling simply resets the list of `FieldFilter` objects.
     var onApplyAction: (() -> Void)?
     
-    var body: some View {
+    /// Creates a view allowing the user to assemble a list of `FieldFilter` objects used to filter a list of features.
+    /// - Parameters:
+    ///   - model: The model used by the view.
+    ///   - onApplyAction: The client-specified action to perform when the `Apply` button is tapped.
+    public init(model: FilterViewModel, onApplyAction: (() -> Void)? = nil) {
```
philium, line 37: Making a view model public and required to initialize a view like this is highly unusual. Before this component is made non-experimental, I would suggest rearchitecting it to use standard SwiftUI patterns (e.g. taking an `isPresented` binding as an initializer parameter, having separate initializers per use case [one that takes a feature table, another that takes fields], etc.). A view model should be a servant, not a leader. It shouldn't comprise the design of the public API. When creating a new view, I recommend building it w/out a view model and then asking "Would a view model be helpful?" Sometimes it is, but sometimes it isn't. I find it telling that none of the views provided by SwiftUI have public-facing view models (as far as I know, at least not by name).

Sources/ArcGISToolkit/Common/FilterView/FilterView.swift
```diff
@@ -16,7 +16,8 @@ import ArcGIS
 import SwiftUI
 
 /// A view allowing the user to assemble a list of `FieldFilter` objects used to filter a list of features.
-struct FilterView: View {
+@_spi(PrototypeFilterView)
```
philium: Consider using `Experimental` as the name.

```suggestion
@_spi(Experimental)
```

If you really want a separate name for this, then I think just `FilterView` would be fine. I think the fact that it is a prototype or experimental is a given due to the use of `_spi`.


```suggestion
@_spi(FilterView)
```

The attribute on `FilterViewModel` would need to be updated as well.

Sources/ArcGISToolkit/Common/FilterView/FilterViewModel.swift
```diff
@@ -17,7 +17,11 @@ import Foundation
 import SwiftUI
 
 @MainActor @Observable
```
philium, line 19: Observable objects used by views are not required to be constrained to the main actor. I see that removing it produces a compile error in the property observer for `featureTable`, but I would argue that the feature table should be loaded in a `task` modifier, not implicitly by the model.

philium (dismissed):

Sources/ArcGISToolkit/Common/FilterView/FilterView.swift
dfeinzimer: Thank you Phil, I've added this as a to-do to Swift 8143.

Sources/ArcGISToolkit/Common/FilterView/FilterViewModel.swift
dfeinzimer: 👍 Also added to Swift 8143

philium (approved):

mhdostal (approved):

PR #1446, opened 2026-05-11
Respect `Visibility.visible` in `editingButtons(_:)`

### LINKED ISSUE #1258: FeatureFormView: Need a way to show save button immediately on show for new features
Consider this case:
- new feature
- no validation errors
- shouldn't need to set any attributes to get the save button

Notice in the gif below that the save button doesn't appear until you set an attribute, but you shouldn't need to set an attribute.

Image

dfeinzimer:
Closes #1258, ref Apollo 1338

The [`editingButtons(_:)`](https://developers.arcgis.com/swift/v200/toolkit-api-reference/documentation/arcgistoolkit/featureformview/editingbuttons(_:)/) modifier doesn't currently respect [`Visibility.visible`](https://developer.apple.com/documentation/SwiftUI/Visibility/visible), but can, in order to support quickly saving newly created features.


mhdostal (dismissed):

Test Runner/Test Runner/TestViews/FeatureFormViewTestView.swift
```diff
@@ -315,7 +317,6 @@ private extension ArcGISFeature {
         if let id = attributes["objectid"] as? Int64 {
             return id
         } else {
-            print(type(of: attributes["objectid"]!))
             return nil
         }
```
philium: Without the print statement, this whole property could be simplified to the following:

```swift
var objectID: Int64? { attributes["objectid"] as? Int64 }
```

philium (dismissed):

philium (dismissed):

philium (changes requested):
There are a couple more project changes necessary:

Image

Just do something that causes Xcode to re-save the project file (e.g. swap two files and then swap them back).

mhdostal (approved):

philium (approved):

PR #1445, opened 2026-05-08
`Scalebar` - Nest public types

dfeinzimer:
Closes #685 

- Nests `ScalebarSettings`, `ScalebarStyle` and `ScalebarUnits` under `Scalebar` as `Settings`,  `Style` and `Units`.
- Adds deprecated `typealias`s for `ScalebarSettings`, `ScalebarStyle` and `ScalebarUnits`.

PR #1447, opened 2026-05-12
Fix build

dfeinzimer:
Amends #1445 to repair build errors.

Test run 4571


philium:
I'm sorry for not being more careful with reviewing the scalebar changes. Another option is to put all the private members in a separate initializer:

```swift
extension Scalebar.Units {
    ...
}

private extension Scalebar.Units {
    ...
}
```

That approach is preferred because the access levels don't need to be applied at the individual member level.

dfeinzimer:
> I'm sorry for not being more careful with reviewing the scalebar changes.

No worries, my fault for not running a build. 😞

philium (approved):
Verified that toolkit and tests build successfully.

chriswebb09 (approved):

PR #1448, opened 2026-05-15
`FeatureEditor` - Move basic `GeometryEditorToolbar` into staging

CalebRas:
## Description

This PR adds a new, basic `GeometryEditorToolbar` component to the `FeatureEditor` staging branch. This component is designed to control the `GeometryEditor` used by the Feature Editor component. It is exposed as a standalone component, since it will need to be placed inside an app's navigation stack (on the map or in the toolbar), while the Feature Editor will need to be placed above the navigation stack to use the inspector. An example, a tutorial, and tests will be added in future PRs.

Closes `swift/issues/8164`.

### New Public API

- `GeometryEditorToolbar(geometryEditor: GeometryEditor, layout: Layout? = .vertical)`
- `GeometryEditorToolbar.Layout`
   - `vertical`
   - `horizontal`

## How to Test

1. Open the ad hoc example in https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/tree/Caleb/GeometryEditorToolbar-Example.
2. Tap the plus button in the bottom left corner to start the geometry editor with a geometry type.
3. Tap on the map to identify a feature and start the geometry editor with a geometry.
4. Use the toolbar controls to help edit.

## Screenshots

The screenshots below show the three different layout options on each platform. Notes on the styling:
- The bottom toolbar example uses the `nil` for the layout, which displays the controls with no custom styling.
- The toolbar's styling was modeled after the native toolbar style on iOS & iPadOS.
- I was unable to figure out how to make the `ToolPicker` icon bigger to match the other controls on Mac Catalyst, so I made an issue to address that later, `swift/issues/8173`.
- More styling work will happen later as part of `swift/issues/7403`.

| | Light | Dark |
|:-:|:-:|:-:|
| **iPhone** | Simulator Screenshot - iPhone 17 Pro - 2026-05-18 at 12 48 21 | Simulator Screenshot - iPhone 17 Pro - 2026-05-18 at 13 57 49 |
| **iPad** | Simulator Screenshot - iPad (A16) - 2026-05-18 at 12 49 44 | Simulator Screenshot - iPad (A16) - 2026-05-18 at 13 50 01 |
| **Mac Catalyst** | Screenshot 2026-05-18 at 1 34 24 PM | Screenshot 2026-05-18 at 1 54 55 PM |


Sources/ArcGISToolkit/Components/GeometryEditorToolbar/GeometryEditorToolbar.swift
```diff
@@ -0,0 +1,243 @@
...
+    public init(geometryEditor: GeometryEditor, layout: Layout? = .vertical) {
+        self.geometryEditor = geometryEditor
+        self.layout = layout
+        
+        let model = GeometryEditorToolbarModel(geometryEditor: geometryEditor)
+        self._model = .init(wrappedValue: model)
+    }
+    
+    /// The padding to add to the control labels to increase their hit box size.
+    private let controlLabelPadding = 8.5
```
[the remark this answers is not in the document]
CalebRas: `Double` and `CGFloat` are actually interchangeable, so this isn't an issue.

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/ToolPicker.swift
```diff
@@ -0,0 +1,173 @@
...
+        } else {
+            return Tool.allCases
+        }
+    }
+}
+
+// MARK: - Tool
+
+/// A tool that can be shown in the tools menu.
+private enum Tool: Hashable {
```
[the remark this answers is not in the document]
CalebRas: This can be handled when tests are added in another PR.

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/ToolPicker.swift
```diff
@@ -0,0 +1,173 @@
...
+    /// The tools that are currently able to be selected.
+    @State private var selectableTools: [Tool] = []
+    /// The tool selected by the picker.
+    @State private var selectedTool: Tool = .vertex
+    
+    var body: some View {
+        Menu {
+            Picker("Tool", selection: $selectedTool) {
+                ForEach(Tool.allCases, id: \.self) { tool in
+                    Label(tool.label, systemImage: tool.systemImage)
```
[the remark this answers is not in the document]
CalebRas: Selection updates fine without this.

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/GeometryEditorToolbar.swift
```diff
@@ -0,0 +1,268 @@
...
+    /// A Boolean value indicating whether the geometry editor has started.
+    private(set) var isStarted = false
+    
+    /// A task that observes the geometry editor's `isStarted` stream.
+    @ObservationIgnored private var isStartedTask: Task<Void, Never>?
+    
+    init(geometryEditor: GeometryEditor) {
+        self.geometryEditor = geometryEditor
+        
+        isStartedTask = Task { @MainActor [weak self] in
```
yo1995:
```suggestion
        isStartedTask = Task { [weak self] in
```
I may be wrong, but I think a Task created in the MainActor annotated class would inherit the context (which means `isStartedTask` is also on main actor), thus no need to repeat the info.

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/GeometryEditorToolbar.swift
```diff
@@ -0,0 +1,268 @@
...
+            for await isStarted in geometryEditor.$isStarted {
+                self?.isStarted = isStarted
+            }
+        }
+    }
+    
+    deinit {
+        isStartedTask?.cancel()
```
yo1995:
```suggestion
        if let task = isStartedTask.take() {
            task.cancel()
        }
```
Setting the task to `nil` make the clean up safer.

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/GeometryEditorToolbar.swift
```diff
@@ -0,0 +1,268 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ArcGIS
+import SwiftUI
+
+/// A group of controls for interacting with a geometry editor.
+/// - Note: The view will be displayed when the geometry editor is started.
+public struct GeometryEditorToolbar: View {
```
yo1995, line 45: Let's add doc-strings for the type. Later when we create tutorials for them, we can amend the doc-strings. They can be added once this PR approaches the finishing line.
The doc-strings would consist of: one line description, features, behaviors, and associated types. See other toolkit views for an example.

A few behaviors that comes to my mind to be included:
- when geometry editor is started, the toolbar will show
- default layout is vertical
- `nil` for not having styling, suitable for people who want to provide custom styling

CalebRas:
I updated the description to include some ad hoc testing instructions.

yo1995 (commented):
Overall looks good

rolson:
I don't want to get deep into styling, but one thing that can be improved now is the spacing at the ends of the overlay toolobar (both vertical and horizontal). There isn't quite enough space between the button and the rounded edge. Notice when the styling is nil and iOS renders it that there is a bit more space and it looks more balanced:

image

CalebRas:
> I don't want to get deep into styling, but one thing that can be improved now is the spacing at the ends of the overlay toolobar (both vertical and horizontal). There isn't quite enough space between the button and the rounded edge. Notice when the styling is nil and iOS renders it that there is a bit more space and it looks more balanced:

I updated the horizontal layout to match the spacing of the system toolbar on iOS as close as possible, and then used the same values for the vertical layout. I also removed the control hit-box padding, since I think it works fine without it as long we disable tapping through the toolbar background. Should be good to go now.

Simulator Screenshot - iPhone 17 Pro - 2026-05-19 at 16 51 46
Simulator Screenshot - iPad (A16) - 2026-05-19 at 16 51 19
Screenshot 2026-05-19 at 4 50 51 PM

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/GeometryEditorToolbar.swift
CalebRas: I added some doc-strings, but I'll wait to add a screenshot once Ryan has reviewed the UI.

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/GeometryEditorToolbar.swift
```diff
@@ -0,0 +1,283 @@
...
+    public var body: some View {
+        // This view uses hardcoded padding and spacing values to match the
+        // system styling of a toolbar group on iOS.
+        Group {
+            if model.isStarted {
+                switch layout {
+                case .vertical:
+                    VStack {
+                        controls
+                            .padding(10.5)
```
rolson: Consider defining constants for these padding values.

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/GeometryEditorToolbar.swift
```diff
@@ -0,0 +1,283 @@
...
+                    }
+                    .padding(.vertical, 5)
+                    .stackStyle()
+                    
+                case .horizontal:
+                    HStack {
+                        controls
+                            .padding(10.5)
```
rolson: For HStack and VStack you can also use the spacing parameter.

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/GeometryEditorToolbar.swift
```diff
@@ -0,0 +1,283 @@
...
+    
+    /// A task that observes the geometry editor's `isStarted` stream.
+    @ObservationIgnored private var isStartedTask: Task<Void, Never>?
+    
+    init(geometryEditor: GeometryEditor) {
+        self.geometryEditor = geometryEditor
+        
+        isStartedTask = Task { [weak self] in
+            for await isStarted in geometryEditor.$isStarted {
+                self?.isStarted = isStarted
```
rolson: you can do a `guard let self else { break }` here so that the task will exit when self is deinited. I like that, but not totally necessary because the task is cancelled in deinit.

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/GeometryEditorToolbar.swift
```diff
@@ -0,0 +1,283 @@
...
+        }
+    }
+}
+
+// MARK: - Helper
+
+private extension View {
+    /// A view modifier that applies styles to the vertical and horizontal stacks in `GeometryEditorToolbar`.
+    @ViewBuilder
+    func stackStyle() -> some View {
```
rolson: toolbarStackStyle?

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/ToolPicker.swift
```diff
@@ -0,0 +1,240 @@
...
+    var body: some View {
+        Menu {
+            Picker(selection: $selectedTool) {
+                ForEach(Tool.allCases, id: \.self) { tool in
+                    Label {
+                        tool.label
+                    } icon: {
+                        Image(systemName: tool.systemImage)
+                    }
+                    .selectionDisabled(!selectableTools.contains(tool))
```
rolson: I might prefer to not see the tool rather than to see it disabled. For example, if I'm just modifying point geometries, I don't think I care to see the shape tools disabled. I'd rather not see them IMO.

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/ToolPicker.swift
```diff
@@ -0,0 +1,240 @@
...
+            return FreehandTool()
+        case .shape(let kind):
+            let shapeTool = ShapeTool(kind: kind)
+            // Allows the shape tool to be used when there is an existing geometry.
+            shapeTool.configuration.allowsPartCreation = true
+            return shapeTool
+        case .vertex:
+            return VertexTool()
+        case .vertexReticle:
+            return ReticleVertexTool()
```
rolson, line 107: Maybe this can be enhanced later, but I feel that for many cases your app would choose one or the other (vertex or reticle). Maybe we think about that more once we figure out how to support custom tools.

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/GeometryEditorToolbar.swift
```diff
@@ -0,0 +1,283 @@
...
+    
+    /// The view model for the view.
+    @State private var model: GeometryEditorToolbarModel
+    
+    /// Creates a geometry editor toolbar view.
+    /// - Parameters:
+    ///   - geometryEditor: The geometry editor that this toolbar controls.
+    ///   - layout: The layout to apply to the controls. A `nil` value renders
+    ///   the controls without built-in layout or styling.
+    public init(geometryEditor: GeometryEditor, layout: Layout? = .vertical) {
```
rolson: I liked Phil's feedback on the constructor. I'm thinking about that a bit more. It does mean for the overlay case the user would have to pass something, but I think it clarifies the non-overlay case.

Do you think that the options should be "horizontal", "vertical" and "system" or "systemBar"? So instead of `nil` we would denote that with a name that specified more information?

rolson: > Do you think that the options should be "horizontal", "vertical" and "system" or "systemBar"? So instead of nil we would denote that with a name that specified more information?

This might be an invalid/bad suggestion.

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/ToolPicker.swift
CalebRas: Yeah, that will most likely require adding the tool customization modifier which we determined was not minship, so I made an issue for this, `swift/issues/8186`. In the meantime, would you like me to leave it as is or just have one vertex tool that is set based on the platform [like this](https://github.com/Esri/arcgis-maps-sdk-swift-samples/blob/v.next/Shared/Samples/Snap%20geometry%20edits/SnapGeometryEditsView.GeometryEditorModel.swift#L32-L38)?

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/GeometryEditorToolbar.swift
CalebRas: > I liked Phil's feedback on the constructor. I'm thinking about that a bit more. It does mean for the overlay case the user would have to pass something, but I think it clarifies the non-overlay case.

As we discussed today, having two constructors clarifies the unstyled case better, but it makes unstyled the default, and the styled functionality less discoverable. Since we anticipate styled being the main use case, I will stick with one constructor for now so that styled remains the default.

> Do you think that the options should be "horizontal", "vertical" and "system" or "systemBar"? So instead of nil we would denote that with a name that specified more information?

I updated the layout naming to clarify the unstyled case better:
- `Layout` -> `Style`: This makes the most sense to me because it describes what we are doing the best: giving the user a view with a predefined style, rather than just arranging the controls. It is also what is used by other components, such as [BasemapGallery](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/v.next/Sources/ArcGISToolkit/Components/BasemapGallery/BasemapGallery.swift#L55-L65) and [Scalebar](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/v.next/Sources/ArcGISToolkit/Components/Scalebar/ScalebarStyle.swift#L16-L28). Other options I considered: `Presentation`, `DisplayMode`, `Display`, `Configuration`, `Variant`, and `Axis`.
- `nil` -> `Style.plain`: Using a case instead of a nil value for unstyled makes the API clearer and the functionality more discoverable/documentable. Other possible options: `unstyled`, `unstyledControls`, `plainControls`, `custom`, and `none`.

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/ToolPicker.swift
```diff
@@ -0,0 +1,239 @@
...
+            // Overwrites the initial geometry editor tool when the editor changes.
+            model.geometryEditor.tool = selectedTool.geometryEditorTool
+            
+            for await geometry in model.geometryEditor.$geometry {
+                selectableTools = selectableTools(for: geometry)
+            }
+        }
+    }
+    
+    /// Returns the tools that should be enabled for a given geometry.
```
yo1995:
```suggestion
    /// Returns the tools that should be displayed for a given geometry.
```
maybe?

yo1995 (commented):
Overall looks good. I'm OK with changing the style parameter to optional as Ryan suggested.

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/GeometryEditorToolbar.swift
CalebRas: `Style.plain` has been removed in favor of `nil`.

yo1995 (approved):

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/GeometryEditorToolbar.swift
```diff
@@ -0,0 +1,289 @@
...
+    
+    deinit {
+        if let task = isStartedTask.take() {
+            task.cancel()
+        }
+    }
+}
+
+public extension GeometryEditorToolbar {
+    /// A style that determines the appearance and layout of a geometry editor toolbar.
```
rolson, line 143: Should this mention commonly used overlayed on a GeoView?

rolson (approved):
Very nice!

yo1995 (approved):

rolson (approved):

CalebRas:
Thank you both for all the feedback!

PR #1449, opened 2026-05-19
feature editor: Add GeometryEditorToolbar Tests

yo1995:
Add unit tests and UI tests for Geometry Editor Toolbar.

### DIFF
--- a/Test Runner/Test Runner.xcodeproj/project.pbxproj
+++ b/Test Runner/Test Runner.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		007CF8DC2FC66BA600005B3F /* GeometryEditorToolbarTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 007CF8DB2FC66BA600005B3F /* GeometryEditorToolbarTests.swift */; };
+		007CF8DE2FC66BC500005B3F /* GeometryEditorToolbarTestView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 007CF8DD2FC66BC500005B3F /* GeometryEditorToolbarTestView.swift */; };
 		75022F962A7AC9A0000ED7B7 /* FeatureFormViewTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 75022F952A7AC9A0000ED7B7 /* FeatureFormViewTests.swift */; };
 		75384C112B2D1BD800D97E6C /* BookmarksTestCase5View.swift in Sources */ = {isa = PBXBuildFile; fileRef = 75384C102B2D1BD800D97E6C /* BookmarksTestCase5View.swift */; };
 		7552A7542A573CBB0023DA5A /* TestRunnerApp.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7552A7532A573CBB0023DA5A /* TestRunnerApp.swift */; };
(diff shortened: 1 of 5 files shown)


CalebRas:
> Other stuff that could be tested but not due to being private
>
> - The default style of the toolbar is vertical
> - `supportedGeometryTypes` and `selectableTools` method

I think it would be good to test these if possible, particularly `selectableTools`. Consider updating the component's implementation to allow this.

CalebRas:
Will UI tests be added in a separate PR?

Tests/ArcGISToolkitTests/GeometryEditorToolbarTests.swift
```diff
@@ -0,0 +1,48 @@
...
+        try await Task.yield(timeout: 0.1) { @MainActor in
+            model.isStarted
+        }
+        #expect(model.isStarted)
+        
+        geometryEditor.stop()
+        try await Task.yield(timeout: 0.1) { @MainActor in
+            !model.isStarted
+        }
+        #expect(model.isStarted == false)
```
rolson: you can probably add one of these assert false on line 35 as well

yo1995:
> I think it would be good to test these if possible, particularly `selectableTools`. Consider updating the component's implementation to allow this.

> Will UI tests be added in a separate PR?

If we add UI tests, I can see these areas being tested

- The appearing/disappearing of the toolbar when geometryEditor.`start`/`stop` is called
- When the tool picker button is tapped, the appropriate tools are shown
  - if we make the `selectableTools` internal, I can see this test being a unit test instead of a UI test
- For delete, undo, redo buttons, since the directly call the Swift API methods, I don't think they need to be tested

So I can see either 2 UI tests, or 1 UI test and 1 more unit test if `selectableTools` becomes internal. Which do you prefer?

CalebRas:
> If we add UI tests, I can see these areas being tested
>
> - The appearing/disappearing of the toolbar when geometryEditor.`start`/`stop` is called
> - When the tool picker button is tapped, the appropriate tools are shown
>   - if we make the `selectableTools` internal, I can see this test being a unit test instead of a UI test
> - For delete, undo, redo buttons, since the directly call the Swift API methods, I don't think they need to be tested
>
> So I can see either 2 UI tests, or 1 UI test and 1 more unit test if `selectableTools` becomes internal. Which do you prefer?

That makes sense. Testing `selectableTools` as a UI test seems a bit more thorough, but I don't have a strong opinion on it, so up to you!

Test Runner/Test Runner.xcodeproj/project.pbxproj
```diff
@@ -228,12 +233,13 @@
 		75B236672A5CB32700AEFACE /* TestViews */ = {
 			isa = PBXGroup;
 			children = (
-				7584B7532B1A798B00772AB7 /* BookmarksTestViews */,
```
yo1995, line 231: Sorted the folder and saw this change.

yo1995:
Ready for another review.

Test Runner/UI Tests/GeometryEditorToolbarTests.swift
```diff
@@ -0,0 +1,111 @@
...
+        let toolButton = app.buttons["Tool"]
+        let deleteButton = app.buttons["Delete Selected Element"]
+        
+        toolButton.assertExistence(timeout: 1)
+        deleteButton.assertExistence(timeout: 1)
+        
+        let toolMidY = toolButton.frame.midY
+        let deleteMidY = deleteButton.frame.midY
+        
+        XCTAssertLessThan(toolMidY, deleteMidY, "Expected the tool button to be above the delete button when the toolbar is vertical")
```
CalebRas: This line is pretty long. Maybe break it up?

CalebRas (approved):
Looks good and tests pass for me.

rolson (approved):

CalebRas (approved):

PR #1450, opened 2026-05-22
`GeometryEditorToolbar` - Add example

CalebRas:
### Description

This PR adds an example for the new `GeometryEditorToolbar` component. I created a new web map for the example (since I couldn't find anything with simple points, lines, and polygons), so let me know if you have any feedback on that as well: [Points, Lines, and Polygons](https://arcgis.com/home/item.html?id=2ffa0c3601144e8bbd2c49cf3a7c0559).

Closes `swift/issues/7407`.

### How To Test

1. Tap on the map to identify a feature to edit.
2. Edit the feature's geometry using the geometry editor and the geometry editor toolbar controls.
3. Use the save and cancel buttons in the system bottom toolbar to finish editing.

### Screenshots

Simulator Screenshot - iPhone 17 Pro - 2026-05-22 at 14 06 13


Examples/Examples/GeometryEditorToolbarExampleView.swift
```diff
@@ -0,0 +1,155 @@
...
+                }
+                .task(id: identifyPoint) {
+                    guard let identifyPoint else { return }
+                    defer { self.identifyPoint = nil }
+                    
+                    do {
+                        try await editFeature(at: identifyPoint, mapViewProxy: mapViewProxy)
+                    } catch {
+                        errorDescription = "\(error)"
+                    }
```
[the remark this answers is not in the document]
CalebRas: The `localizedDescription` is not very helpful with ArcGIS errors, so I will leave it as is.

Examples/Examples/GeometryEditorToolbarExampleView.swift
```diff
@@ -0,0 +1,155 @@
...
+        .disabled(!canFinishEditing)
+        .task(id: isSavingEdits) {
+            guard isSavingEdits else { return }
+            defer { isSavingEdits = false }
+            
+            do {
+                try await saveEdits()
+            } catch {
+                self.errorDescription = "\(error)"
+            }
```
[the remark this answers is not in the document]
CalebRas: Same as above.

Examples/Examples/GeometryEditorToolbarExampleView.swift
```diff
@@ -0,0 +1,155 @@
...
+        let geometry = geometryEditor.stop()
+        
+        guard let featureBeingEdited else { return }
+        
+        featureBeingEdited.geometry = geometry
+        featureBeingEdited.setVisible(true)
+        self.featureBeingEdited = nil
+        
+        try await featureBeingEdited.table?.update(featureBeingEdited)
+    }
```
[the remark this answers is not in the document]
CalebRas: I don't think applying edits is needed for this example, since the point is to just show geometry editing with the toolbar. The save button is only part of this example to show the developer that the toolbar doesn't handle saving or discarding edits.

The UI is also cleared before the table update, so that the app doesn't get stuck on editing if the update fails for some reason.

Examples/Examples/GeometryEditorToolbarExampleView.swift
```diff
@@ -0,0 +1,155 @@
...
+    
+    /// Buttons for ending a geometry editing session.
+    @ViewBuilder private var finishEditingButtons: some View {
+        let canFinishEditing = featureBeingEdited != nil && !isSavingEdits
+        
+        Button("Cancel", systemImage: "xmark", role: .cancel) {
+            geometryEditor.stop()
+            featureBeingEdited?.setVisible(true)
+            featureBeingEdited = nil
```
yo1995, line 91:
```swift
            if let feature = featureBeingEdited.take() {
                feature.setVisible(true)
            }
```
*take* it or leave it ;-)

Examples/Examples/GeometryEditorToolbarExampleView.swift
```diff
@@ -0,0 +1,155 @@
...
+        }
+        .disabled(!canFinishEditing)
+        .task(id: isSavingEdits) {
+            guard isSavingEdits else { return }
+            defer { isSavingEdits = false }
+            
+            do {
+                try await saveEdits()
+            } catch {
+                self.errorDescription = "\(error)"
```
yo1995, line 106:
```suggestion
                errorDescription = "\(error)"
```

Examples/Examples/GeometryEditorToolbarExampleView.swift
```diff
@@ -0,0 +1,155 @@
...
+              let geometry = feature.geometry else {
+            return
+        }
+        
+        geometryEditor.start(withInitial: geometry)
+        feature.setVisible(false)
+        featureBeingEdited = feature
+    }
+    
+    /// Saves the geometry editor edits to the feature being edited.
```
yo1995, line 132: Maybe add a note to explain that the edits aren't applied to the feature service, and is only saved to the local feature table.

Examples/Examples/GeometryEditorToolbarExampleView.swift
```diff
@@ -0,0 +1,155 @@
...
+    private func saveEdits() async throws {
+        let geometry = geometryEditor.stop()
+        
+        guard let featureBeingEdited else { return }
+        
+        featureBeingEdited.geometry = geometry
+        featureBeingEdited.setVisible(true)
+        self.featureBeingEdited = nil
```
yo1995, line 140: Should the `featureBeingEdited` becomes visible and the cache cleaned up after the `featuretable.update` succeeds?

Examples/Examples/GeometryEditorToolbarExampleView.swift
```diff
@@ -0,0 +1,155 @@
...
+        
+        try await featureBeingEdited.table?.update(featureBeingEdited)
```
yo1995, line 142: Guard the table as well, so no need optional chaining.

Examples/Examples/GeometryEditorToolbarExampleView.swift
```diff
@@ -0,0 +1,155 @@
...
+    /// The geometry editor used to edit geometries on the map.
+    @State private var geometryEditor = GeometryEditor()
+    /// The point on the screen that is being identified.
+    @State private var identifyPoint: CGPoint?
+    /// A Boolean value indicating whether the geometry edits are being saved.
+    @State private var isSavingEdits = false
+    /// A map containing example point, line, and polygon features to edit.
+    @State private var map: Map = {
+        let url = URL(
+            string: "https://arcgis.com/home/item.html?id=2ffa0c3601144e8bbd2c49cf3a7c0559"
```
yo1995, line 33: Maybe style the points on AGOL to make them bigger?

rolson (commented):
Looks good. Did you consider building out a simple measure line or area experience instead of a feature editing experience? I wonder if that would simplify the code a bit and also allow the example to not depend on maintaining a feature service.

CalebRas:
> Did you consider building out a simple measure line or area experience instead of a feature editing experience?

@rolson I didn't! My only thoughts against that is the current example represents a more common workflow, and it demonstrates that edit saving/discarding needs to happen at the app level. It would likely simplify the code, though, and I don't feel too strongly about it. Would you like me to go with the measure idea instead?

rolson:
> > Did you consider building out a simple measure line or area experience instead of a feature editing experience?
> 
> @rolson I didn't! My only thoughts against that is the current example represents a more common workflow, and it demonstrates that edit saving/discarding needs to happen at the app level. It would likely simplify the code, though, and I don't feel too strongly about it. Would you like me to go with the measure idea instead?

I think it might be something to ask @dg0yal about. I'm ok either way, but not having to maintain a new service might be nice.

CalebRas:
> I think it might be something to ask @dg0yal about. I'm ok either way, but not having to maintain a new service might be nice.

I discussed this with Ryan and Divesh, and it was recommended to go with a simpler workflow, so I am closing this in favor of https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/1451, which uses the toolbar to create geometries for new graphics, rather than editing features from a feature service. The service will be deleted once it has been determined not to be needed for the upcoming Feature Editor example.

Thank you both for your feedback on this!

PR #1456, opened 2026-06-02
AR: Update `ARSwiftUIView` and `ARCoachingOverlay`

des12437:
Updates the `ARSwiftUIView` and `ARCoachingOverlay` used in world-scale and tabletop AR.

Related to `swift/8215`.

- Renames `ARSwiftUIView` `ARSwiftUIViewProxy` to `ARSessionProvider<ARView>`.
- Sets `ARSessionDelegate.sessionShouldAttemptRelocalization` to `true` and `ARCoachingOverlayViewDelegate.coachingOverlayViewDidRequestSessionReset(_:)` to enable relocalization via the coaching overlay `Start Over` button.
- Moves `raycast` method to `ARView` extension.
- Adds `activatesAutomatically(_:)` view modifier to `ARCoachingOverlay` to disable automatic activation for tabletop AR.
- Sets  `activatesAutomatically(false)` for the `TableTopSceneView` since `activatesAutomatically = true` overrides the manual setting. The coaching overlay is managed manually since it should not be shown again after the tabletop content is placed.

### DIFF
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARCoachingOverlay.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARCoachingOverlay.swift
@@ -46,8 +46,7 @@ extension ARCoachingOverlay {
         return view
     }
     
-    
-    /// Controls whether the coaching overlay activates automatically.
+    /// Sets whether the coaching overlay activates automatically.
     /// - Parameter activatesAutomatically: A Boolean value that
     /// indicates if the coaching overlay activates automatically.
     func activatesAutomatically(_ activatesAutomatically: Bool) -> Self {
@@ -108,28 +107,30 @@ extension ARCoachingOverlay: UIViewRepresentable {
         uiView.goal = goal
     }
     
-    class Coordinator: NSObject, ARCoachingOverlayViewDelegate {
+    class Coordinator: NSObject {
         var coachingOverlay: ARCoachingOverlay
         
         init(coachingOverlay: ARCoachingOverlay) {
             self.coachingOverlay = coachingOverlay
         }
-        
-        func coachingOverlayViewWillActivate(_ coachingOverlayView: ARCoachingOverlayView) {
-            coachingOverlay.onCoachingOverlayActivateAction?(coachingOverlayView)
-        }
-        
-        func coachingOverlayViewDidDeactivate(_ coachingOverlayView: ARCoachingOverlayView) {
-            coachingOverlay.onCoachingOverlayDeactivateAction?(coachingOverlayView)
-        }
-        
-        func coachingOverlayViewDidRequestSessionReset(_ coachingOverlayView: ARCoachingOverlayView) {
-            coachingOverlay.onCoachingOverlayRequestSessionResetAction?(coachingOverlayView)
-        }
     }
     
     func makeCoordinator() -> Coordinator {
         return .init(coachingOverlay: self)
+    }
+}
+
+extension ARCoachingOverlay.Coordinator: ARCoachingOverlayViewDelegate {
+    func coachingOverlayViewWillActivate(_ coachingOverlayView: ARCoachingOverlayView) {
+        coachingOverlay.onCoachingOverlayActivateAction?(coachingOverlayView)
+    }
+    
+    func coachingOverlayViewDidDeactivate(_ coachingOverlayView: ARCoachingOverlayView) {
+        coachingOverlay.onCoachingOverlayDeactivateAction?(coachingOverlayView)
+    }
+    
+    func coachingOverlayViewDidRequestSessionReset(_ coachingOverlayView: ARCoachingOverlayView) {
+        coachingOverlay.onCoachingOverlayRequestSessionResetAction?(coachingOverlayView)
     }
 }
 #endif
--- a/Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARCoachingOverlay.swift
+++ b/Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARCoachingOverlay.swift
@@ -31,9 +31,12 @@ struct ARCoachingOverlay {
     var goal: ARCoachingOverlayView.Goal
     /// A Boolean value that indicates if coaching is in progress.
     var active: Bool = false
-    
+}
+
+extension ARCoachingOverlay {
     /// Controls whether the coaching is in progress.
-    /// - Parameter active: A Boolean value indicating if coaching is in progress.
+    /// - Parameter active: A Boolean value indicating if coaching is in
+    /// progress.
     /// - Returns: The `ARCoachingOverlay`.
     func active(_ active: Bool) -> Self {
         var view = self
@@ -50,7 +53,8 @@ struct ARCoachingOverlay {
         return view
     }
     
-    /// Sets the closure to call when the coaching experience is completely deactivated.
+    /// Sets the closure to call when the coaching experience is completely
+    /// deactivated.
     func onCoachingOverlayDeactivate(
         perform action: @escaping (ARCoachingOverlayView) -> Void
     ) -> Self {
@@ -59,8 +63,8 @@ struct ARCoachingOverlay {
         return view
     }
     
-    /// Sets the closure to call when the user taps the coaching overlay view's Start Over button
-    /// while the session is relocalizing.
+    /// Sets the closure to call when the user taps the coaching overlay view's
+    /// Start Over button while the session is relocalizing.
     func onCoachingOverlayRequestSessionReset(
         perform action: @escaping (ARCoachingOverlayView) -> Void
     ) -> Self {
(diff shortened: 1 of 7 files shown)


Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARCoachingOverlay.swift
```diff
@@ -31,16 +31,31 @@ struct ARCoachingOverlay {
...
+    /// - Parameter active: A Boolean value indicating if coaching is in
+    /// progress.
     /// - Returns: The `ARCoachingOverlay`.
     func active(_ active: Bool) -> Self {
         var view = self
         view.active = active
         return view
     }
     
+
```
philium: We don't need this blank line.

```suggestion    
```

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARCoachingOverlay.swift
```diff
@@ -31,16 +31,31 @@ struct ARCoachingOverlay {
...
+    /// progress.
     /// - Returns: The `ARCoachingOverlay`.
     func active(_ active: Bool) -> Self {
         var view = self
         view.active = active
         return view
     }
     
+    
+    /// Controls whether the coaching overlay activates automatically.
```
philium:
```suggestion
    /// Sets whether the coaching overlay activates automatically.
```

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARCoachingOverlay.swift
```diff
@@ -83,41 +99,37 @@ extension ARCoachingOverlay: UIViewRepresentable {
...
         func coachingOverlayViewDidDeactivate(_ coachingOverlayView: ARCoachingOverlayView) {
-            onCoachingOverlayDeactivateAction?(coachingOverlayView)
+            coachingOverlay.onCoachingOverlayDeactivateAction?(coachingOverlayView)
         }
         
         func coachingOverlayViewDidRequestSessionReset(_ coachingOverlayView: ARCoachingOverlayView) {
-            onCoachingOverlayRequestSessionResetAction?(coachingOverlayView)
+            coachingOverlay.onCoachingOverlayRequestSessionResetAction?(coachingOverlayView)
         }
     }
```
philium: Consider moving the conformance to `ARCoachingOverlayViewDelegate` and these implementations to an extension.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
```diff
@@ -18,203 +18,215 @@ import ARKit
...
+    public func session(_ session: ARSession, didUpdate anchors: [ARAnchor]) {
         // Filter updated plane anchors to use for horizontal plane visualization.
         let planeAnchors = anchors.compactMap { $0 as? ARPlaneAnchor }
         for anchor in planeAnchors {
-            Task { @MainActor [onUpdateAnchorAction] in
+            Task { @MainActor [onUpdateAnchorAction = arSwiftUIView.onUpdateAnchorAction] in
                 onUpdateAnchorAction?(anchor)
             }
         }
     }
```
philium, line 159: If you want, you could go ahead and update the implementation of these methods with the newer version (which avoids creating an intermediate array).

philium (dismissed):

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
```diff
@@ -18,203 +18,215 @@ import ARKit
...
+    /// Creates an instance with the given session provider.
+    /// - Parameter sessionProvider: The session provider which will have its
+    /// view set when available.
+    public init(sessionProvider: ARSessionProvider<ARView>) {
+        self.sessionProvider = sessionProvider
     }
-    
+}
+
+public extension ARSwiftUIView {
```
rolson: I'm having trouble parsing the name. Should this be `SwiftUIARView`?

rolson: Is this public because custom VPS providers will need it for implementing camera feed views?

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
```diff
@@ -18,203 +18,215 @@ import ARKit
...
-    func session(_ session: ARSession, didAdd anchors: [ARAnchor]) {
-        // Filter new plane anchors to use for horizontal plane visualization.
-        let planeAnchors = anchors.compactMap { $0 as? ARPlaneAnchor }
-        for anchor in planeAnchors {
-            Task { @MainActor [onAddAnchorAction] in
-                onAddAnchorAction?(anchor)
+    public func session(_ session: ARSession, didAdd anchors: [ARAnchor]) {
+        for anchor in anchors {
+            // Filter new plane anchors to use for horizontal plane visualization.
+            guard let planeAnchor = anchor as? ARPlaneAnchor else { continue }
```
rolson: I guess you could use compactMap to transform with `as? ARPlaneAnchor` instead of guard stmt, if you want.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
```diff
@@ -18,203 +18,215 @@ import ARKit
...
+            Task { @MainActor [onAddAnchorAction = arSwiftUIView.onAddAnchorAction] in
+                onAddAnchorAction?(planeAnchor)
```
rolson: should a new task be created for every anchor, or just one task around the for loop?

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
```diff
@@ -33,7 +34,7 @@ public struct WorldScaleSceneView: View {
     var calibrationViewIsHidden = false
 #if os(iOS)
     /// The proxy for the ARSwiftUIView.
-    @State private var arViewProxy = ARSwiftUIViewProxy()
+    @State private var arViewProxy = ARSessionProvider<ARView>()
```
rolson, line 37: should this variable name change from arViewProxy to `sessionProvider`?

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
philium: The guard statement approach is more efficient as it doesn't create an intermediate array. Another efficient approach that I didn't think of until just now is to lazily compact map, but that does come with some overhead that may not be worth it depending on the average size of `anchors`.

Sources/ArcGISToolkit/Components/Augmented Reality/WorldScaleSceneView.swift
philium: This property will be going away and the corresponding property (on a different type) will be called `arSessionProvider`.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
des12437: Yes.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
des12437: Is `for case let` a better option since it does not create an intermediate array like compactMap and may improve readability?

```
for case let planeAnchor as ARPlaneAnchor in anchors {
      onAddAnchorAction?(planeAnchor)
}
```

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
des12437: One task around the for loop is sufficient.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
philium: That works! I don't know if I knew about that syntax.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
philium: Reporting all added anchors in a single task makes me think that they should be reported to the user all at once:

```swift
/// Sets the closure to call when new plane anchors have been added to the scene.
func onAnchorsAdded(
    perform action: @escaping @MainActor ([ARPlaneAnchor]) -> Void
) -> Self {
    var copy = self
    copy.onAddAnchorAction = action
    return copy
}
```

des12437: That could be done. For-loops would be needed now when accessing the anchors in the `TableTopSceneView`:
```
.onAnchorsAdded { planeAnchors in
     for anchor in planeAnchors {
           addPlane(for: anchor)
          }
      }
.onAnchorsUpdated { planeAnchors in
     for anchor in planeAnchors {
          updatePlane(for: anchor)
      }
}
```

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
des12437: I think `SwiftUIARView` is not as clear since it starts with `SwiftUI`. What about `ARViewRepresentable`, `ARViewWrapper`, or `ARViewContainer` to make "ARView" more emphasized?

rolson: Maybe the current name is fine, maybe it's just me. You and Phil can think about it, doesn't need to change in this initial PR.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
rolson: Nice one @des12437

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
philium: I'm fine with `SwiftUIARView`. I'm not sure there's much of a difference between that and what we have now. Both roughly communicate the same idea.

des12437: Ok, renamed the view to `SwiftUIARView `.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
des12437: Applied this feedback.

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
```diff
@@ -121,61 +121,61 @@ extension ARSwiftUIView: UIViewRepresentable {
     }
     
     public func updateUIView(_: ARView, context: Context) {
-        context.coordinator.arSwiftUIView = self
+        context.coordinator.swiftUIARView = self
     }
     
     public class Coordinator: NSObject {
-        var arSwiftUIView: ARSwiftUIView
+        var swiftUIARView: SwiftUIARView
```
rolson: consider calling this property just `view`

Sources/ArcGISToolkit/Components/Augmented Reality/Utilities/ARSwiftUIView.swift
```diff
@@ -121,61 +121,61 @@ extension ARSwiftUIView: UIViewRepresentable {
...
         
-        init(arSwiftUIView: ARSwiftUIView) {
-            self.arSwiftUIView = arSwiftUIView
+        init(swiftUIARView: SwiftUIARView) {
```
rolson: consider just taking in `view` here

rolson (approved):

philium (approved):
Didn't realize this was ready for re-review. Sorry about that!

PR #1457, opened 2026-06-02
`GeometryEditorToolbar` - Add snap settings button

CalebRas:
## Description

This PR adds a button that presents a sheet for configuring snap settings to the `GeometryEditorToolbar`.

Closes `swift/issues/8165`.

### Behavior

- The toolbar button is hidden when `snapSettings.sourceSettings` is empty.
- The "Snap to Geometry Guides", and "Snap to Features" toggles are hidden when "Enable Snapping" is off.

#### Snap Source Toggles

- Are disabled if their parent toggle is off.
- Are disabled if their `ruleBehavior` is `rulesPreventSnapping`.
- Display a subtitle for non-nil `ruleBehavior`.

## How To Test

- Open the "Geometry Editor Toolbar Snapping" example in the [Caleb/TestSnapSettings](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/tree/Caleb/TestSnapSettings) branch.
- Tap a feature on the map to start the geometry editor.
- Tap the gear button in the toolbar to open the settings sheet
- Use the controls to configure snap settings.

**Note**: The data in the test example doesn't have any snap rules, so the view will look different than the screenshots.

## Screenshots

|Before|After|
|:-:|:-:|
| Simulator Screenshot - iPhone 17 Pro - 2026-06-02 at 12 20 04 | Simulator Screenshot - iPhone 17 Pro - 2026-06-02 at 12 20 40 |

### Platforms 

|||
|:-:|:-:|
| **iPhone** | Simulator Screenshot - iPhone 17 Pro - 2026-06-02 at 15 01 10 |
| **iPad** | Simulator Screenshot - iPad (A16) - 2026-06-02 at 15 04 57 |
| **Mac Catalyst** | Screenshot 2026-06-02 at 3 05 53 PM |
| **Mac Catalyst** | Screenshot 2026-06-02 at 3 07 00 PM |


Sources/ArcGISToolkit/Components/GeometryEditorToolbar/GeometryEditorToolbar.swift
```diff
@@ -82,7 +86,7 @@ public struct GeometryEditorToolbar: View {
                     .padding(.vertical, stackEdgePadding)
                     .toolbarStackStyle()
                 case .horizontal:
-                    HStack(spacing: 30) {
+                    HStack(spacing: stackSpacing) {
```
CalebRas, line 87: Fixed since I missed this in https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/1448.

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/GeometryEditorToolbar.swift
```diff
@@ -241,6 +246,38 @@ private struct UndoButton: View {
...
+                    )
+                } icon: {
+                    Image(systemName: "gear")
+                }
+            }
+            .sheet(isPresented: $isShowingSettings) {
+                SnapSettingsView(settings: model.geometryEditor.snapSettings)
+                // Needed to override the font set in toolbarStackStyle.
+                    .font(nil)
+            }
```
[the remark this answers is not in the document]
CalebRas: This will be done in another PR.

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/SnapSettingsView.swift
```diff
@@ -0,0 +1,270 @@
...
+import ArcGIS
+import SwiftUI
+
+/// A view with controls for configuring `SnapSettings` properties.
+struct SnapSettingsView: View {
+    /// The snap settings to configure.
+    let settings: SnapSettings
+    
+    /// A Boolean value indicating whether the settings have haptic feedback enabled.
+    @State private var hapticFeedbackIsEnabled = false
```
rolson: I feel like this is a developer setting, not a user setting. One approach is that we enable it by default and then if users want to be able to turn it off, we can give them a programmatic way of turning it off.

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/SnapSettingsView.swift
```diff
@@ -0,0 +1,270 @@
...
+    
+    /// A Boolean value indicating whether the disclosure group is expanded.
+    @State private var groupIsExpanded = false
+    /// The view models for the root `SnapSourceSettingsToggle` views in the outline group.
+    @State private var rootSourceSettingsModels: [SnapSourceSettingsToggle.Model] = []
+    /// A Boolean value indicating whether the settings allow snapping to features.
+    @State private var snapsToFeatures = false
+    
+    var body: some View {
+        DisclosureGroup(isExpanded: $groupIsExpanded) {
```
rolson: Do we need a disclosure group for this? There are no UI elements below this group, so I'm not sure why it would need to be collapsed by the end user. I would just show all the subitems when snapsToFeatures is true and hide them when not.

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/SnapSettingsView.swift
CalebRas: That makes sense. I removed the toggle but didn't set `SnapSettings.hapticFeedbackIsEnabled` because it is enabled by default in the SDK, and the developer can set the property before passing the geometry editor to the toolbar if they want it disabled.

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/SnapSettingsView.swift
CalebRas: I was using a `DisclosureGroup` to nest the toggles under the "Snap to Features" toggle, so the association between the two was clear. You are probably right that it is probably unneeded, though, so I removed it and moved the snap sources toggles to their own section instead (suggested by Phil R. in sprint review).

|Before|After|
|:-:|:-:|
| Simulator Screenshot - iPhone 17 Pro - 2026-06-03 at 12 15 45 | Simulator Screenshot - iPhone 17 Pro - 2026-06-03 at 13 09 16 |

rolson (approved):

rolson (changes requested):
The overall enable snapping button isn't needed: 

Image

yo1995:
I'll make some changes based on today's feedback

rolson:
> I'll make some changes based on today's feedback

Can you be specific with what other changes? I don't think anything needs to be done with persistence yet. I think that will be at the feature editor or app level.

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/GeometryEditorToolbar.swift
```diff
@@ -241,6 +246,38 @@ private struct UndoButton: View {
...
+/// A button for presenting a settings view for configuring snapping.
+private struct SnapSettingsButton: View {
+    /// The model for the parent geometry editor toolbar.
+    @Environment(GeometryEditorToolbarModel.self) private var model
+    
+    /// A Boolean value indicating whether the settings view is presented.
+    @State private var isShowingSettings = false
+    
+    var body: some View {
+        if !model.geometryEditor.snapSettings.sourceSettings.isEmpty {
```
yo1995, line 257: With this condition, the snap settings button is only shown when `syncSourceSettings()` is called when the geometry editor starts. According to what we discussed, this might be the behavior we want.

I have discussed with Caleb and we agreed the current approach works. In case we change our minds in the future, I'll leave these thoughts below as a note.

---

I came across 2 edge cases:

1. Say if the user starts the geometry editor but only want the geometry being edited to snap to guides and itself, but not the feature's or graphic's geometries. For instance, to ensure the edited geometry has closed paths, or has 90 degrees angles.

- set `snapSettings.isEnabled = true`
- set `snapSettings.snapsToGeometryGuides = true` 

then the snap settings button won't show up, and the snapping cannot be configured.

2. The map simply doesn't have a snap source, i.e., doesn't have a feature layer or a graphics overlay. Maybe the user just want to sketch on the map without saving the geometry. Then they cannot configure snapping at all.
 
This would not be a problem at all if the toolbar is only used with the feature editor (because in the editing workflow, a feature must come from a feature layer in the map). However, as a standalone component, the toolbar might be used in a blank map without snap sources.

Are they too niche usecases?

Alternatively, if we don't enforce the `snapSettings.sourceSettings.isEmpty` in the UI here, but check for `geometryEditor.snapSettings.isEnabled = true`, then we'll need to have a "sync source settings" button in the snap settings sheet.

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/SnapSettingsView.swift
```diff
@@ -0,0 +1,237 @@
...
+                            """
+                    )
+                    .frame(maxWidth: .infinity, alignment: .leading)
+                }
+                .animation(.default, value: snappingIsEnabled)
+                .onChange(of: snappingIsEnabled) {
+                    settings.isEnabled = snappingIsEnabled
+                }
+                
+                if snappingIsEnabled {
```
yo1995: June 4th in the HC, Lucas commented that `snapSettings.isEnabled` was the result of a hierarchical design process, and might not be needed in the UI.

Related to the comment above, if we don't care about the 2 edge cases:

- snapping enabled, but only snap to geometry guides, not other geometries
- snapping enabled, but there isn't a snap source

then we can get rid of this toggle, and set `snapSettings.isEnabled = true` by default when the geometry editor starts, and use `snapSettings.snapsToFeatures` toggle to control the snapping. I'll apply this suggestion and see what people think.

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/SnapSettingsView.swift
```diff
@@ -0,0 +1,218 @@
...
+            overlay.id
+        default:
+            nil
+        }
+        
+        if let name, !name.isEmpty {
+            return name
+        } else {
+            return "\(type(of: source))"
+        }
```
yo1995, line 161:
```suggestion
        let name = switch source {
        case let layer as LayerContent:
            layer.name
        case let overlay as GraphicsOverlay:
            overlay.id
        default:
            ""
        }
        
        if !name.isEmpty {
            return name
        } else {
            return "\(type(of: source))"
        }
```
Since source with empty string name, and unknown source type both result in a type-of string, what about combining them and make it non-optional?

Will there be another type other than a layer or a graphics overlay, to conform to `SnapSource` protocol?

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/GeometryEditorToolbar.swift
```diff
@@ -60,6 +64,8 @@ struct GeometryEditorToolbar: View {
     ///   - style: The style that determines the toolbar’s appearance and layout.  A `nil` value
     ///   displays the toolbar's controls without built-in layout or styling.
     init(geometryEditor: GeometryEditor, style: Style? = .vertical) {
+        // Snapping is enabled by default to simplify the UI.
+        geometryEditor.snapSettings.isEnabled = true
```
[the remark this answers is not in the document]
yo1995: > an app that deliberately disables snapping cannot keep it off while using the toolbar

What do you think? Would this be a legit usecase?

rolson (approved):

Sources/ArcGISToolkit/Components/GeometryEditorToolbar/GeometryEditorToolbar.swift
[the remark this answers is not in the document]
rolson: App can disable the guides and feature snapping which is effectively the same thing

yo1995 (approved):

PR #1461, opened 2026-06-08
Introduce `WorldTrackingProvider`

philium:


Examples/Examples/WorldScaleExampleView.swift
```diff
@@ -48,6 +48,7 @@ struct WorldScaleExampleView: View {
     /// The location datasource that is used to access the device location.
     @State private var locationDataSource = SystemLocationDataSource()
     
+    @available(*, deprecated)
```
rolson: do we not want to switch this out to the new way of doing it?

philium: Are you referring to [this](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0522-source-warning-control.md)? We could do that, but it would need to be gated behind a compiler check as it is only available in Swift 6.4 (which comes with Xcode 27). Might not be worth it now as this is a fairly simple and straightforward solution. And we might (will?) end up refactoring this example to use `AppleWorldTracking`.

rolson: Sorry, no I meant why is this example using the deprecated way of initializing WorldScaleSceneView, why not refactor it to use AppleWorldTracking?

philium: Oh! Because, well, I mean, um…please excuse me while I go make another commit.

rolson (dismissed):

Examples/Examples/WorldScaleExampleView.swift
```diff
@@ -75,17 +86,21 @@ struct WorldScaleExampleView: View {
...
-            guard let initialLocation = await locationDataSource.locations.first(where: { @Sendable _ in true }) else { return }
+            guard let initialLocation = await provider.dataSource.locations.first(where: { @Sendable _ in true }) else { return }
             
             // Put a circle graphic around the initial location.
             let circle = GeometryEngine.geodeticBuffer(around: initialLocation.position, distance: 20, distanceUnit: .meters, maxDeviation: 1, curveType: .geodesic)
             graphicsOverlay.addGraphic(Graphic(geometry: circle, symbol: SimpleLineSymbol(color: .red, width: 3)))
             
             // Stop the location data source after the initial location is retrieved.
-            await locationDataSource.stop()
+            await provider.dataSource.stop()
```
des12437:
```suggestion
```
The provider datasource should not be stopped here since the AppleWorldTrackingCameraFeedView uses the provider datasource for location updates.

des12437 (approved):

rolson (approved):

philium:
I missed that there is at least one test that uses `WorldScaleSceneView`. I'll submit another PR to address the build failures that were introduced by this PR.

PR #1469, opened 2026-06-12
fe: Add snap sources tests

yo1995:
Add unit tests for the `GeometryEditorToolbar.SnapSources` OptionSet.

There is no unit test for the Feature Editor Toolbar due to its properties and style initializers being private.


CalebRas (approved):
Looks good! Just a couple of questions I am unsure about:
- Do we also need a test for `SnapSources.contains(source:)`?
- Do we need any UI tests for the snap settings button/view?

PR #1470, opened 2026-06-15
`FeatureEditor` - Stage basic implementation

CalebRas:
## Description

This PR adds a basic implementation of the Feature Editor component and the `FeatureFormView` changes needed to support it.
Closes `swift/issues/8195`.

### Supported Features

- Using an `ArcGISFeature` to display a `FeatureFormView` in an inspector and start a `GeometryEditor`.
- Editing attributes via the `FeatureFormView`.
- Editing geometry via the `GeometryEditor` and `FeatureEditorToolbar`.
- Editing while navigating through associations via the `FeatureFormView`.
- Saving/discarding geometry/attribute edits via the `FeatureFormView`.

To keep this PR small, this implementation does NOT support the following list of features:

<details><summary><b>Unsupported Features</b></summary><p>

- Content insets/viewpoint centering.
- Syncing snap rules.
- Geometry-aware feature form.  
- Non-spatial features.
- Feature Form View's Show on Map button.
- `Popup` or `FeatureForm` overloads for the modifier.
- A Docc markdown file. 

</p></details>  

### New Public API

```swift
View.featureEditor(_:geometryEditor:)
```

## How To Test

- Run the [FeatureEditorExample](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/tree/Caleb/FeatureEditor-Example/FeatureEditorExample) project in the [Caleb/FeatureEditor-Example](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/tree/Caleb/FeatureEditor-Example) branch.
- Tap on a feature to edit.
- Use the `FeatureFormView` and `FeatureEditorToolbar` to edit the attributes/geometry. 
- Use the form's "editing buttons" to end. 

## Screenshots

| | |
|:-:|:-:|
| **iPhone** | Screenshot iPhone 17 Pro 06-15-2026 at 3 09 32 PM |
| **iPad** | Screenshot iPad (A16) 06-15-2026 at 4 48 09 PM |
| **Mac Catalyst** | Screenshot 2026-06-15 at 4 50 32 PM |

PR #1474, opened 2026-06-17
`FeatureEditor` - Add example project

CalebRas:
## Description

This PR adds an example project for the Feature Editor component. The functionality is based on the current implementation of the Feature Editor and will be updated as features are added.
Closes `swift/issues/7408`.

## How To Test

- Run the project and tap "Open Map View".
- Tap on a feature to edit and use the Feature Editor to edit geometry/attributes.
- Use the form's "editing" buttons to end.

## Screenshots

| Device | ContentView | ExampleMapView |
|:-:|:-:|:-:|
| **iPhone** | Simulator Screenshot - iPhone 17 Pro - 2026-06-17 at 12 03 03 | Simulator Screenshot - iPhone 17 Pro - 2026-06-17 at 12 05 43 |
| **iPad** | Simulator Screenshot - iPad (A16) - 2026-06-17 at 12 03 48 | Simulator Screenshot - iPad (A16) - 2026-06-17 at 12 04 19 |
| **Mac Catalyst** | Screenshot 2026-06-17 at 12 08 35 PM | Screenshot 2026-06-17 at 12 08 11 PM |


rolson:
It doesn't need to be done in this PR, but if you can create an issue for this that would be good:

The Snap Settings button, I think, should be a "Snapping" button, not a "Settings" button. It takes you to snapping settings, not general feature editor or geometry editor settings.

image

rolson:
Show on Map doesn't work:

image

CalebRas:
> The Snap Settings button, I think, should be a "Snapping" button, not a "Settings" button. It takes you to snapping settings, not general feature editor or geometry editor settings.

Are you suggesting an update to the label text, icon, internal naming, or all of the above?

CalebRas:
> Show on Map doesn't work: 

That's expected since we haven't added viewpoint support yet. See "Unsupported Features" in https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/1470#issue-4669824726 and `swift/issues/8274`.

rolson:
> > The Snap Settings button, I think, should be a "Snapping" button, not a "Settings" button. It takes you to snapping settings, not general feature editor or geometry editor settings.
> 
> Are you suggesting an update to the label text, icon, internal naming, or all of the above?

Sorry. I meant icon. But yes, the label text should be "Snapping" and the internal name should also represent that I'd say.

Later down the road if this encompasses more than just snapping settings and is a true general settings then we can change it accordingly.

FeatureEditorExample/FeatureEditorExample/ExampleMapView.swift
```diff
@@ -0,0 +1,130 @@
...
+    }
+    
+    /// Loads the map using the required credentials.
+    private func loadMap() async throws {
+        // Note: Never hardcode login information in a production application.
+        // This is done solely for the sake of the example.
+        let credential = try await TokenCredential.credential(
+            for: map.url!,
+            username: "viewer01",
+            password: "I68VGU^nMurF"
```
rolson, line 95: Do we do this in other public code? Will the secret scanning we have in place pick this up? @njarecha

FeatureEditorExample/FeatureEditorExample/ExampleMapView.swift
```diff
@@ -0,0 +1,130 @@
...
+    var body: some View {
+        MapViewReader { mapViewProxy in
+            MapView(map: map)
+                .geometryEditor(featureEditorModel.geometryEditor)
+                .onDrawStatusChanged { drawStatus in
+                    // Enables the snap settings when the map view first completes drawing.
+                    guard !hasEnabledSnapSettings, viewState == nil, drawStatus == .completed else {
+                        return
+                    }
+                    enableSnapSettings(featureEditorModel.geometryEditor.snapSettings)
```
rolson: can this not just be called after the map loads? Why does drawing have to be completed for this to be called?

FeatureEditorExample/FeatureEditorExample/ExampleMapView.swift
njarecha: The current secret scanning rules are only for API keys. We need to come up with rule to scan plain username/passwords. Not sure CodeQL scan will pick up this but we do not run those scans other than actual toolkit source code.

CalebRas: Yes. This is done several places in the Samples, Toolkit Examples, and Toolkit UI Tests with the same credential ([example](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/v.next/Examples/Examples/UtilityNetworkTraceExampleView.swift#L119-L123)).

rolson: It seems like the service should just be public.

FeatureEditorExample/FeatureEditorExample/ExampleMapView.swift
CalebRas: It wasn't working after map load, but it turns out I just needed to load the layers, so I removed the `onDrawStatusChanged` and updated the naming a bit. Thanks for calling this out!

FeatureEditorExample/FeatureEditorExample/ExampleMapView.swift
CalebRas: I created the web map being used, but I set the sharing to organizational to reflect what is done with the other Naperville web maps. @yo1995, do you know why those maps usually require a login?

CalebRas:
> > > The Snap Settings button, I think, should be a "Snapping" button, not a "Settings" button. It takes you to snapping settings, not general feature editor or geometry editor settings.
> > 
> > 
> > Are you suggesting an update to the label text, icon, internal naming, or all of the above?
> 
> Sorry. I meant icon. But yes, the label text should be "Snapping" and the internal name should also represent that I'd say.
> 
> Later down the road if this encompasses more than just snapping settings and is a true general settings then we can change it accordingly.

Issue: `swift/issues/8286`

FeatureEditorExample/FeatureEditorExample/ExampleMapView.swift
yo1995: I commented in the feature editor chat.

FeatureEditorExample/FeatureEditorExample/ContentView.swift
```diff
@@ -0,0 +1,57 @@
...
+        )
+    }
+}
+
+// MARK: - FeatureEditorModel
+
+/// A view model containing objects needed for the feature editor.
+@MainActor
+@Observable
+final class FeatureEditorModel {
```
yo1995, line 52: A follow-up PR will update the `featureEditor` implementation to read the `geometryEditor` from the model in the environment, and this code will be moved, right? If so, I'm good with that.

FeatureEditorExample/FeatureEditorExample/ContentView.swift
```diff
@@ -0,0 +1,57 @@
...
+            }
+            .buttonStyle(.borderedProminent)
+            .navigationTitle("Feature Editor Example")
+            .navigationBarTitleDisplayMode(.inline)
+        }
+        .environment(featureEditorModel)
+        // The modifier that presents the feature editor.
+        // Note that because the feature editor uses a NavigationStack for internal
+        // navigation, the modifier must be placed outside of the app's NavigationStack.
+        .featureEditor(
```
yo1995, line 40: We can add this info to the doc-string for the modifier as well later, so that people understands the modifier has to be added to the top level `NavigationStack`.

FeatureEditorExample/FeatureEditorExample/ExampleMapView.swift
```diff
@@ -0,0 +1,119 @@
...
+        let initialExtent = Envelope(xRange: -9815340 ... -9815040, yRange: 5129550 ... 5130070)
+        map.initialViewpoint = Viewpoint(boundingGeometry: initialExtent)
+        
+        return map
+    }()
+    
+    /// The state of the view.
+    private enum ViewState: Equatable {
+        /// The view is being set up.
+        case settingUp
```
yo1995: Up to you, I feel like the `loadingMap` state (like an entry state) is only used when the app first launches, and will not occur again, so maybe not needed and can be handled via a separate task (non-id) modifier instead.

FeatureEditorExample/FeatureEditorExample/ExampleMapView.swift
```diff
@@ -0,0 +1,119 @@
...
+        ArcGISEnvironment.authenticationManager.arcGISCredentialStore.add(credential)
+        try await map.retryLoad()
+        
+        // Enables the geometry editor's snap settings and sources.
+        let snapSettings = featureEditorModel.geometryEditor.snapSettings
+        snapSettings.isEnabled = true
+        snapSettings.snapsToGeometryGuides = true
+        
+        await map.operationalLayers.load()
+        try? snapSettings.syncSourceSettings()
```
yo1995:
```suggestion
        try snapSettings.syncSourceSettings()
```
since the method already throws

FeatureEditorExample/FeatureEditorExample/ExampleMapView.swift
```diff
@@ -0,0 +1,119 @@
...
+            for: map.url!,
+            username: "viewer01",
+            password: "I68VGU^nMurF"
+        )
+        ArcGISEnvironment.authenticationManager.arcGISCredentialStore.add(credential)
+        try await map.retryLoad()
+        
+        // Enables the geometry editor's snap settings and sources.
+        let snapSettings = featureEditorModel.geometryEditor.snapSettings
+        snapSettings.isEnabled = true
```
yo1995:
```suggestion
```
The toolbar enables the overall snapping [by default](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/FeatureEditor/Sources/ArcGISToolkit/Components/GeometryEditorToolbar/GeometryEditorToolbar.swift#L80).

FeatureEditorExample/FeatureEditorExample/ContentView.swift
CalebRas: I am not sure I understand what you mean. Are you suggesting the `geometryEditor` should be passed to the `FeatureEditorToolbar` view instead of the `featureEditor` modifier?

FeatureEditorExample/FeatureEditorExample/ContentView.swift
CalebRas: I was planning on adding that to a markdown file that is referenced in the doc-string, similar to what was done for the [floating panel modifier](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/v.next/Sources/ArcGISToolkit/Components/FloatingPanel/FloatingPanelModifier.swift#L18-L19). See `swift/issues/8278`. Would you still like it in the doc-string?

yo1995: That's good, a standalone markdown is fine

FeatureEditorExample/FeatureEditorExample/ContentView.swift
yo1995: `featureEditorModel` is set in the environment by the example, and `geometryEditor` is set in the environment by `FeatureEditorModifier`.
Shall we make the changes in follow up PRs to only set `featureEditorModel` in the environment and read from it for the whole feature editor component?

CalebRas: We talked about this offline, and we don't need to make any example-level changes for this.

rolson (approved):

yo1995 (approved):

CalebRas:
Thanks for all feedback!

PR #1485, opened 2026-06-26
`FeatureEditor`: Merge `GeometryEditorToolbar` implemenation

CalebRas:
## Description

This PR merges the `GeometryEditorToolbar` implementation into the Feature Editor component.
Closes `swift/issues/8308`.

#### Architecture Changes

- Renamed `GeometryEditorToolbar` to `FeatureEditorToolbar`. (b76071a, 36ae01a)
- Removed `GeometryEditorToolbarModel` in favor of `FeatureEditorModel`. (8128efd, 3c10464) 
- Removed `GeometryEditorToolbar.SnapSources` & `GeometryEditorToolbar.snapSources(_:)` in favor of always only showing layer snap sources in `SnapSettingsView`. (15635b5)
- Removed `GeometryEditorToolbar.Style` in favor of `FeatureEditor.ToolbarStyle`. (dcb7939)
- Moved geometry editor stream monitoring to `FeatureEditorModel` when it was happening in multiple places. (417f345)
- Removed redundant "feature" `onChange` modifier by changing `FeatureEditorModel.feature` to `FeatureEditorModel.featureForm`. (79e27c6)

#### Minor Behavior Changes

- Fixed a bug where the inspector wouldn't dismiss when the `ExampleMapView` was closed. (7cbe888)
- Fixed a bug where the geometry editor stop and toolbar disappear would happen after the inspector dismissed, instead of at the same time. (7cbe888)
- Enabling `SnapSettings.isEnabled` by default now happens when `SnapSettingsButton` appears or the `geometryEditor.snapSettings` object changes instead of in `GeometryEditorToolbar.init`. (937afca, 058a70e)

## How To Test

Run the `FeatureEditorExample` and ensure it still works as expected.

**Note**: I didn't make any test changes since this PR is already large, so the `TestRunner` project currently doesn't build. (4250c26)


Sources/ArcGISToolkit/Components/FeatureEditor/Toolbar/FeatureEditorToolbar.swift
```diff
@@ -0,0 +1,236 @@
...
+                    bundle: .toolkitModule,
+                    comment: "A label for a button to redo the last undone geometry editor action."
+                )
+            } icon: {
+                Image(systemName: "arrow.uturn.forward")
+            }
+        }
+        .disabled(!canRedo)
+        .task(id: ObjectIdentifier(featureEditorModel.geometryEditor)) {
+            for await canRedo in featureEditorModel.geometryEditor.$canRedo {
```
yo1995, line 121: Should the `canRedo` be moved to `FeatureEditorModel` as well?

Sources/ArcGISToolkit/Components/FeatureEditor/Toolbar/FeatureEditorToolbar.swift
```diff
@@ -0,0 +1,236 @@
...
+                    bundle: .toolkitModule,
+                    comment: "A label for a button to delete the selected geometry editor element."
+                )
+            } icon: {
+                Image(systemName: "circle.badge.minus")
+            }
+        }
+        .disabled(!canDeleteSelectedElement)
+        .task(id: ObjectIdentifier(featureEditorModel.geometryEditor)) {
+            for await selectedElement in featureEditorModel.geometryEditor.$selectedElement {
```
yo1995, line 92: Should this be moved to `FeatureEditorModel` as well?

yo1995 (commented):
Overall looks great!

Sources/ArcGISToolkit/Components/FeatureEditor/Toolbar/FeatureEditorToolbar.swift
CalebRas: It could, but I'm not sure it should. I only moved stream observing to the model if the stream property was being observed both in the modifier and the view/toolbar, to keep as much off the model as possible. ($isStarted is only used in the toolbar, but it is in the model because it will be used in the modifier for making the form geometry aware very soon.) Would you still like me to move it?

yo1995: I see, that makes sense. No need to move.

rolson (approved):

yo1995 (approved):

PR #1486, opened 2026-06-26
fe: Update unit tests for FeatureEditor

yo1995:
Fix unit test failures due to API change in #1485 .

UI test fixes will be in the next PR.


Tests/ArcGISToolkitTests/FeatureEditorTests.swift
```diff
@@ -0,0 +1,79 @@
...
+    func modelInitializer() {
+        let model = FeatureEditorModel()
+        #expect(model.featureForm == nil)
+        #expect(model.geometryEditorGeometry == nil)
+        #expect(!model.geometryEditorCanUndo)
+        #expect(!model.geometryEditorIsStarted)
+    }
+    
+    @Test
+    func geometryEditorIsStartedStateChanges() async throws {
```
yo1995: This test is really testing 3 things:

- feature form identity remains the same throughout an editing session
- geometryEditorIsStarted state changes
- geometryEditorGeometry state changes

I want to break them into separate tests, but don't like the repeated overhead of creating the feature from a web map. Any suggestions?

Tests/ArcGISToolkitTests/FeatureEditorTests.swift
```diff
@@ -0,0 +1,76 @@
...
+
+@Suite("FeatureEditor Tests")
+@MainActor
+struct FeatureEditorTests {
+    @Test
+    func modelInitializer() {
+        let model = FeatureEditorModel()
+        #expect(model.featureForm == nil)
+        #expect(model.geometryEditorGeometry == nil)
+        #expect(!model.geometryEditorCanUndo)
```
CalebRas:
```suggestion
        #expect(!model.geometryEditorCanUndo)
        #expect(model.geometryEditorGeometry == nil)
```

Nitpick, but consider sorting these alphabetically like the properties declarations in the model.

Tests/ArcGISToolkitTests/FeatureEditorTests.swift
CalebRas: To me, this test really verifies that `monitorGeometryEditorStreams()` updates (or doesn't update) the model's properties correctly. Given that, I think its okay to leave it as one test. Maybe calling the test something like `monitorGeometryEditorStreams()` would help clarify that.

Tests/ArcGISToolkitTests/FeatureEditorTests.swift
```diff
@@ -0,0 +1,76 @@
...
+        let map = Map(item: PortalItem(
+            portal: .arcGISOnline(connection: .anonymous),
+            id: PortalItem.ID("acc027394bc84c2fb04d1ed317aac674")!
+        ))
+        try await map.load()
+        let layer = try #require(map.operationalLayers.first { $0.name == "Main" } as? FeatureLayer)
+        let parameters = QueryParameters()
+        parameters.addObjectID(3651)
+        let result = try await #require(layer.featureTable?.queryFeatures(using: parameters))
+        let feature = try #require(result.features().compactMap { $0 }.first as? ArcGISFeature)
```
CalebRas: Maybe use map instead since `features()` doesn't have optionals?

```suggestion
        let feature = try #require(result.features().map(\.self).first as? ArcGISFeature)
```

Or avoid mapping altogether?

```suggestion
        let feature = try #require(result.features().makeIterator().next() as? ArcGISFeature)
```

CalebRas (approved):
Looks good! Thanks for updating these.

CalebRas (approved):

des12437 (approved):

PR #1487, opened 2026-06-29
Fix FeatureForm UI test 7.1

mhdostal:
Update Test 7.1 to allow it to pass.


dfeinzimer (commented):
I gave this a closer look and it seems like the feature under test was edited. After restoring the attributes to the values the test expects it is passing again without these changes.

PR #1488, opened 2026-06-29
`FeatureEditor` - Add viewpoint centering support

CalebRas:
## Description

This PR adds support for automatic viewpoint centering to the Feature Editor. The viewpoint is centered when a `MapViewProxy` is provided and:
- The component starts editing.
- The user navigates to a new feature through an association.
- The user presses the "Show on Map" button on an association.

Closes `swift/issues/8274`.

### New Public API:
```swift
FeatureEditor(
    _ feature: Binding<ArcGISFeature?>,
    geometryEditor: GeometryEditor,
    mapViewProxy: MapViewProxy? = nil,
    toolbarStyle: ToolbarStyle? = .vertical
)
```

### Other Minor Changes

- Set content insets on the example's map view to prevent the viewpoint from being centered behind the sheet on iPhone.
- Added `FeatureEditorModel` resetting to prevent its state (like the `viewpointGeometry`) from carrying across multiple `FeatureEditor` view lifecycles. (201fd6a)
- Sorted the `FeatureEditorView.handleFormEditingEvent(_:)` switch cases and `FeatureEditor` properties. (f623aa5)

## How To Test

- Run the `FeatureEditorExample` and tap on a feature to start editing.
- Navigate to a new feature through associations or tap the "Show on Map" button on an association.
  -  **Note**: Most of the example's associations have the same location as the origin feature, so you will have to adjust the viewpoint from the default before trying to center the viewpoint to see anything happen.

https://github.com/user-attachments/assets/2049dd7d-1756-461b-a370-81fb9acc2082

### DIFF
--- a/FeatureEditorExample/FeatureEditorExample/ExampleMapView.swift
+++ b/FeatureEditorExample/FeatureEditorExample/ExampleMapView.swift
@@ -39,9 +39,20 @@ struct ExampleMapView: View {
     /// The point on the screen where the user tapped.
     @State private var tapPoint: CGPoint?
     
+    /// Insets that indicate the area obscured by the feature editor's UI to the map view.
+    private var contentInsets: EdgeInsets {
+        if UIDevice.current.userInterfaceIdiom == .phone, featureToEdit != nil {
+            // Accounts for the feature editor inspector and view on iPhone.
+            EdgeInsets(top: 0, leading: 0, bottom: 400, trailing: 75)
+        } else {
+            EdgeInsets()
+        }
+    }
+    
     var body: some View {
         MapViewReader { mapViewProxy in
             MapView(map: map)
+                .contentInsets(contentInsets)
                 .geometryEditor(geometryEditor)
                 .onSingleTapGesture { screenPoint, _ in
                     guard tapPoint == nil else { return }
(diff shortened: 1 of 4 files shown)


FeatureEditorExample/FeatureEditorExample/ExampleMapView.swift
```diff
@@ -39,9 +39,20 @@ struct ExampleMapView: View {
...
+            EdgeInsets(top: 0, leading: 0, bottom: 400, trailing: 75)
+        } else {
+            EdgeInsets()
+        }
+    }
+    
     var body: some View {
         MapViewReader { mapViewProxy in
             MapView(map: map)
+                .contentInsets(contentInsets)
```
CalebRas, line 55: @rolson If you want to try it out without content insets, comment this out and follow the testing instructions. Also, try it with the reticle tool.

rolson (approved):

yo1995 (approved):

FeatureEditorExample/FeatureEditorExample/ExampleMapView.swift
```diff
@@ -39,9 +39,20 @@ struct ExampleMapView: View {
     /// The point on the screen where the user tapped.
     @State private var tapPoint: CGPoint?
     
+    /// Insets that indicate the area obscured by the feature editor's UI to the map view.
+    private var contentInsets: EdgeInsets {
+        if UIDevice.current.userInterfaceIdiom == .phone, featureToEdit != nil {
```
philium, line 44: What about split screen or slide over on iPad? If those should be taken into account, then size classes should be checked instead of user interface idiom. That would model best practices.

CalebRas: When I tested it, content insets were not needed on iPad because the inspector presents next to the content rather than over it, as it does on iPhone. I updated `swift/issues/8341` to include this, though, so we can take another look when we make other viewpoint-centering improvements.

philium: Thanks!

PR #1489, opened 2026-07-01
UN Trace: Support function name

des12437:
Related to `swift/8293`

The `UtilityNetworkTrace` component displays function results with a label and value and now uses the function name for the label when available, otherwise, falls back to use the network attribute name combined with the function type.

Here is an example of the labels for function results before and after this change:

Before:
Before_Catalyst

After:
After_Catalyst

### DIFF
--- a/Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTrace.swift
+++ b/Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTrace.swift
@@ -476,21 +476,15 @@ public struct UtilityNetworkTrace: View {
                     ) {
                         if let selectedTrace = viewModel.selectedTrace {
                             ForEach(selectedTrace.functionOutputs, id: \.objectID) { item in
-                                HStack {
-                                    if !item.function.name.isEmpty {
-                                        Text(item.function.name)
-                                    } else {
-                                        Text(item.function.networkAttribute.name)
-                                    }
-                                    Spacer()
+                                LabeledContent {
                                     VStack(alignment: .trailing) {
                                         if item.function.name.isEmpty {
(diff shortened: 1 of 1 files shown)


mhdostal (dismissed):

Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTrace.swift
dfeinzimer: If you want to take the oppurtunity to modernize this area a bit, the `HStack` and `Spacer` should be replaceable with `LabeledContent`:

```swift
LabeledContent {
    VStack(alignment: .trailing) {
        if item.function.name.isEmpty {
            Text(item.function.kind.title)
                .font(.caption)
                .foregroundStyle(.secondary)
        }
        if let result = item.result as? Double {
            Text(result, format: .number)
        } else {
            Text(
                "Not Available",
                bundle: .toolkitModule,
                comment: "A trace function output result was not provided."
            )
        }
    }
} label: {
    if !item.function.name.isEmpty {
        Text(item.function.name)
    } else {
        Text(item.function.networkAttribute.name)
    }
}
```

dfeinzimer (dismissed):
Looks good. Just one suggestion, take or leave it.

Sources/ArcGISToolkit/Components/UtilityNetworkTrace/UtilityNetworkTrace.swift
des12437: Thanks for the suggestion. I applied this feedback and the UI looks as expected:


named_labeled_content

unnamed_labeled_content

dfeinzimer (approved):

mhdostal (approved):

des12437:
Thank you both.

PR #1491, opened 2026-07-03
`FeatureEditor` - Add snap settings syncing

CalebRas:
## Description

This PR adds automatic snap source setting syncing to the Feature Editor. Closes `swift/issues/8279`.

### Public API Changes

```swift
FeatureEditor(
    _ feature: Binding<ArcGISFeature?>,
    geometryEditor: GeometryEditor,
    map: Map?,
    mapViewProxy: MapViewProxy? = nil,
    toolbarStyle: ToolbarStyle? = .vertical
)
```

### Behavior Changes

- Snap settings are synced when:
  - The FE starts geometry editing.
  - A new feature is navigated to in the form.
  - The snap setting button is pressed.
- Snap rules are used if a map with an applicable utility network is passed into the `FeatureEditor` view.
- `SnapSettingsButton` no longer hides when there are no snap sources.
- There is a delay between when the inspector and toolbar appear, due to the increased async snap settings setup.

### Architecture Changes

- Moved editing start/stop logic into `FeatureEditorModel`: This was needed to support snap settings setup in both the modifier and the view.

## How To Test

- Run the `FeatureEditorExample` and tap on a feature to start editing.
- Tap on the snap settings button to configure snapping.
- Edit and start/stop the feature editor to ensure everything still works correctly.

**Note**: Unit tests for these changes will be added in another PR, so the `TestRunner` project currently doesn't build (`swift/issues/8354`).

### DIFF
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
@@ -85,7 +85,7 @@ public struct FeatureEditor: View {
     public init(
         _ feature: Binding<ArcGISFeature?>,
         geometryEditor: GeometryEditor,
-        map: Map? = nil,
+        map: Map?,
         mapViewProxy: MapViewProxy? = nil,
         toolbarStyle: ToolbarStyle? = .vertical
     ) {
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
@@ -42,9 +42,7 @@ import SwiftUI
 ///
 /// **Behavior**
 ///
-/// The toolbar is visible only while a feature's geometry is being edited. The
-/// snap settings button is shown only when the geometry editor has non-empty
-/// snap source settings.
+/// The toolbar is visible only while a feature's geometry is being edited.
 ///
 /// **Associated Types**
 ///
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import ArcGIS
-import OSLog
+internal import os
 import SwiftUI
 
 /// A feature editing component that provides both the geometry editing via a
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
@@ -96,6 +96,13 @@ public struct FeatureEditor: View {
         self.toolbarStyle = toolbarStyle
     }
     
+    /// A collection of object ids used to determine when to start editing.
+    /// This updates when the `feature` or `map` instances change.
+    private var startEditingIDs: [ObjectIdentifier] {
+        let objects: [AnyObject?] = [feature, map]
+        return objects.compactMap { $0.map(ObjectIdentifier.init) }
+    }
+    
     public var body: some View {
         FeatureEditorToolbar(style: toolbarStyle)
             .task(id: ObjectIdentifier(geometryEditor)) {
@@ -103,7 +110,7 @@ public struct FeatureEditor: View {
                 model.restartGeometryEditor()
                 await model.monitorGeometryEditorStreams()
             }
-            .task(id: EquatableBox(objects: [feature, map])) {
+            .task(id: startEditingIDs) {
                 if let feature {
                     do {
                         try await model.startEditing(rootFeature: feature, on: map)
@@ -136,15 +143,6 @@ public struct FeatureEditor: View {
         let expandedGeometry = viewpointGeometry.extent.withBuilder { $0.expand(by: 2) }
         let viewpoint = Viewpoint(boundingGeometry: expandedGeometry)
         await mapViewProxy.setViewpoint(viewpoint, duration: 0.5)
-    }
-    
-    /// A box that provides `Equatable` conformance for an array of objects using their identifiers.
-    private struct EquatableBox: Equatable {
-        private let objectIdentifiers: [ObjectIdentifier]
-        
-        init(objects: [AnyObject?]) {
-            objectIdentifiers = objects.compactMap { $0.map(ObjectIdentifier.init) }
-        }
     }
 }
 
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
@@ -109,7 +109,7 @@ public struct FeatureEditor: View {
                         try await model.startEditing(rootFeature: feature, on: map)
                     } catch {
                         Logger.featureEditor.error(
-                            "Error starting feature editor: \(String(describing: error))"
+                            "Error starting feature editor: \(error.localizedDescription)"
                         )
                     }
                 } else {
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
@@ -104,14 +104,16 @@ public struct FeatureEditor: View {
                 await model.monitorGeometryEditorStreams()
             }
             .task(id: EquatableBox(objects: [feature, map])) {
-                guard let feature else { return }
-                
-                do {
-                    try await model.startEditing(rootFeature: feature, on: map)
-                } catch {
-                    Logger.featureEditor.error(
-                        "Error starting geometry editor: \(String(describing: error))"
-                    )
+                if let feature {
+                    do {
+                        try await model.startEditing(rootFeature: feature, on: map)
+                    } catch {
+                        Logger.featureEditor.error(
+                            "Error starting feature editor: \(String(describing: error))"
+                        )
+                    }
+                } else {
+                    model.stopEditing()
                 }
             }
             .onChange(of: model.isPresented) {
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import ArcGIS
+import OSLog
 import SwiftUI
 
 /// A feature editing component that provides both the geometry editing via a
@@ -55,6 +56,8 @@ public struct FeatureEditor: View {
     @Binding private var feature: ArcGISFeature?
     /// A geometry editor used to edit the feature's geometry on an associated `MapView`.
     private let geometryEditor: GeometryEditor
+    /// The map that `feature` is part of, used to set up rule-based snapping.
+    private let map: Map?
     /// A proxy for performing map view operations.
     private let mapViewProxy: MapViewProxy?
     /// The style to apply to the toolbar's controls.
@@ -69,6 +72,10 @@ public struct FeatureEditor: View {
     ///   The Feature Editor is displayed when the value is non-`nil`.
     ///   - geometryEditor: A geometry editor used to edit the feature's
     ///   geometry on an associated `MapView`.
+    ///   - map: The map that `feature` is part of, used to set up rule-based
+    ///   snapping for the geometry editor. If `nil` is passed, or the feature
+    ///   is not part of a utility network contained in the map, snapping is
+    ///   set up without snap rules.
     ///   - mapViewProxy: A proxy used to set the viewpoint on an associated
     ///   `MapView`.
     ///   - toolbarStyle: The style that determines the toolbar's appearance and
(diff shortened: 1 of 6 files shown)


Sources/ArcGISToolkit/Components/FeatureEditor/Toolbar/FeatureEditorToolbar.swift
```diff
@@ -154,34 +154,26 @@ private struct SnapSettingsButton: View {
     /// A Boolean value indicating whether the settings view is presented.
     @State private var isShowingSettings = false
     
-    /// The snap settings for the feature editor model's geometry editor.
-    private var snapSettings: SnapSettings { featureEditorModel.geometryEditor.snapSettings }
-    
     var body: some View {
-        if !snapSettings.sourceSettings.isEmpty {
```
CalebRas, line 161: I decided to remove hiding the `SnapSettingsButton` when the snap source settings are empty because:
1. It is very unlikely that `sourceSettings` will be empty while the toolbar is showing now that we have automatic snap settings syncing.
2. Keeping this logic would require keeping track of the `sourceSettings` state outside of this view to prevent some cases where the button wouldn't appear, and the added complexity didn't seem worth it to me.

@rolson, let me know if you are not okay with this, and I can find another approach.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
```diff
@@ -14,18 +14,37 @@
 
 import ArcGIS
 import Observation
+import OSLog
 
 /// A data model that contains various properties that are needed for the
 /// feature editor and shared between the modifier and the view.
 @MainActor
 @Observable
 final class FeatureEditorModel {
```
CalebRas, line 23: @yo1995 Please let me know if there are any other things you think should be moved into the model to make the FE more testable.

Sources/ArcGISToolkit/Components/FeatureEditor/Toolbar/FeatureEditorToolbar.swift
rolson: Should be fine.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
```diff
@@ -115,6 +137,15 @@ public struct FeatureEditor: View {
         let viewpoint = Viewpoint(boundingGeometry: expandedGeometry)
         await mapViewProxy.setViewpoint(viewpoint, duration: 0.5)
     }
+    
+    /// A box that provides `Equatable` conformance for an array of objects using their identifiers.
+    private struct EquatableBox: Equatable {
```
yo1995: I'm hesitating a little bit about its name. See the `EquatableBox` generic in the Swift repo that is used to compare the content, rather than ID.

To me, this is more like an "BoxForCombinedIdentity", but it is used for satisfying `task(id:)`'s `Equatable` conformance.
So I guess it is OK.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
```diff
@@ -80,29 +85,46 @@ public struct FeatureEditor: View {
     public init(
         _ feature: Binding<ArcGISFeature?>,
         geometryEditor: GeometryEditor,
+        map: Map? = nil,
```
yo1995: No need to change because I think the current main story for feature editor is to edit a feature in a utility network.

I'm inclined to have 2 separate initializers, one without the map parameter used for non-UN features, and the other with the map parameter for features that have UN rule based snapping. Some pros and cons from my view

- Pros: 
  - we can more clearly document each initializer about their respective usage
  - if in the future some enhancements are made in core, that doesn't need us to use the utility network in the web map to configure snapping, we can deprecate this initializer and make an API change to the other, instead of deprecating the current one initializer to create a new one.
- Cons:
  - more duplicate code to maintain

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
```diff
@@ -80,29 +85,46 @@ public struct FeatureEditor: View {
...
+                model.restartGeometryEditor()
                 await model.monitorGeometryEditorStreams()
             }
+            .task(id: EquatableBox(objects: [feature, map])) {
+                if let feature {
+                    do {
+                        try await model.startEditing(rootFeature: feature, on: map)
+                    } catch {
+                        Logger.featureEditor.error(
+                            "Error starting feature editor: \(String(describing: error))"
```
yo1995: Other toolkit logger errors seem to use `localizedDescription` instead of `String(describing:)`, which may provide value to keep the toolkit fully localized.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
```diff
@@ -57,15 +85,114 @@ final class FeatureEditorModel {
...
+    /// - Parameters:
+    ///   - feature: The root feature to edit.
+    ///   - map: The map that `feature` is part of, used to set up rule-based snapping.
+    func startEditing(rootFeature feature: ArcGISFeature, on map: Map?) async throws {
+        stopEditing()
+        rootFeatureForm = FeatureForm(feature: feature)
+        
+        // Sets up the utility network so snap rules can be created.
+        if let map {
+            try? await map.load()
```
yo1995:
```suggestion
            try await map.load()
```

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
```diff
@@ -57,15 +85,114 @@ final class FeatureEditorModel {
...
+        
+        // Loads the feature's table if the geometry is nil so geometryType can be accessed.
+        // It is always nil otherwise.
+        if feature.geometry == nil, let table = feature.table {
+            try await table.load()
+        }
+        
+        // Sets up the snap rules and syncs the snap source settings.
+        if let utilityNetwork, let element = utilityNetwork.makeElement(arcGISFeature: feature) {
+            snapRules = try? await .rules(for: utilityNetwork, assetType: element.assetType)
```
yo1995, line 180:
```suggestion
            snapRules = try await .rules(for: utilityNetwork, assetType: element.assetType)
```

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
CalebRas: This optional throw is intentional. My thought was that we would still want to try to start the feature editor if the map fails to load since it is only needed for snap rules. Would you still like me to remove it?

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
CalebRas: This optional throw is also intentional. If creating snap rules fails, we want `model.snapRules` to be set to `nil` so the old rules don't get used in future snap source settings syncing.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
CalebRas: I agree that the we want the default functionality to be snapping with utility network rules if possible, so I would prefer to keep one initializer to help indicate that. Having one also follows the "Take advantage of defaulted parameters" [API design guideline](https://www.swift.org/documentation/api-design-guidelines/#conventions) a bit better, in my opinion. I don't feel strongly about it, though, so if you or @zkline101 still think it should be two, I'm cool with changing it.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
CalebRas: I agree the name isn't the best, so I decided to remove the type and replace it with a computed property to make it more readable. Let me know if you have any more feedback on that.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
yo1995: Yes, I'd like to remove it. If I pass in the map, I would expect the snap rules are applied. If the process silently fails, and falls back to calling the plain `syncSourceSettings()`, that's an unexpected outcome to me.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
yo1995: Sounds good. Worth adding a comment to explain the behavior on this line.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
zkline101: I'm coming into this a bit new with this component but I like keeping one init and documenting what it means when a nil and non-nil map are passed in (which you have done). It keeps it simple and we have a similar pattern in other components.

>  I think the current main story for feature editor is to edit a feature in a utility network.

If that is the case should we only have one init that requires a map?

yo1995: Thought a bit more on this. To me, if a function call's parameter has a `nil` default value, it typically means a broad outcome (such as "don't apply change", "don't filter", etc.), while non-`nil` parameter means a more specific outcome.

So, if we think `syncSourceSettings(rules:sourceEnablingBehavior:)` is a more specific variant of `syncSourceSettings()`, then I think a single init is better. If we think they are 2 diverging behaviors, then I like 2 separate inits.

> > I think the current main story for feature editor is to edit a feature in a utility network.
>
> If that is the case should we only have one init that requires a map?

Feature editor will eventually support editing both UN and non-UN features, so it has to support both. From what I can tell so far, editing with the snap rules applied is the more popular workflow. Its initializer being hidden behind a default parameter hinders its discoverability a tiny bit, but not a big deal.

zkline101: >  Its initializer being hidden behind a default parameter hinders its discoverability a tiny bit, but not a big deal.

Yeah, I do agree with that. What if we make it optional but don't give it a default value? That way the user needs to make an explicit decision.

yo1995: > What if we make it optional but don't give it a default value?

I low-key like this ⬆️, despite it adds one more line to the initializer call-site… 😶‍🌫️ it resembles the `TableTopSceneView .clippingDistance`, that a user needs to make a decision to choose which way they want.

zkline101: > Having one also follows the "Take advantage of defaulted parameters" [API design guideline](https://www.swift.org/documentation/api-design-guidelines/#conventions) a bit better, in my opinion.

I think it becomes clearer when you read both sentences:

>  Take advantage of defaulted parameters when it simplifies common uses. Any parameter with a single commonly-used value is a candidate for a default.

And from what Ting is saying, the commonly-used value is a valid map. So in this case a `nil` default value may not make sense.

CalebRas: Makes sense to me, so I removed the default parameter. Thanks for the discussion!

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
```diff
@@ -14,18 +14,37 @@
...
+    /// This maps `rootFeatureForm` to a Boolean value.
+    var isPresented: Bool {
+        get { rootFeatureForm != nil }
+        set {
+            guard !newValue else { return }
+            stopEditing()
+        }
+    }
+    /// The form currently presented in the feature editor's `FeatureFormView`.
+    /// This is non-`nil` after navigating to an association feature.
```
yo1995, line 40: Should it say: "This is non-`nil` after a new feature form is shown in the view.", as navigating back to the root feature form from an associated feature will still keep it non-`nil`?

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
```diff
@@ -36,6 +55,15 @@ final class FeatureEditorModel {
     /// A Boolean value indicating whether the geometry editor has started.
     private(set) var geometryEditorIsStarted = false
     
+    // MARK: Snapping Properties
+    
+    /// The snap rules for the `feature`, if applicable, used to sync snap source settings.
+    private var snapRules: SnapRules?
+    /// The `feature`'s utility network, if applicable, used to create snap rules.
+    private var utilityNetwork: UtilityNetwork?
```
yo1995, line 63: I think it is worth adding a short explanation to these 2 on when they are `nil`. While it may seem obvious now, in a few years when this code is revisited, people may need to read across a few files to figure out when is "if applicable".

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
```diff
@@ -152,43 +111,17 @@ private struct FeatureEditorView: View {
         case .discardedEdits(let willNavigate):
             // Restarts the geometry editor when the form footer discard button is pressed.
             guard !willNavigate else { break }
-            startGeometryEditor()
+            model.restartGeometryEditor()
         case .savedEdits(let willNavigate):
             // Closes the inspector when the form footer save button is pressed.
             guard !willNavigate else { break }
-            isPresented = false
+            model.isPresented = false
```
yo1995, line 118: I know we haven't decided on what to do with apply edits. Do you want to add a `model.stopEditing()` now or later? I'm fine with either.

yo1995 (commented):
Overall LGTM.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
CalebRas: Yes, good call!

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
CalebRas: I am not quite sure what you mean. Are you suggesting removing [the `stopEditing()` call](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/Caleb/FeatureEditor-SnapSettingsSync/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift#L36) that happens when `isPresented` is set to `false`?

yo1995: I forgot it was called in the property observer.

yo1995 (approved):

zkline101 (approved):
Code looks good just a few questions.

Can you update the code snippet in the PR description to reflect the current status of this PR?

> There is a delay between when the inspector and toolbar appear, due to the increased async snap settings setup.

Should we resolve this by putting a progress view on the settings button while it is getting set up?

CalebRas:
> Can you update the code snippet in the PR description to reflect the current status of this PR?

Done.

> > There is a delay between when the inspector and toolbar appear, due to the increased async snap settings setup.
> 
> Should we resolve this by putting a progress view on the settings button while it is getting set up?

That was my first thought as well, but Ryan suggested not starting the geometry editor or displaying the toolbar until the snap settings are set up. It makes sense because we don't want the toolbar displaying if the geometry editor isn't started and we don't want the geometry editor to start without everything set up, so that end-user doesn't start editing without all the functionality.

zkline101:
Ok I see. Sounds like we could make that work if we showed an empty toolbar with no buttons + progress view so they can't use anything yet but we are indicating that some work is going on.

But we can always do that later and it doesn't need to hold up this PR.

CalebRas:
> Ok I see. Sounds like we could make that work if we showed an empty toolbar with no buttons + progress view so they can't use anything yet but we are indicating that some work is going on.

Hmm, I kind of like that. The only issue I can see is that the toolbar doesn't show for every feature, only those that support geometries. We can only determine geometry support after the set up, so if we showed a progress view, the toolbar could appear while loading and then disappear. I am not sure that's a deal breaker, though. I'll let @rolson comment his thoughts on that when he gets back.

> But we can always do that later and it doesn't need to hold up this PR.

Sounds good. Thank you both for the feedback!

PR #1497, opened 2026-07-09
`FeatureEditor`: Fix presentation issues

CalebRas:
## Description

This PR fixes some Feature Editor presentation issues:
1. A bug where opening the snap settings sheet would dismiss the feature editor inspector when the horizontal size class was compact. On iPhone, this would cause some state, like the form view navigation, to reset. On iPad, this would prevent the sheet from being closed while the window was minimized in Stage Manager.
2. A bug where minimizing the window using iPad Stage Manager would stop the Feature Editor, which would then result in an empty inspector when the window was re-expanded.
3. The inspector width on Mac Catalyst, which was smaller than the default.

Closes `swift/issues/8276`.

## How To Test

Run the `FeatureEditorExample` and ensure the inspector and snapping sheet work as expected. If you want to verify the bug fixes, use the workflows in the gifs below.

### 1. Snapping Sheet Bug
|Before|After|
|:-:|:-:|
| Simulator Screen Recording - iPhone 17 Pro - 2026-07-10 at 13 12 46 | Simulator Screen Recording - iPhone 17 Pro - 2026-07-10 at 13 21 08 |

### 2. Stage Manager Bug
|Before|After|
|:-:|:-:|
| Simulator Screen Recording - iPad (A16) - 2026-07-10 at 13 17 08 | Simulator Screen Recording - iPad (A16) - 2026-07-10 at 13 19 58 |

### 3. Mac Catalyst Inspector Width
|Before|After|
|:-:|:-:|
| Screenshot 2026-07-10 at 12 19 09 PM | Screenshot 2026-07-10 at 12 20 28 PM |

### DIFF
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
@@ -50,6 +50,9 @@ private struct FeatureEditorModifier: ViewModifier {
                 )
                 .inspectorColumnWidth(ideal: 320)
                 .interactiveDismissDisabled()
+                .sheet(isPresented: $model.snapSettingsSheetIsPresented) {
+                    SnapSettingsView(settings: model.geometryEditor.snapSettings)
+                }
             }
             .environment(model)
     }
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
@@ -38,21 +38,22 @@ private struct FeatureEditorModifier: ViewModifier {
     func body(content: Content) -> some View {
         content
             .safeInspector(isPresented: $model.isPresented) {
-                // VStack is needed for presentation modifiers to be applied.
-                VStack(spacing: 0) {
-                    FeatureEditorFormView(isMinimized: selectedPresentationDetent == .bar)
-                }
-                .presentationBackgroundInteraction(.enabled)
-                .presentationContentInteraction(.scrolls)
-                .presentationDetents(
-                    [.bar, .medium, .large],
-                    selection: $selectedPresentationDetent
-                )
-                .inspectorColumnWidth(ideal: 320)
-                .interactiveDismissDisabled()
-                .sheet(isPresented: $model.snapSettingsSheetIsPresented) {
-                    SnapSettingsView(settings: model.geometryEditor.snapSettings)
-                }
+                FeatureEditorFormView(isMinimized: selectedPresentationDetent == .bar)
+                    .presentationBackgroundInteraction(.enabled)
+                    .presentationContentInteraction(.scrolls)
+                    .presentationDetents(
+                        [.bar, .medium, .large],
+                        selection: $selectedPresentationDetent
+                    )
+#if !targetEnvironment(macCatalyst)
+                // Needed to ensure the inspector presents at full width on iPad.
+                // This is not done on Mac Catalyst because 320 is smaller than its default.
+                    .inspectorColumnWidth(ideal: 320)
+#endif
+                    .interactiveDismissDisabled()
+                    .sheet(isPresented: $model.snapSettingsSheetIsPresented) {
+                        SnapSettingsView(settings: model.geometryEditor.snapSettings)
+                    }
             }
             .environment(model)
     }
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
@@ -171,7 +171,10 @@ private extension View {
         if UIDevice.current.userInterfaceIdiom == .phone {
             sheet(isPresented: isPresented, content: content)
         } else {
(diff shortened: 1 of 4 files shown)


Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
```diff
@@ -38,18 +38,22 @@ private struct FeatureEditorModifier: ViewModifier {
     func body(content: Content) -> some View {
         content
             .safeInspector(isPresented: $model.isPresented) {
-                // VStack is needed for presentation modifiers to be applied.
-                VStack(spacing: 0) {
-                    FeatureEditorFormView(isMinimized: selectedPresentationDetent == .bar)
```
CalebRas, line 43: The `VStack` is no longer needed now that we are using a sheet in `safeInspector`.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
```diff
@@ -168,7 +171,10 @@ private extension View {
         if UIDevice.current.userInterfaceIdiom == .phone {
             sheet(isPresented: isPresented, content: content)
         } else {
-            inspector(isPresented: isPresented, content: content)
+            // The inspector is given a constant to prevent it from setting isPresented to false
+            // when the window is minimized in iPad Stage Manager. Otherwise, the feature editor
+            // will stop editing and show an empty inspector when the window is re-expanded.
+            inspector(isPresented: .constant(isPresented.wrappedValue), content: content)
```
CalebRas, line 181: Ting, I know that we discussed stopping the FE when the window is minimized and then hiding the inspector when the window is re-expanded, but I didn't realize at the time how simple it would be to prevent the inspector from stopping the FE. If you would still like to go with the approach we discussed, it would look like:

```swift
inspector(
    isPresented: isPresented.wrappedValue ? isPresented : .constant(false),
    content: content
)
```

Image

yo1995: Ha, let me give it a try! 👀

CalebRas: @rolson, could I get your opinion on how we should approach this bug? The issue is that when the app window is minimized and re-expanded on iPad, an empty inspector is shown because the feature editor is stopped when the inspector is closed, but it can't restart itself when it opens again:

Simulator Screen Recording - iPad (A16) - 2026-07-10 at 13 17 08

### Possible Solutions

#### 1. Prevent the inspector from stopping the Feature Editor.
Cons: There is no way to stop the FE when the window is minimized.

Simulator Screen Recording - iPad (A16) - 2026-07-10 at 13 19 58


#### 2. Force the close the inspector window is re-expanded.
Cons:
- There is an animation of the inspector opening and re-closing.
- In-progress edits are lost.

Image


Ting and I are leaning toward # 2 because it doesn't seem great to not have a way to stop the FE (# 1). Any thoughts on which you would prefer?

rolson: I'd lean towards option 1 personally. Considering these two downsides:
- user not able to stop editing while app is minimized
- user loses edits if he minimizes app

I think the first is the least bad option. And I don't think it's that bad. Losing data, however, is not good.

CalebRas: Cool, I'll this as is then. Thanks!

yo1995, line 181: This have a small problem - if the app is launched in a shrunk compact width window on iPad, the inspector will show as a full screen cover sheet, until you maximize the window to full screen when the inspector shows on the side.

I tried passing the size class via `@Environment(\.isPortraitOrientation) var isPortraitOrientation` to the `safeInspector` to always use sheet on compact horizontal size class, but there seems to be some modal presentation issues.

Since the aforementioned problem will be a rare usecase, I will give it a pass, unless there is some easy way to solve it.

yo1995 (approved):

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
CalebRas: Ah, it turns out the VStack I removed is still needed to fix this. Should be good to go now.

rolson (approved):

yo1995 (approved):

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
CalebRas: Ting mentioned offline that he was still seeing this bug with my changes, but it looks like it is only reproducible on some iPad models (iPad Pro 13-inch (M5) and iPad mini (A17 Pro) for me) with iOS 26 and is not reproducible at all with iOS 27. Given that the issue is localized to specific device models, has been fixed in later versions, and is likely a SwiftUI bug, we will not be pursuing a fix.

CalebRas:
Thanks for the discussion on this!

PR #1498, opened 2026-07-09
fe: Address build and test failures

yo1995:
#1491 added snap rules and it breaks the UI test. The previous "Structure Line" layer disables snapping due to snap rules. Change the snap source toggle test to the "NapervilleElectricV5 - Dirty Areas" source at the top of the section.

snap-rules


CalebRas (approved):
Test passes for me. Thanks for doing this!

PR #1501, opened 2026-07-16
`FeatureFormView` - Refresh attachments when edits are discarded

dfeinzimer:
Apollo 1131

- When form edits are discarded:
  - Refreshes the attachments in each `AttachmentsFormElement`
  - Re-evaluates expressions

For manual testing, run the Test Runner with arguments: `-testCase testAttachmentsResetOnDiscard`.

Test run # 4908
Release test run # 3


Sources/ArcGISToolkit/Common/AttachmentsFeatureElementView.swift
```diff
@@ -40,6 +40,10 @@ struct AttachmentsFeatureElementView: View {
     @State private var isExpanded = true
     /// The last locally added attachment.
     @State private var lastAttachmentAdded: AttachmentModel?
+    /// If `embeddedFeatureFormViewModel` is set, this holds the last recorded value of
+    /// `FeatureForm.hasEdits`. If the value changes from `true` to `false` edits were discarded
+    /// and the attachments should be refreshed.
```
mhdostal, line 45: Is this ONLY when edits are discarded? What about when saved, `hasEdits` would go from `true` to `false` in that case, correct? Do the attachments get refreshed automatically when the edits are saved?

dfeinzimer: Yes - but there's some caching at play that I believe lowers the impact. Documentation on the underlying method states:
> Attachments are not re-fetched if they have already been fetched.

I can conceive a mechanism that would exclusively work when the Discard button is used in toolkit UI but there would still be a gap. If an API user has hidden the toolkit buttons and is using their own button to discard directly on the `FeatureForm` we'd have no way to track that event  at the toolkit level. We would need some type of new discarded event on `FeatureForm`. 

Given that gap and the caching I'm not sure it's worth the extra complication to ignore saves.

mhdostal: Ok, that makes sense. Thanks.

mhdostal (approved):

zkline101 (approved):

Sources/ArcGISToolkit/Common/AttachmentsFeatureElementView.swift
dfeinzimer: Opened Apollo 1818 to see if this can be improved in the future.

PR #1505, opened 2026-07-17
fe: Add error handling for setup

yo1995:
## Description

- Show a `ContentUnavailableView` in the inspector/sheet when the feature editor fails to set up.

cuv

- Added a `retryStartEditing` method for the "Try Again" button, and made subsequent adjustments to the model and tests.

- Added a unit test for the `retryStartEditing` method.

## Error Origin Synopsis

There are 4 main origins of errors for the editing workflow:

1. Setting up the geometry editing environment.
  - This involves loading the feature and its feature table. This is to ensure that the geometry editor can start.
  - If this step fails, we cannot edit the feature's geometry, which makes the feature editor useless. Thus, we block the user from using the feature editor.

2. Setting up snap rules when a `Map` is passed in. 
  - This involves loading the map, and loading the utility networks from the map. If either of these steps fails, snap rules cannot be set.
  - Because the expectation is that when a `Map` is passed to the feature editor, the snap rules will be created for the feature. Thus, if snap rules setup fails, we block the user from using the feature editor.

3. Syncing snap sources.
  - If the geometry editor on the map view is not the same instance being passed into the feature editor, syncing snap source will fail.
  - Because passing which geometry editor instance to the feature editor is an intentional decision, I didn't provide a "Try Again" for this error, and kept the current implementation, which shows no snap sources when it fails to sync snap sources.

4. When save edits in the feature form, `InvalidGeometryError` could be thrown. This error is presented in an alert by the feature form, so feature editor doesn't need to handle it.

## How To Ad-Hoc Test

Anywhere in the edited code, call `throw CancellationError()` and see if the `ContentUnavailableView` shows up as expected. Tap "Try Again" to see if the feature editor starts correctly.

### DIFF
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
@@ -201,6 +201,9 @@ final class FeatureEditorModel {
         // Loads the feature so canUpdateGeometry can be accessed. It is always
         // false otherwise.
         try await feature.retryLoad()
+        // No need to load the feature's table upfront if the we don't edit its
+        // geometry.
+        guard feature.canUpdateGeometry else { return }
         
         // Loads the feature's table if the geometry is nil so geometryType
         // can be accessed. It is always nil otherwise.
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
@@ -49,6 +49,8 @@ final class FeatureEditorModel {
     var snapSettingsSheetIsPresented = false
     /// The geometry used to set the viewpoint.
     var viewpointGeometry: Geometry?
+    /// A Boolean value indicating whether the model is loaded.
+    var isLoaded = false
     
     // MARK: Geometry Editor Properties
     
@@ -117,6 +119,8 @@ final class FeatureEditorModel {
         presentedFeatureForm = featureForm
         
         do {
+            defer { isLoaded = true }
+            try await loadFeature()
             try await setUpGeometryEditing()
         } catch {
             startEditingError = error
@@ -135,6 +139,8 @@ final class FeatureEditorModel {
         rootFeatureForm = FeatureForm(feature: feature)
         self.map = map
         do {
+            defer { isLoaded = true }
+            try await loadFeature()
             if let map {
                 // Only try to set up snap rules when a map is provided.
                 // Sets up the utility network so snap rules can be created.
@@ -155,6 +161,8 @@ final class FeatureEditorModel {
     func retryStartEditing() async {
         guard startEditingError != nil else { return }
         do {
+            defer { isLoaded = true }
+            try await loadFeature()
             if let map {
                 try await map.retryLoad()
                 await map.utilityNetworks.retryLoad()
@@ -188,6 +196,7 @@ final class FeatureEditorModel {
         
         snapRules = nil
         map = nil
+        isLoaded = false
     }
     
     /// Syncs the `geometryEditor.snapSettings`' source settings.
@@ -210,10 +219,9 @@ final class FeatureEditorModel {
         }
     }
     
-    /// Performs setup needed for geometry editing and starts the geometry editor if applicable.
-    private func setUpGeometryEditing() async throws {
+    /// Loads the feature and its table if needed to allow geometry editing.
+    private func loadFeature() async throws {
         guard let feature else { return }
-        
         // Loads the feature so canUpdateGeometry can be accessed. It is always false otherwise.
         try await feature.retryLoad()
         guard feature.canUpdateGeometry else { return }
@@ -223,7 +231,11 @@ final class FeatureEditorModel {
         if feature.geometry == nil, let table = feature.table {
             try await table.retryLoad()
         }
-        
+    }
+    
+    /// Performs setup needed for geometry editing and starts the geometry editor if applicable.
+    private func setUpGeometryEditing() async throws {
+        guard let feature else { return }
         // Sets up the snap rules and syncs the snap source settings.
         if let utilityNetwork, let element = utilityNetwork.makeElement(arcGISFeature: feature) {
             // Errors are ignored to set snapRules to nil if creation fails, so
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
@@ -38,6 +38,7 @@ final class FeatureEditorModel {
     }
     /// The form currently presented in the feature editor's `FeatureFormView`.
     /// This is non-`nil` after a new feature form is shown in the view.
+    @ObservationIgnored
     private var presentedFeatureForm: FeatureForm?
     /// The root feature form to edit with the feature editor.
     private(set) var rootFeatureForm: FeatureForm?
@@ -66,9 +67,11 @@ final class FeatureEditorModel {
     // MARK: Snapping Properties
     
     /// The map that contains the utility network being edited.
+    @ObservationTracked
     private var map: Map?
     /// The snap rules for the `feature`, used to sync snap source settings.
     /// This is non-`nil` when snap rules were successfully created using the `utilityNetwork`.
+    @ObservationTracked
     private var snapRules: SnapRules?
     /// The `feature`'s utility network used to create snap rules.
     /// This is non-`nil` when a map containing the feature's UN is used to start editing.
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
@@ -119,9 +119,9 @@ final class FeatureEditorModel {
         presentedFeatureForm = featureForm
         
         do {
-            defer { isLoaded = true }
             try await loadFeature()
             try await setUpGeometryEditing()
+            isLoaded = true
         } catch {
             startEditingError = error
             Logger.featureEditor.error(
@@ -139,7 +139,6 @@ final class FeatureEditorModel {
         rootFeatureForm = FeatureForm(feature: feature)
         self.map = map
         do {
-            defer { isLoaded = true }
             try await loadFeature()
             if let map {
                 // Only try to set up snap rules when a map is provided.
@@ -148,6 +147,7 @@ final class FeatureEditorModel {
                 await map.utilityNetworks.load()
             }
             try await setUpGeometryEditing()
+            isLoaded = true
         } catch {
             // Don't start editing whichever error occurs, and let user retry.
             startEditingError = error
@@ -161,7 +161,6 @@ final class FeatureEditorModel {
     func retryStartEditing() async {
         guard startEditingError != nil else { return }
         do {
-            defer { isLoaded = true }
             try await loadFeature()
             if let map {
                 try await map.retryLoad()
@@ -169,6 +168,7 @@ final class FeatureEditorModel {
             }
             try await setUpGeometryEditing()
             startEditingError = nil
+            isLoaded = true
         } catch {
             // Don't start editing whichever error occurs, and let user retry.
             startEditingError = error
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
@@ -140,9 +140,7 @@ final class FeatureEditorModel {
                 // Only try to set up snap rules when a map is provided.
                 // Sets up the utility network so snap rules can be created.
                 try await map.load()
-                for utilityNetwork in map.utilityNetworks {
-                    try await utilityNetwork.load()
-                }
+                await map.utilityNetworks.load()
             }
             try await setUpGeometryEditing()
         } catch {
@@ -160,9 +158,7 @@ final class FeatureEditorModel {
         do {
             if let map {
                 try await map.retryLoad()
-                if let utilityNetwork {
-                    try await utilityNetwork.retryLoad()
-                }
+                await map.utilityNetworks.retryLoad()
             }
             try await setUpGeometryEditing()
         } catch {
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
@@ -133,10 +133,9 @@ final class FeatureEditorModel {
     func startEditing(rootFeature feature: ArcGISFeature, on map: Map?) async {
         stopEditing()
         rootFeatureForm = FeatureForm(feature: feature)
-        
+        self.map = map
         do {
             if let map {
-                self.map = map
                 // Only try to set up snap rules when a map is provided.
                 // Sets up the utility network so snap rules can be created.
                 try await map.load()
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
@@ -73,13 +73,9 @@ final class FeatureEditorModel {
     /// The `feature`'s utility network used to create snap rules.
     /// This is non-`nil` when a map containing the feature's UN is used to start editing.
     private var utilityNetwork: UtilityNetwork? {
-        if let map, let feature {
-            return map.utilityNetworks.first { utilityNetwork in
-                utilityNetwork.makeElement(arcGISFeature: feature) != nil
-            }
-        } else {
-            return nil
-        }
+        guard let map, let feature else { return nil }
+        return map.utilityNetworks
+            .first(where: { $0.makeElement(arcGISFeature: feature) != nil })
     }
     
     // MARK: Methods
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
@@ -131,8 +131,9 @@ final class FeatureEditorModel {
         loadResult = await Result {
             try await loadFeature()
             if let map {
-                // Only tries to set up snap rules when a map is provided.
-                // Sets up the utility network so snap rules can be created.
+                // When a map is provided, it implies the user wants to set up
+                // rule-based snapping for the geometry editor. Loads the
+                // utility network so snap rules can be created.
                 try await map.load()
                 await map.utilityNetworks.load()
             }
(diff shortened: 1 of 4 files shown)


Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
```diff
@@ -100,31 +112,66 @@ final class FeatureEditorModel {
...
         
-        try await setUpGeometryEditing()
+        do {
+            try await setUpGeometryEditing()
+        } catch {
+            startEditingError = error
+            Logger.featureEditor.error(
+                "Error starting feature editor: \(error.localizedDescription)"
+            )
+        }
```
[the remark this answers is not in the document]
yo1995: `CancellationError` only happens when a different feature (or map) is passed in, at which point the feature editor UI will not be showing the old feature.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
```diff
@@ -61,12 +63,22 @@ final class FeatureEditorModel {
     
     // MARK: Snapping Properties
     
+    /// The map that contains the utility network being edited.
+    private var map: Map?
```
yo1995, line 68: Because the model previously didn't store the map, it cannot retry loading the map. I stored the map and make the UN computed based on the map.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
```diff
@@ -40,7 +40,29 @@ private struct FeatureEditorModifier: ViewModifier {
...
+                                Text(
+                                    "Failed to start editing",
+                                    bundle: .toolkitModule,
+                                    comment: "A title shown when feature editing could not start."
+                                )
+                            } icon: {
+                                Image(systemName: "exclamationmark.triangle")
+                            }
+                        } description: {
+                            Text(error.localizedDescription)
```
yo1995: Please comment on whether we show the actual error message to the end user, or a generic "Failed to start editing" message is good enough, i.e., having description or not for the CUV.

Tests/ArcGISToolkitTests/FeatureEditorModelTests.swift
```diff
@@ -154,6 +154,36 @@ struct FeatureEditorModelTests {
...
+        defer { monitorGeometryEditorStreamsTask.cancel() }
+        
+        let geodatabaseFile = try await TemporaryGeodatabaseFile()
+        let table = try await geodatabaseFile.geodatabase.makeTable(description: .points)
+        let geometry = Point(latitude: 2, longitude: 2)
+        let feature = try #require(table.makeFeature(geometry: geometry) as? ArcGISFeature)
+        let map = Map(spatialReference: nil)
+        
+        // Simulates map fails to load by creating a map with nil spatial reference.
+        await model.startEditing(rootFeature: feature, on: map)
```
yo1995: This is a bit hacky… though I can't think of a better way to cause `startEditing` to fail. Tried using `Task.cancel()` with no avail. Suggestions are welcome!

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
```diff
@@ -40,7 +40,36 @@ private struct FeatureEditorModifier: ViewModifier {
...
+                            } label: {
+                                Text.tryAgain
+                            }
+                            
+                            Button {
+                                model.stopEditing()
+                                model.isPresented = false
+                            } label: {
+                                Text.cancel
+                            }
```
yo1995: Do you want a `.buttonStyle(.bordered)` or `.buttonStyle(.borderedProminent)` style for these buttons? Which may look like below.

Screenshot 2026-07-20 at 10 32 02 AM

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
```diff
@@ -40,7 +40,36 @@ private struct FeatureEditorModifier: ViewModifier {
...
+                                Image(systemName: "exclamationmark.triangle")
+                            }
+                        } description: {
+                            Text(error.localizedDescription)
+                        } actions: {
+                            Button {
+                                Task { await model.retryStartEditing() }
+                            } label: {
+                                Text.tryAgain
+                            }
```
CalebRas: Is it possible to use a task modifier here instead, so the async operation gets canceled when the view disappears, and to disable the button, so the user can't spam it?

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
```diff
@@ -100,31 +112,63 @@ final class FeatureEditorModel {
...
         
-        // Sets up the utility network so snap rules can be created.
-        if let map {
-            try await map.load()
-            await map.utilityNetworks.load()
-            utilityNetwork = map.utilityNetworks.first { utilityNetwork in
-                utilityNetwork.makeElement(arcGISFeature: feature) != nil
+        do {
+            if let map {
+                self.map = map
```
CalebRas:
```suggestion
            self.map = map
            if let map {
```

This only sets the map if it is non-nil, so if the developer passes a map into the FE and then later passes in nil, the `model.map` won't be reset. Should we always set the map instead?

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
```diff
@@ -41,6 +41,8 @@ final class FeatureEditorModel {
     private var presentedFeatureForm: FeatureForm?
     /// The root feature form to edit with the feature editor.
     private(set) var rootFeatureForm: FeatureForm?
+    /// The error when starting editing fails.
+    private(set) var startEditingError: (any Error)?
```
CalebRas: Should `FeatureEditorModel.expectDefaultPropertyValues(sourceLocation:)` in `FeatureEditorModelTests.swift` be updated to include a check for this property?

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
CalebRas: My vote is that we show the error description, like you currently have it, so that the user has enough info to make an informed decision on what to do next. For example, they wouldn't want to retry if it fails due to no network connection, but they can't tell that is why it failed without the description.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
```diff
@@ -40,7 +40,36 @@ private struct FeatureEditorModifier: ViewModifier {
...
+                            
+                            Button {
+                                model.stopEditing()
+                                model.isPresented = false
```
CalebRas:
```suggestion
```
Setting `model.isPresented` just calls `stopEditing()`, so consider doing one or the other.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
```diff
@@ -40,7 +40,36 @@ private struct FeatureEditorModifier: ViewModifier {
...
+                            } label: {
+                                Text.cancel
+                            }
+                        }
+                    } else {
+                        FeatureEditorFormView(isMinimized: selectedPresentationDetent == .bar)
+                    }
```
CalebRas: When starting the FE currently, the form appears while the load is in progress, and then disappears when the error is thrown. This means that in a low-network environment, the user might actually be able to make some attribute edits before being blocked. Is it possible to show a `ProgressView` while loading instead? Maybe using a `Result`?

Tests/ArcGISToolkitTests/FeatureEditorModelTests.swift
CalebRas: I think this is a great solution, honestly.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
CalebRas: I think `borderedProminent` for try again and `bordered` for cancel would be good, since the buttons are a bit hard to see otherwise.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
yo1995: After discussion, we decided to adjust the order of loading. 

- First, we load the feature and the table
- Second, we load the map, utility network, and snap rules, when the map is provided

In this way, we don't unnecessarily delay the appearance of the feature form. Say we are editing a non-spatial feature.

The progress view will still be displayed and block the UI interactions before everything loads.

yo1995: Changes are in 5b4c954. We can discuss a bit more if this is an ideal UX.

rolson:
@philium please review in-lieu of me

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
```diff
@@ -100,31 +114,48 @@ final class FeatureEditorModel {
...
+            try await setUpGeometryEditing()
+        }
     }
     
     /// Starts an editing session for the given `feature`.
     /// - Parameters:
     ///   - feature: The root feature to edit.
     ///   - map: The map that `feature` is part of, used to set up rule-based snapping.
-    func startEditing(rootFeature feature: ArcGISFeature, on map: Map?) async throws {
+    func startEditing(rootFeature feature: ArcGISFeature, on map: Map?) async {
```
philium: FYI: Remember that "if the first argument forms part of a grammatical phrase, omit its label." Also, "Omit needless words. Every word in a name should convey salient information at the use site." For this method, I think just omitting `rootFeature` would be fine (`startEditing:(_:on:)`). But if you think `feature` needs to be included, but it outside the parentheses (`startEditingFeature(_:on:)`).

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
```diff
@@ -100,31 +114,48 @@ final class FeatureEditorModel {
...
+                await map.utilityNetworks.load()
             }
+            try await setUpGeometryEditing()
+        }
+    }
+    
+    /// Retries starting an editing session.
+    func retryStartEditing() async {
+        // Makes sure the previous load failed and sets the loadResult to nil.
+        guard case .failure = loadResult.take() else { return }
```
philium, line 147: Kudos for using `Optional.take()`!

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
```diff
@@ -34,13 +34,56 @@ private struct FeatureEditorModifier: ViewModifier {
...
+                            Text(error.localizedDescription)
+                        } actions: {
+                            Button {
+                                isRetrying = true
+                            } label: {
+                                Text.tryAgain
+                            }
+                            .task(id: isRetrying) {
+                                guard isRetrying else { return }
+                                defer { isRetrying = false }
```
philium: Could `isRetrying` change while waiting on `retryStartEditing()`? If so, you should check whether the task has been cancelled before updating the state. Probably a good idea to do that either way.

Tests/ArcGISToolkitTests/FeatureEditorModelTests.swift
```diff
@@ -154,6 +154,38 @@ struct FeatureEditorModelTests {
         }
     }
     
+    /// Verifies setup failure leads to loadResult failure, blocks editing,
+    /// and `retryStartEditing()` clears the error on success.
+    @Test
+    func startEditingErrorAndRetry() async throws {
```
philium: Perhaps a better name would be something like `retryStartEditingAfterFailure`?

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
yo1995: I've renamed the methods

- `startEditing(rootFeature:on:)` -> `startEditingFeature(_:on:)`
- `startEditing(newFeatureForm:)` -> `startEditingFeatureForm(_:)`

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
yo1995: I've added task canceled check. In theory, `isRetrying` shouldn't change while `retryStartEditing` is running, as the "Try Again" button is disabled.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
```diff
@@ -61,12 +63,24 @@ final class FeatureEditorModel {
...
     /// This is non-`nil` when a map containing the feature's UN is used to start editing.
-    private var utilityNetwork: UtilityNetwork?
+    private var utilityNetwork: UtilityNetwork? {
+        if let map, let feature {
+            return map.utilityNetworks.first { utilityNetwork in
+                utilityNetwork.makeElement(arcGISFeature: feature) != nil
+            }
+        } else {
+            return nil
+        }
```
philium:
```suggestion
        guard let map, let feature else { return nil }
        return map.utilityNetworks
            .first(where: { $0.makeElement(arcGISFeature: feature) != nil })
```

philium (commented):
Changes look good. Just one more suggestion.

philium (approved):

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
```diff
@@ -34,13 +34,61 @@ private struct FeatureEditorModifier: ViewModifier {
...
+                                // When the inspector is dismissed, the
+                                // feature editor will also stop editing.
+                                model.isPresented = false
+                            } label: {
+                                Text.cancel
+                            }
+                            .buttonStyle(.bordered)
+                        }
+                    case .none:
+                        ProgressView()
```
CalebRas: When the Progress View is showing, there is no way to dismiss the FE, which could be an issue in low network environments where the loading takes a long time. Maybe add a `DismissButton` to dismiss the FE and cancel loading? If you go with that, consider also updating the error view to use a dismiss button instead the cancel, so that it matches the Progress View and Feature Form View.

Tests/ArcGISToolkitTests/FeatureEditorModelTests.swift
```diff
@@ -154,7 +154,39 @@ struct FeatureEditorModelTests {
         }
     }
     
-    /// Verifies `startEditing(rootFeature:on:)` using a feature without a geometry.
+    /// Verifies setup failure leads to loadResult failure, blocks editing,
+    /// and `retryStartEditing()` clears the error on success.
+    @Test
+    func retryStartEditingAfterFailure() async throws {
```
CalebRas: Nitpick, but maybe move this so the test methods remain sorted alphabetically? Up to you.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
```diff
@@ -166,20 +194,23 @@ final class FeatureEditorModel {
...
+    /// Loads the feature and its table if needed to allow geometry editing.
+    private func loadFeature() async throws {
         guard let feature else { return }
+        // Loads the feature so canUpdateGeometry can be accessed. It is always
+        // false otherwise.
+        try await feature.retryLoad()
         
-        // Loads the feature so canUpdateGeometry can be accessed. It is always false otherwise.
-        try await feature.load()
-        guard feature.canUpdateGeometry else { return }
```
CalebRas, line 175: Was this removed intentionally? It seems like we wouldn't want unnecessarily to load the table if the feature's geometry can't be edited.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
```diff
@@ -100,31 +110,48 @@ final class FeatureEditorModel {
...
-            try await map.load()
-            await map.utilityNetworks.load()
-            utilityNetwork = map.utilityNetworks.first { utilityNetwork in
-                utilityNetwork.makeElement(arcGISFeature: feature) != nil
+        self.map = map
+        loadResult = await Result {
+            try await loadFeature()
+            if let map {
+                // Only tries to set up snap rules when a map is provided.
+                // Sets up the utility network so snap rules can be created.
```
CalebRas: Do these comments need updated now that `syncSnapSourceSettings()` isn't called here?

yo1995: I'll adjust the phrasing.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
yo1995: I'll add it back. It was somewhat intentional, but after more thoughts, it should be there.

There are 3 scenarios I can think of that a feature cannot update geometry:

- License.Lite. When having a lite license, secured non-ArcGIS Platform feature table cannot be edited. `feature.canUpdateGeometry` will be `true` in this case, so adding or not adding this check doesn't make a difference as the table will be loaded either way.
- Feature's geometry cannot be edited for other reasons, such as [it has an m-value](https://support.esri.com/en-us/knowledge-base/how-to-edit-feature-services-from-arcgis-server-10-5-x--000016652#:~:text=updates%20without%20m--,value,-.). In this case, the feature has a geometry, so we don't need the table to decide its geometry type.
- Non-spatial feature. In this case, we actually don't want to edit its geometry, so the no need to load the table to get the geometry type.

In all 3 cases, guarding `feature.canUpdateGeometry` don't have a negative impact, while can skip the unnecessary table load to save time, so good to add it back.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
```diff
@@ -34,13 +34,61 @@ private struct FeatureEditorModifier: ViewModifier {
...
+                            } label: {
+                                Text.tryAgain
+                            }
+                            .task(id: isRetrying) {
+                                guard isRetrying else { return }
+                                defer {
+                                    // Avoids state updates from cancelled tasks.
+                                    if !Task.isCancelled {
+                                        isRetrying = false
+                                    }
```
yo1995: Once the `retryStartEditing` method starts running and sets the `loadResult` to `nil`, the CUV is replaced by a ProgressView, and the task is canceled, so this part cannot be reached 🤔  Need to re-organize the order of stuff to see if this will work.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModifier.swift
yo1995: https://github.com/user-attachments/assets/d92b370d-fb7b-4f5d-ac97-26b68aa9b116

`DismissButton` added to both views.

CalebRas (approved):
Looks good! Thank you for making those changes

PR #1508, opened 2026-07-21
`FeatureFormView` - Fix attachment cell menus

dfeinzimer:
Apollo 1815
Release test runs # 4, 8

Fixes a bug where the first attachment in the list is always the one deleted or renamed no matter which one the user selected.

The core of the fix is replacing a combination of [`contextMenu(menuItems:)`](https://developer.apple.com/documentation/swiftui/view/contextmenu(menuitems:)) (which held the delete and rename buttons) and [`onTapGesture(count:perform:)`](https://developer.apple.com/documentation/swiftui/view/ontapgesture(count:coordinatespace:perform:)-21n4i) (which triggered the load action) with [`Menu.init(content:label:primaryAction:)`](https://developer.apple.com/documentation/swiftui/menu/init(content:label:primaryaction:)) where the primary action (single tap) now triggers loading.

The current setup was broken by #1283 which caused the entire carousel to start functioning as one big button. It went uncaught as there was no UI test that edited elements with multiple attachments. This PR also adds a case that does.

For manual testing, run the Test Runner with arguments: `-testCase testAttachmentDeletion`.

| Before | After |
|---|---|
|  |  |


mhdostal (approved):

dfeinzimer:
@chriswebb09, please review when you have some time, thanks!

chriswebb09 (approved):
LGTM

PR #1509, opened 2026-07-21
`FeatureEditor`: Viewpoint centering improvements

CalebRas:
## Description

This PR makes some improvements related to viewpoint centering in the Feature Editor. Closes `swift/issues/8341` & `swift/issues/7722`.

### Behavior Changes

- "Show on Map" features are briefly selected.
- Candidate features are selected while association creation view is showing.
- Viewpoint zooms to include the candidate feature and geometry editor geometry when association creation view appears.
- Geometry editor briefly hides whenever it's geometry intersects a selected feature. 
- Geometry editor fill symbol is semi-transparent.

### Other Changes

- Example viewpoint was updated to include a polygon for easier testing.
- `FeatureEditorFormView` was moved to its own file.
  - Actual `FeatureEditorFormView` changes can be seen in: https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/compare/5526539..Caleb/FE-ViewpointImprovements

## How To Test

- Run `FeatureEditorExample` and tap on a feature to start editing.
- Try using "Show on Map" and adding an association. (Follow workflow in videos below.)

### Show on Map / Add Association

|Before|After|
|:-:|:-:|
| https://github.com/user-attachments/assets/f160636e-630a-4e58-b012-9137853be709 | https://github.com/user-attachments/assets/1af97c9f-b073-4211-b5f1-657bc0c514e9 |

### Example Viewpoint

|Before|After|
|:-:|:-:|
| Simulator Screenshot - iPad (A16) - 2026-07-16 at 15 22 19 | Simulator Screenshot - iPad (A16) - 2026-07-16 at 15 21 52 |

### DIFF
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
@@ -131,19 +131,19 @@ public struct FeatureEditor: View {
     /// Sets the viewpoint using `model.viewpointGeometry` and the `mapViewProxy`.
     private func setViewpoint() async {
         guard let viewpointGeometry = model.viewpointGeometry else { return }
-        guard let mapViewProxy else {
+        if let mapViewProxy {
+            let expandedGeometry = viewpointGeometry.extent.withBuilder { $0.expand(by: 2) }
+            let viewpoint = Viewpoint(boundingGeometry: expandedGeometry)
+            await mapViewProxy.setViewpoint(viewpoint, duration: 0.5)
+            
+            // Prevents canceling an animation started while this task was running.
+            if !Task.isCancelled {
+                model.viewpointGeometry = nil
+            }
+        } else {
             // Clears viewpointGeometry to prevent setting the viewpoint if the proxy is set later.
             model.viewpointGeometry = nil
-            return
         }
-        
-        let expandedGeometry = viewpointGeometry.extent.withBuilder { $0.expand(by: 2) }
-        let viewpoint = Viewpoint(boundingGeometry: expandedGeometry)
-        await mapViewProxy.setViewpoint(viewpoint, duration: 0.5)
-        
-        // Prevents canceling an animation started while this task was running.
-        guard !Task.isCancelled else { return }
-        model.viewpointGeometry = nil
     }
 }
 
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
@@ -131,12 +131,19 @@ public struct FeatureEditor: View {
     /// Sets the viewpoint using `model.viewpointGeometry` and the `mapViewProxy`.
     private func setViewpoint() async {
         guard let viewpointGeometry = model.viewpointGeometry else { return }
-        defer { model.viewpointGeometry = nil }
-        guard let mapViewProxy else { return }
+        guard let mapViewProxy else {
+            // Clears viewpointGeometry to prevent setting the viewpoint if the proxy later is set.
+            model.viewpointGeometry = nil
+            return
+        }
         
         let expandedGeometry = viewpointGeometry.extent.withBuilder { $0.expand(by: 2) }
         let viewpoint = Viewpoint(boundingGeometry: expandedGeometry)
         await mapViewProxy.setViewpoint(viewpoint, duration: 0.5)
+        
+        // Prevents cancelling an animation started while this task was running.
+        guard !Task.isCancelled else { return }
+        model.viewpointGeometry = nil
     }
 }
 
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
@@ -132,7 +132,7 @@ public struct FeatureEditor: View {
     private func setViewpoint() async {
         guard let viewpointGeometry = model.viewpointGeometry else { return }
         guard let mapViewProxy else {
-            // Clears viewpointGeometry to prevent setting the viewpoint if the proxy later is set.
+            // Clears viewpointGeometry to prevent setting the viewpoint if the proxy is set later.
             model.viewpointGeometry = nil
             return
         }
@@ -141,7 +141,7 @@ public struct FeatureEditor: View {
         let viewpoint = Viewpoint(boundingGeometry: expandedGeometry)
         await mapViewProxy.setViewpoint(viewpoint, duration: 0.5)
         
-        // Prevents cancelling an animation started while this task was running.
+        // Prevents canceling an animation started while this task was running.
         guard !Task.isCancelled else { return }
         model.viewpointGeometry = nil
     }
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorFormView.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorFormView.swift
@@ -25,8 +25,14 @@ struct FeatureEditorFormView: View {
     /// the geometry editor.
     @Environment(FeatureEditorModel.self) private var model
     
+    /// The current editing event from the form view. This is non-`nil` while
+    /// the event is being processed by this view.
+    @State private var editingEvent: EquatableEditingEvent?
     /// The form currently presented in the `FeatureFormView`.
     @State private var presentedFeatureForm: FeatureForm?
+    /// The feature currently selected on the map. This is non-`nil` when
+    /// adding an association or "Show on Map" is pressed in the form view.
+    @State private var selectedFeature: ArcGISFeature?
     
     /// A closure that saves geometry edits to the form's feature. This is
     /// non-`nil` only when there are edits, so the `FeatureFormView`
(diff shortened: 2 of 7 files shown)


Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorFormView.swift
```diff
@@ -0,0 +1,286 @@
...
+    
+    /// Clears `selectedFeature` and unselects it on its layer.
+    private func clearSelectedFeature() {
+        guard let selectedFeature = selectedFeature.take(),
+              let layer = selectedFeature.table?.layer as? FeatureSelectableLayer else {
+            return
+        }
+        
+        layer.unselectFeature(selectedFeature)
+    }
```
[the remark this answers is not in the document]
CalebRas: It is defined in Swift.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorFormView.swift
```diff
@@ -0,0 +1,286 @@
...
+// Only 2D layers are extended since the Feature Editor is map-specific.
+extension AnnotationLayer: FeatureSelectableLayer {}
+extension DimensionLayer: FeatureSelectableLayer {}
+extension FeatureLayer: FeatureSelectableLayer {}
+extension OrientedImageryLayer: FeatureSelectableLayer {}
+
+// MARK: On Animation Change Modifier
+
+/// A view modifier that performs an action when the animation data of a given value changes.
+private struct OnAnimationChangeModifier<Value: VectorArithmetic>: @MainActor AnimatableModifier {
```
yo1995:
```suggestion
private struct OnAnimationChangeModifier<Value: VectorArithmetic>: @MainActor Animatable & ViewModifier {
```
[AnimatableModifier](https://developer.apple.com/documentation/swiftui/animatablemodifier) is deprecated.

yo1995: This is a different approach I came up with after I read the doc for [`Animatable`](https://developer.apple.com/documentation/swiftui/animatable).
I got rid of the `let value: Value` as I thought it serves the same purpose as the `animatableData`. It also alleviates the main-actor isolation on the type by saying the value is sendable (which in our case, is a Float that we know is sendable).
It doesn't make a huge difference, and I do like your version's synthesized initializer's callsite, so up to you.

```swift
/// A view modifier that performs an action when the animation data of a given value changes.
private struct OnAnimationChangeModifier<Value: Sendable & VectorArithmetic>: Animatable & ViewModifier {
    /// The value's animation data that is observed for changes.
    var animatableData: Value {
        didSet {
            action(animatableData)
        }
    }
    
    /// The action to perform when the animation data changes.
    let action: (_ animatableData: Value) -> Void
    
        content
    }
}
```

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorFormView.swift
```diff
@@ -0,0 +1,286 @@
...
+        layer.selectFeature(feature)
+        selectedFeature = feature
+        
+        // Briefly hides the geometry editor so the user can still locate the
+        // selected feature when it's covered by the geometry editor symbology.
+        guard let editorGeometry = model.geometryEditorGeometry,
+              !editorGeometry.isEmpty,
+              let featureGeometry = feature.geometry,
+              // Editor geometry is buffered to account for the vertex and line symbols.
+              let bufferedGeometry = GeometryEngine.buffer(around: editorGeometry, distance: 10),
```
yo1995, line 185: Had a discussion with Caleb. The purpose of this buffer is to account for the geometry editor's geometry covering the feature's geometry. Think of a big red polygon covering a point. 

https://github.com/user-attachments/assets/df7d3624-eba1-438e-b7d5-53b5a863585e

While the buffer could be unreasonably big if the unit of the geometry editor's geometry is in a spatial reference like WGS 84, it nonetheless solve the problem, which the geometry editor's geometry will be hidden due to it covers the feature's geometry.

Since this idea works for the most part, I'm OK with the magic number here. Alternatively, we can `geodeticBuffer` the geometry, or expand the geometry to a certain factor, to avoid making it too big, but it still requires a magic number.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorFormView.swift
```diff
@@ -0,0 +1,286 @@
...
+              GeometryEngine.isGeometry(featureGeometry, intersecting: bufferedGeometry) else {
+            throw GeometryEditorError.notHidden
+        }
+        
+        model.geometryEditor.clearSelection()
+        geometryEditorOpacity = 0.1
+        try? await Task.sleep(for: .seconds(1))
```
yo1995: The sleep duration can be a private constant property, as it is used in more than one place.

yo1995 (commented):
Overall looks good

rolson:
@philium please review

Sources/ArcGISToolkit/Components/FeatureEditor/Toolbar/ToolPicker.swift
```diff
@@ -91,19 +91,28 @@ private enum Tool: Hashable {
...
+            tool = VertexTool()
         case .vertexReticle:
-            return ReticleVertexTool()
+            tool = ReticleVertexTool()
         }
+        
+        // Makes the fill symbol semi-transparent to avoid obscuring the map beneath polygons.
+        if let fillSymbol = tool.style.fillSymbol as? FillSymbol {
+            fillSymbol.color = fillSymbol.color.withAlphaComponent(0.5)
+            tool.style.fillSymbol = fillSymbol
```
philium: Since `Symbol` (the type of `GeometryEditorStyle.fillSymbol`) is a class, there is no need to assign `fillSymbol` back, correct?


```suggestion
```

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift
```diff
@@ -137,12 +137,19 @@ public struct FeatureEditor: View {
...
+            return
+        }
         
         let expandedGeometry = viewpointGeometry.extent.withBuilder { $0.expand(by: 2) }
         let viewpoint = Viewpoint(boundingGeometry: expandedGeometry)
         await mapViewProxy.setViewpoint(viewpoint, duration: 0.5)
+        
+        // Prevents canceling an animation started while this task was running.
+        guard !Task.isCancelled else { return }
+        model.viewpointGeometry = nil
```
philium: I strongly recommend using an if…else statement instead of a guard statement since forward progress is being made either way.


```suggestion
        if let mapViewProxy {
            let expandedGeometry = viewpointGeometry.extent.withBuilder { $0.expand(by: 2) }
            let viewpoint = Viewpoint(boundingGeometry: expandedGeometry)
            await mapViewProxy.setViewpoint(viewpoint, duration: 0.5)
            
            // Prevents canceling an animation started while this task was running.
            if !Task.isCancelled {
                model.viewpointGeometry = nil
            }
        } else {
            // Clears viewpointGeometry to prevent setting the viewpoint if the proxy is set later.
            model.viewpointGeometry = nil
        }
```

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorFormView.swift
```diff
@@ -0,0 +1,286 @@
...
+        return {
+            guard let geometry = model.geometryEditorGeometry, geometry.sketchIsValid else {
+                throw GeometryEditorError.invalidGeometry
+            }
+            model.feature?.geometry = geometry
+        }
+    }
+    
+    var body: some View {
+        if let rootFeatureForm = model.rootFeatureForm {
```
philium: It seems like the parent view should unwrap this and use it to initialize this view. Without it, this view is useless, is it not? Better to not create the view at all if there is no root feature form to display.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorFormView.swift
```diff
@@ -0,0 +1,286 @@
...
+                    guard let editingEvent else { return }
+                    await handleEditingEvent(editingEvent.event)
+                    
+                    // Prevents clearing an editing event set while this task was running.
+                    guard !Task.isCancelled else { return }
+                    self.editingEvent = nil
+                }
+                .task(id: presentedFeatureForm.map(ObjectIdentifier.init)) {
+                    guard let presentedFeatureForm else { return }
+                    defer { self.presentedFeatureForm = nil }
```
philium: Similar with `task(id: editingEvent)`, should this only be set after the do block and after verifying that the task hasn't been cancelled?

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorFormView.swift
```diff
@@ -0,0 +1,286 @@
...
+            guard !willNavigate else { break }
+            model.isPresented = false
+        case .showOnMapRequested(let feature):
+            await showFeatureOnMap(feature)
+        }
+    }
+    
+    /// Zooms to and selects an association candidate feature on the map.
+    /// - Parameter candidate: The `UtilityAssociationFeatureCandidate` to show.
+    private func showCandidateOnMap(_ candidate: UtilityAssociationFeatureCandidate) async {
```
philium: This reads as though the parameter is a map, not a candidate. I wonder if `onMap` could be removed? If not then perhaps `showOnMap(candidate:)`?

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorFormView.swift
```diff
@@ -0,0 +1,286 @@
...
+            candidateGeometry,
+            geometryEditorGeometry
+        )
+        
+        try? await selectFeature(candidate.feature)
+    }
+    
+    /// Zooms to and briefly selects a feature on the map.
+    /// - Parameter feature: The `ArcGISFeature` to show.
+    private func showFeatureOnMap(_ feature: ArcGISFeature) async {
```
philium: Similar to showing a candidate, either `showFeature(_:)` or `showOnMap(feature:)`.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorFormView.swift
```diff
@@ -0,0 +1,286 @@
...
+        guard feature.geometry?.isEmpty == false else { return }
+        
+        model.viewpointGeometry = feature.geometry
+        
+        do {
+            try await selectFeature(feature)
+        } catch {
+            // Sleeps if the geometry editor wasn't hidden so there is still a
+            // delay before the selection is cleared.
```
philium: Why is a delay necessary?

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorFormView.swift
```diff
@@ -0,0 +1,286 @@
...
+        geometryEditorOpacity = 0.1
+        try? await Task.sleep(for: .seconds(1))
+        geometryEditorOpacity = 1
+    }
+}
+
+// MARK: - Helpers
+
+/// A wrapper for `FeatureFormView.EditingEvent` that conforms to `Equatable`.
+private struct EquatableEditingEvent: Equatable {
```
philium: I'm curious why `FeatureFormView.EditingEvent` couldn't just publicly conform to `Equatable`?

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorFormView.swift
```diff
@@ -0,0 +1,286 @@
...
+    
+    var errorDescription: String? {
+        switch self {
+        case .invalidGeometry:
+            String(
+                localized: "The geometry is invalid. It must be corrected before saving.",
+                bundle: .toolkitModule,
+                comment: "An error message shown when trying to save an invalid geometry."
+            )
+        default:
```
philium: I suggest being explicit about which case this is.


```suggestion
        case .notHidden:
```

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorFormView.swift
```diff
@@ -0,0 +1,286 @@
...
+            nil
+        }
+    }
+}
+
+// MARK: Feature Selectable Layer
+
+/// A layer that can select and unselect a feature.
+private protocol FeatureSelectableLayer {
```
philium: "Protocols that describe what something is should read as nouns." If you want this to represent a layer, then go all the way and specify that this can only be a layer: `protocol FeatureSelectableLayer: Layer { ... }`. But if that isn't necessary, then "Protocols that describe a capability should be named using the suffixes." I suggest `FeatureSelecting`.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorFormView.swift
CalebRas: @dfeinzimer, do you happen to remember why `FeatureFormView.EditingEvent` doesn't conform to `Equatable`?

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorFormView.swift
CalebRas: Would using `geodeticBuffer` be more ideal for this?

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorFormView.swift
CalebRas: I think I previously chose to unwrap it here since it seemed a bit odd to have a `rootFeatureForm` property on the view if it also has access to the `model.rootFeatureForm`, but I agree it would be better to unwrap at the parent view, so I moved it.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorFormView.swift
CalebRas: The goal is to have the feature selected for 1 second. That selection delay usually happens in `selectFeature(_:)` when the geometry editor opacity is also dimmed for one second. If the geometry editor cannot be dimmed for some reason, then that delay won't happen, so `selectFeature(_:)` throws to let `showFeatureOnMap(_:)` know that it has to perform the delay, which happens here. Any suggestions on how to make that more clear?

CalebRas:
I rebased this branch on the latest changes in `FeatureEditor`.

yo1995 (approved):

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorFormView.swift
dfeinzimer: There was likely just no need for it. If you need it now feel free to add it.

CalebRas:
@philium, could you take another look at this when you get a minute, please?

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorFormView.swift
philium: Perhaps the comment could be made clearer?

philium (approved):

CalebRas:
Thank you both for the feedback!

PR #1503, opened 2026-07-17
`FeatureFormView` - Add upcoming strings

dfeinzimer:
Pulls the new strings from #1437 to help begin the translation process for 300.1.

In #1437 this extension will be removed and the implementations will be moved into position.

PR #1519, opened 2026-07-30
Merge final translations for 300.1

dfeinzimer:
Swift 8408

### DIFF
--- a/Sources/ArcGISToolkit/Resources/ar.lproj/Localizable.strings
+++ b/Sources/ArcGISToolkit/Resources/ar.lproj/Localizable.strings
@@ -68,6 +68,15 @@ content on a remote host. The variable is the host that prompted the challenge. 
 /* A label in reference to utility associations elements contained within a popup. */
 "Associations" = "اقترانات";
 
+/* A message indicating the element requires at least one attachment. */
+"At least %lld attachment is required." = "بيت_At least %lld attachment is required____________________لاحقة.";
+
+/* A message indicating the element requires multiple attachments. */
+"At least %lld attachments are required." = "بيت_At least %lld attachments are required_____________________لاحقة.";
+
+/* A label for a button to choose a file from the user's files. */
+"Attach File" = "بيت_Attach File____________لاحقة";
+
 /* A label for the "Attachment" utility association kind. */
 "Attachment" = "مرفق";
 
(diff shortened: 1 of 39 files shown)


des12437 (approved):

Sources/ArcGISToolkit/Resources/ar.lproj/Localizable.strings
```diff
@@ -150,16 +159,25 @@ Cancel your current action to close/hide a sheet or pop-up. */
...
 
 /* A label for a button to choose a photo or video from the user's photo library. */
-"Choose From Library" = "اختيار من المكتبة";
+"Choose Photo or Video" = "اختيار صورة أو فيديو";
+
+/* A label for a button to choose a photo from the user's photo library. */
+"Choose Photo" = "اختيار صورة";
+
+/* A label for a button to choose a photo or video from the user's photo library. */
+"Choose Photo or Video" = "اختيار صورة أو فيديو";
```
pgruenler, line 173: It looks like lines 168 to 172 are a duplicate of lines 162 to 166, is that correct? That looks to be the case in every file.

dfeinzimer: TLDR: Yes it is correct, harmless and accurately reflects the state of the `v.next` branch when this round of strings were sent for translation. They're also already set to be deleted in the next round.

---

The keys were preexisting in extensions on `String` but #1437 changed them to be generated as extensions on `Text`. It was unknown how quickly #1437 was going to be merged so #1503 added all of the new keys from #1437 in a focused PR so that the translations process could begin ASAP. Therefore, there was a brief period of time where the keys existed twice. `xcodebuild -exportLocalizations` doesn't de-duplicate keys and this is an artifact of that.

The older `String` versions have since been removed and as part of the next release, when the English strings file is re-generated the duplicated entries will drop off and subsequently be auto-removed from each strings file during the next round of translations. You can check this yourself now by running the [`update-translations`](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/v.next/Scripts/update-translations) script and seeing them deleted in the generated English strings file.

Here's photo evidence the duplicated keys are still resolved correctly:
Simulator Screenshot - swift-test-ipad (26 1) - 2026-07-30 at 15 23 36

pgruenler (approved):

PR #1526, opened 2026-07-31
Offline: Refactor set download status

des12437:
Related to `swift/8438`.

Refactors how the `.downloaded` status is set. Previously the status could be set to `.downloaded` in both methods `updateDownloadStatus(for:)` and `loadAndUpdateMobileMapPackage(mmpk:)`, causing a race condition occasionally in the tests that verify the map is not nil once the model status is `.downloaded` since `updateDownloadStatus(for:)` sets the download status and `loadAndUpdateMobileMapPackage(mmpk:)` sets the map value after.

Verified the offline component download status works as expected with this change and the tests pass locally. This PR should be merged after the release to verify this PR fixes the flakey tests.


CalebRas (approved):
Looks good and tests pass for me.

philium (approved):

PR #1535, opened 2026-08-13
Revert "Enable `InternalImportsByDefault`" in the toolkit package

yo1995:
## Description

Reverts Esri/arcgis-maps-sdk-swift-toolkit#1532 . This upcoming feature causes a build problem for the toolkit in Xcode 26, while works in Xcode 27, so we pull it out for now until we bump the toolchain version.

I only reverted the changes in the toolkit package, 97b8662 , and not the changes in the example projects as they aren't built by `ArcGISBuild`.

## How to Test

Make sure the package and example project still build with no warnings.

PR #1532, opened 2026-08-06
Enable `InternalImportsByDefault`

CalebRas:
### Description

This PR enables the `InternalImportsByDefault` upcoming Swift feature in the package (97b8662) and sub-projects (4a8d752). Whitespace changes are from re-indenting the files after making changes.

### Linked Issue(s)

- Part of https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/1412.

### How To Test

Ensure the Examples project builds without issues. (Missing public imports will show as errors, and unneeded public imports will show as warnings.)

PR #1538, opened 2026-08-14
Fix build error in ArcGISToolkitTests

dfeinzimer:
Reverts a change from #1532 that was missing in #1535

Broken test run on v.next: # 5034
Fixed test run: # ~~5035~~ 5036

### DIFF
--- a/Tests/ArcGISToolkitTests/OnDemandMapModelTests.swift
+++ b/Tests/ArcGISToolkitTests/OnDemandMapModelTests.swift
@@ -309,7 +309,7 @@ class OnDemandMapModelTests: XCTestCase {
 }
 
 extension OnDemandMapModel.Status: Equatable {
-    static func == (lhs: OnDemandMapModel.Status, rhs: OnDemandMapModel.Status) -> Bool {
+    public static func == (lhs: OnDemandMapModel.Status, rhs: OnDemandMapModel.Status) -> Bool {
         return switch (lhs, rhs) {
         case (.initialized, .initialized),
             (.downloading, .downloading),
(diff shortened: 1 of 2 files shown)


Tests/ArcGISToolkitTests/OnDemandMapModelTests.swift
```diff
@@ -309,7 +309,7 @@ class OnDemandMapModelTests: XCTestCase {
 }
 
 extension OnDemandMapModel.Status: Equatable {
-    static func == (lhs: OnDemandMapModel.Status, rhs: OnDemandMapModel.Status) -> Bool {
+    public static func == (lhs: OnDemandMapModel.Status, rhs: OnDemandMapModel.Status) -> Bool {
```
yo1995, line 312: I have to admit I didn't rerun the tests when I applied the [suggestion](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/1535#discussion_r3778730362). 

@philium I believe `public` is needed here because the unit tests are in a difference module/target (`ArcGISToolkitTests` vs `ArcGISToolkit`), so the added `Equatable` conformance on `OnDemandMapModel.Status` is a retroactive conformance, hence needs to be public. Is my understanding correct?

yo1995 (approved):
Thanks for fixing these errors.

Tests/ArcGISToolkitTests/OnDemandMapModelTests.swift
philium: Oh! [This](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/attributes/#testable) is why:

> Entities in the imported module that are marked with the internal access-level modifier are imported as if they were declared with the public access-level modifier.

philium: Sorry about that.

yo1995: No worries! Thanks for the doc reference.

CalebRas (approved):
Builds for me on all platforms.

PR #1539, opened 2026-08-17
Fix double-trigger of examples in compact width

### LINKED ISSUE #1478: Example views are presented twice in compact width environments
In compact width environments, when selecting subsequent examples (not the first one), the example view is presented twice: it appears, it disappears, and reappears. This can be seen by adding the following to any example (or multiple examples):

```swift
@State private var id = UUID()
```

```swift
.onAppear {
    print("\(id) onAppear")
}
.onDisappear {
    print("\(id) onDisappear")
}
```

Run the app, open an example, tap the back button, and select one of the modified examples. You'll see something like this in the console:

```
724BE33D-EB4E-445C-BD49-8B74CF6695FE onAppear
724BE33D-EB4E-445C-BD49-8B74CF6695FE onDisappear
85767EC3-A465-4F9C-A1F3-753F726AFE49 onAppear
85767EC3-A465-4F9C-A1F3-753F726AFE49 onDisappear
85767EC3-A465-4F9C-A1F3-753F726AFE49 onAppear
```

The first onAppear/onDisappear pair is from the initial example. The second example appears, disappears, and then appears again. This is unexpected, inefficient, and in some cases harmful (see swift/8234).

I suspect that using navigation links would address this problem. I checked a different app that uses navigation links and it does not exhibit this same problem.

NoDancing:
NavigationSplitView, when collapsed to a single column, double-triggers onAppear/onDisappear after selecting multiple examples.

This fix replaces NavigationSplitView in compact environments with plain NavigationStack + NavigationLink. Functionality unchanged in other environments. 

Fixes #1478

### DIFF
--- a/Examples/ExamplesApp/Examples.swift
+++ b/Examples/ExamplesApp/Examples.swift
@@ -30,27 +30,23 @@ struct Examples: View {
                 case .category(let name, let examples):
                     DisclosureGroup(name) {
                         ForEach(examples, id: \.name) { example in
-                            Text(example.name)
-                                .tag(example)
+                            NavigationLink(example.name, value: example)
                         }
                     }
                 case .example(let example):
-                    Text(example.name)
-                        .tag(example)
+                    NavigationLink(example.name, value: example)
                 }
             }
             .navigationTitle("Toolkit Examples")
         } detail: {
-            // Workaround for https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/1249 with
-            // Apple Feedback #FB19470509 "Inconsistent task(priority:_:) behavior on views shown in NavigationSplitView detail column".
-            NavigationStack {
-                if let selectedExample {
-                    selectedExample.view
-                        .navigationTitle(selectedExample.name)
-                        .navigationBarTitleDisplayMode(.inline)
-                } else {
-                    Text("Select an example")
-                }
+            // Do not wrap this in a `NavigationStack`. With `NavigationLink` rows in the sidebar
+            // it is not needed for #1249 and causes a spurious `onDisappear` in compact width (#1478).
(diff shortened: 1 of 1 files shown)


mhdostal:
@NoDancing - thank you for your PR, we're always glad to see contributions from our users! Question: prior to settling on the approach in this PR, did you try the suggestion @philium had in the issue: _I suspect that using navigation links would address this problem. I checked a different app that uses navigation links and it does not exhibit this same problem._?

NoDancing:
@mhdostal Hello! Happy to contribute.

Yeah! First I tried to swap in `NavigationLink`s, and it didn't seem to work. I figured since the Panes don't matter in the compact width anyway, I'd swap it out with a `NavigationStack`, which worked immediately, but the code is definitely uglier.

With a couple weeks hindsight, and looking at some other examples online, maybe I had made a simple mistake and bounced off my initial solution too quickly. Let me see if I can make this a little cleaner!

NoDancing:
Alright! I gave this another pass. I got tripped up by the pre-existing bug fix, but the `NavigationLink` implementation solves that problem too, so it was safe to remove. Let me know if I'm off base, but I think this is a simpler way of doing it!

mhdostal (approved):
Looks great and works as advertised. Thanks!

philium (approved):
Thank you!

mhdostal:
@NoDancing - This looks good and has been reviewed, so please merge it when you get a chance! Thanks again!

mhdostal:
> so please merge it when you get a chance!

I'm actually not sure if you have permission to merge. If not, ping me and I can do it here.

NoDancing:
@mhdostal Yeah, I don't have merge permission, so you can go ahead!

PR #1541, opened 2026-08-20
`FeatureEditor`: Support geometry-dependant forms

CalebRas:
## Description

This PR adds Feature Editor support for geometry-dependent form elements by re-evaluating the form's expressions when geometry is changed. 
- Closes `swift/issues/8275`

## How To Test

- Replace the [`ExampleMapView.map`](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/FeatureEditor/FeatureEditorExample/FeatureEditorExample/ExampleMapView.swift#L27-L38) declaration with:
    ```swift
    @State private var map = Map(url: URL(string: "https://www.arcgis.com/home/item.html?id=cc9bd34a82814c82b273a580c116f5af")!)!
    ```
- Run `ExampleMapView` project and tap on a map feature.
- Edit the feature's geometry and observe the `Area (m)` form element update accordingly. 

## Screenshots

|Before|After|
|:-:|:-:|
| Before | After |

### DIFF
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
@@ -110,7 +110,13 @@ final class FeatureEditorModel {
     func restartGeometryEditor() async {
         guard geometryEditorIsStarted else { return }
         
-        await setFormGeometry(to: initialGeometry)
+        do {
+            try await setFormGeometry(to: initialGeometry)
+        } catch {
+            Logger.featureEditor.error(
+                "Error updating form geometry: \(error.localizedDescription)"
+            )
+        }
         startGeometryEditor()
     }
     
@@ -201,14 +207,14 @@ final class FeatureEditorModel {
     
     /// Updates the form's feature geometry using the geometry editor's current
     /// geometry to update possible geometry-dependent form elements.
-    func updateFormGeometry() async {
+    func updateFormGeometry() async throws {
         guard geometryEditorIsStarted else { return }
         
         // Uses initialGeometry if the geometry editor has no edits to prevent
         // an empty geometry from being used when the geometry editor was
         // started using a geometryType (when feature.geometry is nil).
         let geometry = geometryEditorCanUndo ? geometryEditorGeometry : initialGeometry
-        await setFormGeometry(to: geometry)
+        try await setFormGeometry(to: geometry)
     }
     
     /// Loads the feature and its table if needed to allow geometry editing.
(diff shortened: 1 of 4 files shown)


Tests/ArcGISToolkitTests/FeatureEditorModelTests.swift
```diff
@@ -136,6 +139,24 @@ struct FeatureEditorModelTests {
...
+        let table = try await geodatabaseFile.geodatabase.makeTable(description: .points)
+        
+        let initialGeometry = Point(latitude: 0, longitude: 0)
+        let feature = try #require(table.makeFeature(geometry: initialGeometry) as? ArcGISFeature)
+        #expect(feature.geometry == initialGeometry)
+        
+        await model.startEditingFeature(feature, on: nil)
+        
+        let newGeometry = Point(latitude: 1, longitude: 1)
+        await model.setFormGeometry(to: newGeometry)
```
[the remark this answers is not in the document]
CalebRas: This is non-trivial to implement, so it will not be added.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
```diff
@@ -212,6 +228,25 @@ final class FeatureEditorModel {
...
+        }
+        
+        do {
+            featureForm.feature.geometry = geometry
+            try await featureForm.evaluateExpressions()
+        } catch {
+            Logger.featureEditor.error(
+                "Error evaluating expressions: \(error.localizedDescription)"
+            )
+        }
```
philium: Consider making this method throwing (and the calling method throwing) and allow the view to determine what to do with the error. Evaluating an expression is really an implementation detail. It is the view that should know 1) whether the operation it initiated succeeded, and 2) what to do with the error if it failed. Dealing with the error here takes control away from the view.

philium (approved):
Code builds and all feature editor model tests are passing.

mhdostal (approved):

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
CalebRas: I decided to make `setFormGeometry(to:)` and `updateFormGeometry()` throwing, but not `restartGeometryEditor()`, because I don't think it makes sense to have an evaluate-expressions block geometry editor restarting, given that geometry-dependent form elements are kind of an edge case in my opinion. Let me know if you had something different in mind.

philium: Ah! I see. That makes sense. Thanks!

philium (approved):

CalebRas:
Thank you!

PR #1543, opened 2026-08-21
`FeatureEditor`: Add a button for adding features

philium:
## Description

This PR adds a + button to the feature editor toolbar that is visible when no feature is selected. Tapping it presents a yet-to-be implemented template picker for adding features and the + button goes away. Tapping on the close button dismisses the template picker and brings back the + button, or tapping on a feature replaces the template picker with the feature editor form view.

- Closes `swift/issues/8538`

## How To Test

- Open the example and tap the + button
- Run the new unit test
- Run the new UI test

## Screenshots

https://github.com/user-attachments/assets/77a99f10-97ef-41c4-83aa-2a6cb3ae5ccb

### DIFF
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
@@ -21,6 +21,21 @@ internal import os
 @MainActor
 @Observable
 final class FeatureEditorModel {
+    // MARK: State
+    
+    /// The various states of the feature editor.
+    enum State: Equatable {
+        /// The feature editor is adding features.
+        case adding
+        /// The feature editor is editing a feature.
+        case editing
+        /// The feature editor is stopped.
+        case stopped
+    }
+    
+    /// The current state of the feature editor.
+    private(set) var state: State = .stopped
+    
     // MARK: Properties
     
     /// The feature currently being edited by the feature editor.
@@ -34,7 +49,7 @@ final class FeatureEditorModel {
     /// A Boolean value that indicates whether the Feature Editor inspector is presented.
     /// This maps `rootFeatureForm` to a Boolean value.
     var isPresented: Bool {
-        get { rootFeatureForm != nil }
+        get { state != .stopped }
         set {
             guard !newValue else { return }
             stopEditing()
@@ -120,6 +135,20 @@ final class FeatureEditorModel {
         startGeometryEditor()
     }
     
+    // MARK: Adding
+    
+    /// Starts adding new features from templates.
+    func startAddingFeatures() {
+        state = .adding
+    }
+    
+    /// Stops adding new features.
+    func stopAddingFeatures() {
+        state = .stopped
+    }
+    
+    // MARK: Editing
+    
     /// Starts editing a new `FeatureForm` that is shown in the feature editor's `FeatureFormView`.
     /// - Parameter featureForm: The new feature form to edit.
     func startEditingFeatureForm(_ featureForm: FeatureForm) async {
@@ -137,7 +166,8 @@ final class FeatureEditorModel {
     ///   - feature: The root feature to edit.
     ///   - map: The map that `feature` is part of, used to set up rule-based snapping.
     func startEditingFeature(_ feature: ArcGISFeature, on map: Map?) async {
-        stopEditing()
+        state = .editing
+        resetProperties()
         rootFeatureForm = FeatureForm(feature: feature)
         self.map = map
         loadResult = await Result {
@@ -172,6 +202,15 @@ final class FeatureEditorModel {
     /// This is needed to prevent the current state from interfering with
     /// future uses of the `FeatureEditor` view.
     func stopEditing() {
+        state = .stopped
+        resetProperties()
+    }
+    
+    /// Resets the model's properties.
+    ///
+    /// This is needed to prevent the current state from interfering with
+    /// future uses of the `FeatureEditor` view.
+    private func resetProperties() {
         initialGeometry = nil
         rootFeatureForm = nil
         presentedFeatureForm = nil
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
@@ -47,7 +47,6 @@ final class FeatureEditorModel {
     @ObservationIgnored
     private var initialGeometry: Geometry?
     /// A Boolean value that indicates whether the Feature Editor inspector is presented.
-    /// This maps `rootFeatureForm` to a Boolean value.
     var isPresented: Bool {
         get { state != .stopped }
         set {
--- a/Sources/ArcGISToolkit/Components/FeatureEditor/Template Picker/FeatureEditorTemplatePicker.swift
+++ b/Sources/ArcGISToolkit/Components/FeatureEditor/Template Picker/FeatureEditorTemplatePicker.swift
@@ -1,0 +1,40 @@
+// Copyright 2026 Esri
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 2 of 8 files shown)


rolson (approved):

philium:
I forgot to document one of the new declarations. I made the change before creating my PR, but forgot to push it. This PR is still ready for review.

rolson (approved):

Sources/ArcGISToolkit/Components/FeatureEditor/Template Picker/FeatureEditorTemplatePicker.swift
```diff
@@ -0,0 +1,40 @@
...
+struct FeatureEditorTemplatePicker: View {
+    @Environment(FeatureEditorModel.self) private var model
+    
+    var body: some View {
+        ContentUnavailableView("Under Construction", systemImage: "exclamationmark.triangle")
+            .navigationTitle(
+                LocalizedStringResource(
+                    "Templates",
+                    bundle: .toolkit,
+                    comment: "The title of the template picker view"
```
CalebRas:
```suggestion
                    comment: "The title of the template picker view."
```

It looks like the Toolkit usually puts a period at the end of the localization comments. Does this need one?

Sources/ArcGISToolkit/Components/FeatureEditor/Toolbar/FeatureEditorToolbar.swift
```diff
@@ -51,6 +52,20 @@ struct FeatureEditorToolbar: View {
...
+                Button(
+                    LocalizedStringResource(
+                        "Add Features",
+                        bundle: .toolkitModule,
+                        comment: "A label for a button to show a feature template picker."
+                    ),
+                    systemImage: "plus",
+                    action: model.startAddingFeatures
+                )
+                .toolbarStackStyle()
```
CalebRas: Since `toolbarStackStyle()` is now being used outside of the context of a stack, I wonder if it should be renamed to something like `toolbarStyle()`. Up to you.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
```diff
@@ -30,7 +45,7 @@ final class FeatureEditorModel {
     /// A Boolean value that indicates whether the Feature Editor inspector is presented.
     /// This maps `rootFeatureForm` to a Boolean value.
```
CalebRas: Now that `FeatureEditorModel.isPresented` is based on `state` instead of `rootFeatureForm`, does this comment need to be updated?

Test Runner/Test Runner/TestViews/FeatureEditorTestView.swift
```diff
@@ -80,16 +80,13 @@ private extension FeatureEditorTestView {
             try await ArcGISEnvironment.authenticationManager.arcGISCredentialStore.add(.publicSample)
             try await map.retryLoad()
             
-            guard let objectID = UserDefaults.standard.objectID,
+            if let objectID = UserDefaults.standard.objectID,
                   let layerName = UserDefaults.standard.layerName,
                   let groupLayer = map.operationalLayers.first as? GroupLayer,
                   let layer = groupLayer.layers.first(where: { $0.name == layerName }),
```
CalebRas: Now that this is an if-statement, it looks like the following let-statements need to be reindented to match.

Tests/ArcGISToolkitTests/FeatureEditorModelTests.swift
```diff
@@ -259,6 +259,14 @@ struct FeatureEditorModelTests {
         model.stopEditing()
         await model.expectDefaultPropertyValues()
     }
+    
+    @Test func startAndStopAddingFeatures() {
```
CalebRas, line 303:
```suggestion
    @Test 
    func startAndStopAddingFeatures() {
```

Should `@Test` be on a separate line to match the rest of the tests?

Sources/ArcGISToolkit/Components/FeatureEditor/Toolbar/FeatureEditorToolbar.swift
```diff
@@ -51,6 +52,20 @@ struct FeatureEditorToolbar: View {
...
+            default:
+                // No controls.
+                EmptyView()
```
CalebRas, line 73: Now that we have the add button, I wonder if it would make sense to always display a view here, i.e., show the toolbar when editing and the add button at all other times, which can be disabled when the `state` is not `stopped`. Something like this:

```swift
switch model.state {
case .editing where model.geometryEditorIsStarted:
    ...
default:
    Button(
        ...
    )
    .toolbarStackStyle()
    .disabled(model.state != .stopped)
}
```

This would prevent the `FeatureEditor` view from disappearing while loading and then reappearing when completed, which seems a little odd to me. I could go either way on this, though, so any thoughts?

Sources/ArcGISToolkit/Components/FeatureEditor/Toolbar/FeatureEditorToolbar.swift
CalebRas: Could you update the `FeatureEditor` "Behavior" doc to mention the add button, please?
https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/FeatureEditor/Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditor.swift#L44-L46

Tests/ArcGISToolkitTests/FeatureEditorModelTests.swift
CalebRas, line 303: Could you also update `expectDefaultPropertyValues(sourceLocation:)` to expect `state == .stopped` and `expectIsEditing(rootFeature:sourceLocation:)` and to expect `state == .editing`, please, so that `state` gets tested in the rest of the cases as well?

CalebRas (commented):
Looks good overall! Just a few comments:

Sources/ArcGISToolkit/Components/FeatureEditor/Template Picker/FeatureEditorTemplatePicker.swift
philium: Didn't know that. Thanks!

Sources/ArcGISToolkit/Components/FeatureEditor/Toolbar/FeatureEditorToolbar.swift
philium: Sure, let's go with that.

Sources/ArcGISToolkit/Components/FeatureEditor/FeatureEditorModel.swift
philium: Yes, that makes sense.

Test Runner/Test Runner/TestViews/FeatureEditorTestView.swift
philium: Thank you.

Tests/ArcGISToolkitTests/FeatureEditorModelTests.swift
philium: I don't see a reason to unless it is a hard requirement of the toolkit. We put other smaller attributes (e.g. `State`) on the same line. Also, having `@Test` on the same like is similar to having `test` in the name:

```swift
// XCTest
func testStartAndStopAddingFeatures() {
// Swift Testing
@Test func startAndStopAddingFeatures() {
```

Sources/ArcGISToolkit/Components/FeatureEditor/Toolbar/FeatureEditorToolbar.swift
philium: I'm not sure that makes sense because depending on the API we come up with, there may or may not be templates for the user to pick, right? In that case, we would only want to display the add button if adding is supported (i.e. there are templates).

Sources/ArcGISToolkit/Components/FeatureEditor/Toolbar/FeatureEditorToolbar.swift
philium: How does this sound?

```swift
/// When no feature is selected, the toolbar includes a button to add features.
/// When a feature is selected, the toolbar includes editing controls.
```

Sources/ArcGISToolkit/Components/FeatureEditor/Toolbar/FeatureEditorToolbar.swift
CalebRas: That makes sense. Good to leave as-is, then. Thanks for thinking about it.

CalebRas (approved):
Looks good!

CalebRas (approved):

rolson (approved):

PR #1544, opened 2026-08-22
Refine LocationButton interactions

rolson:
## Summary

- Fix `LocationButton` so a sole configured auto-pan mode is restored after map interaction turns auto-pan off.
- Add `allowsHidingLocationDisplay(_:)`, enabled by default, to control whether **Hide Location** appears in the context menu.
- Keep hiding context-menu-only and show auto-pan menu options only when multiple modes are configured.
- Limit context-menu choices to explicitly configured modes; `.off` is moved first only when supplied.
- Guard asynchronous start and stop actions against duplicate requests.
- Update component documentation and regression coverage for mode cycling and menu configurations.

## Testing

- Added unit coverage for current-mode action availability, hide configuration, and context-menu options.

### DIFF
--- a/Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
+++ b/Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
@@ -331,7 +331,7 @@ extension LocationButton {
     /// The context menu auto-pan mode options.
     ///
     /// The context menu options will be in the order the user specifies
-    /// except the off option, when included, will be first.
+    /// except the off option, which, when included, will be first.
     var contextMenuAutoPanOptions: [LocationDisplay.AutoPanMode] {
         autoPanModes.filter { $0 == .off } +
         autoPanModes.filter { $0 != .off }
--- a/Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
+++ b/Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
@@ -91,7 +91,7 @@ public struct LocationButton: View {
     
     /// Sets whether the user can hide the location display.
     /// - Parameter isAllowed: A Boolean value indicating whether hiding the
-    ///   location display is included in the context menu.
+    /// location display is included in the context menu.
     /// - Returns: A new location button configured with the given value.
     public func allowsHidingLocationDisplay(_ isAllowed: Bool) -> Self {
         var copy = self
--- a/Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
+++ b/Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
@@ -44,7 +44,7 @@ public struct LocationButton: View {
     
     /// A Boolean value indicating whether the user can hide the location display.
     private(set) var allowsHidingLocationDisplay = true
-
+    
     /// The action performed when the button is pressed.
     var currentAction: Action? {
         Action(
@@ -53,7 +53,7 @@ public struct LocationButton: View {
             autoPanOptions: autoPanModes
         )
     }
-
+    
     /// A value indicating whether the button is disabled.
     var buttonIsDisabled: Bool {
         status == .starting ||
@@ -98,7 +98,7 @@ public struct LocationButton: View {
         copy.allowsHidingLocationDisplay = isAllowed
         return copy
     }
-
+    
     public var body: some View {
         Button {
             buttonAction()
@@ -322,7 +322,7 @@ extension LocationButton {
     var shouldShowContextMenu: Bool {
         allowsHidingLocationDisplay || autoPanModes.count > 1
     }
-
+    
     /// The initial auto-pan mode to be used.
     var initialAutoPanMode: LocationDisplay.AutoPanMode {
         autoPanModes.first ?? .off
--- a/Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
+++ b/Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
@@ -22,14 +22,13 @@ internal import os
 /// Gives the user a variety of options to set the auto-pan mode or stop the
 /// location data source.
 ///
-/// The button will cycle through the specified auto-pan modes on tap. The user
-/// can also hide the location display or select the auto-pan mode through a
-/// context menu.
+/// After starting the location display, the button cycles through the specified
+/// auto-pan modes on tap.
 ///
-/// If there are no auto-pan modes specified, or if the only specified mode is
-/// `off`, then:
-/// - the location display is toggled on/off when pressed
-/// - the context menu is not shown when long pressed
+/// By default, a context menu allows the user to hide the location display.
+/// The menu can also be shown when hiding is disabled if at least two auto-pan
+/// modes are specified. Auto-pan mode options appear in the menu only when at
+/// least two modes are specified.
 public struct LocationButton: View {
     /// The location display controlled by the location button.
     let locationDisplay: LocationDisplay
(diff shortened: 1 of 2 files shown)


Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
```diff
@@ -43,9 +42,24 @@ public struct LocationButton: View {
     /// A Boolean value indicating that the button action is being performed.
     @State private(set) var isPerformingButtonAction = false
     
+    /// A Boolean value indicating whether the user can hide the location display.
+    private(set) var allowsHidingLocationDisplay = true
```
yo1995, line 46: Should this be just `private` instead as it is only set by the modifier?
Other than in the test, developers don't seem to have a way to get it.
```swift
    private var allowsHidingLocationDisplay = true
```

Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
```diff
@@ -75,6 +89,16 @@ public struct LocationButton: View {
         return copy
     }
     
+    /// Sets whether the user can hide the location display.
+    /// - Parameter isAllowed: A Boolean value indicating whether hiding the
+    ///   location display is included in the context menu.
```
yo1995:
```suggestion
    /// location display is included in the context menu.
```
nitpick: Xcode auto-indent like this

Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
```diff
@@ -290,6 +318,11 @@ public struct LocationButton: View {
 }
 
 extension LocationButton {
+    /// A Boolean value indicating whether the context menu should be shown.
+    var shouldShowContextMenu: Bool {
```
yo1995, line 322: Can be private, but since it in the unit test to verify the context menu behavior, good to keep internal.

Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
```diff
@@ -43,9 +42,24 @@ public struct LocationButton: View {
     /// A Boolean value indicating that the button action is being performed.
     @State private(set) var isPerformingButtonAction = false
     
+    /// A Boolean value indicating whether the user can hide the location display.
+    private(set) var allowsHidingLocationDisplay = true
+
```
dfeinzimer:
```suggestion
    
```

Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
```diff
@@ -43,9 +42,24 @@ public struct LocationButton: View {
...
+    /// The action performed when the button is pressed.
+    var currentAction: Action? {
+        Action(
+            status: status,
+            autoPanMode: autoPanMode,
+            autoPanOptions: autoPanModes
+        )
+    }
+
```
dfeinzimer:
```suggestion
    
```

Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
rolson: getter needs to be internal for test code

Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
```diff
@@ -75,6 +89,16 @@ public struct LocationButton: View {
...
+    /// Sets whether the user can hide the location display.
+    /// - Parameter isAllowed: A Boolean value indicating whether hiding the
+    ///   location display is included in the context menu.
+    /// - Returns: A new location button configured with the given value.
+    public func allowsHidingLocationDisplay(_ isAllowed: Bool) -> Self {
+        var copy = self
+        copy.allowsHidingLocationDisplay = isAllowed
+        return copy
+    }
+
```
dfeinzimer:
```suggestion
    
```

Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
```diff
@@ -290,6 +318,11 @@ public struct LocationButton: View {
 }
 
 extension LocationButton {
+    /// A Boolean value indicating whether the context menu should be shown.
+    var shouldShowContextMenu: Bool {
+        allowsHidingLocationDisplay || autoPanModes.count > 1
+    }
+
```
dfeinzimer:
```suggestion
    
```

Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
```diff
@@ -298,9 +331,10 @@ extension LocationButton {
     /// The context menu auto-pan mode options.
     ///
     /// The context menu options will be in the order the user specifies
-    /// except the off option will be first.
+    /// except the off option, when included, will be first.
```
dfeinzimer: The context menu options will be in the order the user specifies, except for the `.off` option, which, when included, will be first.

Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
```diff
@@ -75,6 +89,16 @@ public struct LocationButton: View {
         return copy
     }
     
+    /// Sets whether the user can hide the location display.
+    /// - Parameter isAllowed: A Boolean value indicating whether hiding the
+    ///   location display is included in the context menu.
+    /// - Returns: A new location button configured with the given value.
+    public func allowsHidingLocationDisplay(_ isAllowed: Bool) -> Self {
```
dfeinzimer, line 96: Name seems verbose, repeated type info.

```suggestion
    public func allowsHiding(_ isAllowed: Bool) -> Self {
```


```suggestion
    public func hiding(isAllowed: Bool) -> Self {
```

Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
```diff
@@ -226,44 +250,48 @@ public struct LocationButton: View {
...
                 }
-            }
-            .accessibilityHint(
-                Text(
-                    "Hide location display",
-                    bundle: .toolkitModule,
-                    comment: "The accessibility hint of the hide location display context menu option."
+                .accessibilityHint(
+                    Text(
+                        "Hide location display",
```
dfeinzimer, line 289: Please consider updating the strings file in this same PR with [update-translations](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/v.next/Scripts/update-translations).

Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
rolson: I think the name is appropriate because it's not the button that will hide, but the location display. Thoughts?

Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
rolson: I don't think the string changed though, I think the indentation changed only.

Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
rolson:
```suggestion
    /// except the off option, which, when included, will be first.
```

Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
yo1995: I'm fine with the current name

yo1995 (approved):

Sources/ArcGISToolkit/Components/LocationButton/LocationButton.swift
dfeinzimer: Apologies, I missed that removal in the diff.

dfeinzimer (approved):

PR #1545, opened 2026-08-25
`FeatureEditor`: Fix example console error

CalebRas:
## Description

This PR fixes a `FeatureEditorExample` bug where an authentication error would print in the console when opening the `ExampleMapView`, even though authentication was successful.
- Closes `swift/issues/8545`

## How To Test

- Run the `FeatureEditorExample` project, tap "Open Map View", and ensure this error is not printed in the console:

    ```
    An ArcGIS authentication error occurred when accessing 'sampleserver7.arcgisonline.com'. An API key or credential may be required. For more information see 'https://developers.arcgis.com/swift/security-and-authentication'.
    ```
- Ensure the example still otherwise works as expected.

### DIFF
--- a/FeatureEditorExample/FeatureEditorExample/ExampleMapView.swift
+++ b/FeatureEditorExample/FeatureEditorExample/ExampleMapView.swift
@@ -23,7 +23,7 @@ struct ExampleMapView: View {
     /// The geometry editor that the feature editor will use to edit geometries
     /// on the `MapView`.
     @State private var geometryEditor = GeometryEditor()
-    /// The map with the features to edit, created from a Naperville Electric web map portal item.
+    /// The map with the features to edit.
     @State private var map = Map()
     /// The point on the screen where the user tapped.
     @State private var tapPoint: CGPoint?
@@ -79,15 +79,16 @@ struct ExampleMapView: View {
         }
         .task {
             do {
-                try await setUp()
+                map = try await makeMap()
             } catch {
-                print("Setup error:", error)
+                print("Failed to make map:", error)
             }
         }
     }
     
-    /// Sets up the map.
-    private func setUp() async throws {
+    /// Returns a map with the features to edit, created from a Naperville
+    /// Electric web map portal item.
+    private func makeMap() async throws -> Map {
         let url = URL(string: "https://sampleserver7.arcgisonline.com/portal/home/item.html?id=b4565e0a4e4c4a4382914128f10864cd")!
         let map = Map(url: url)!
         
@@ -107,6 +108,6 @@ struct ExampleMapView: View {
         ArcGISEnvironment.authenticationManager.arcGISCredentialStore.add(credential)
         try await map.load()
         
-        self.map = map
+        return map
     }
 }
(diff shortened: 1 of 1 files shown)


FeatureEditorExample/FeatureEditorExample/ExampleMapView.swift
```diff
@@ -107,6 +105,8 @@ struct ExampleMapView: View {
             password: "I68VGU^nMurF"
         )
         ArcGISEnvironment.authenticationManager.arcGISCredentialStore.add(credential)
-        try await map.retryLoad()
+        try await map.load()
+        
+        self.map = map
```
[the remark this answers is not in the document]
CalebRas: `setUp()` runs on the Main Actor, so this is not an issue.

FeatureEditorExample/FeatureEditorExample/ExampleMapView.swift
```diff
@@ -24,18 +24,7 @@ struct ExampleMapView: View {
...
-        
-        // Enables full resolution feature tiling to improve snapping accuracy.
-        map.loadSettings.featureTilingMode = .enabledWithFullResolutionWhenSupported
-        
-        let initialExtent = Envelope(xRange: -9812750 ... -9812387, yRange: 5131332 ... 5131965)
-        map.initialViewpoint = Viewpoint(boundingGeometry: initialExtent)
-        
-        return map
-    }()
+    @State private var map = Map()
```
philium: Instead of using a placeholder map, what would you think of making this optional, and then in the body, when there isn't a map, displaying a placeholder view (e.g. progress view)?

FeatureEditorExample/FeatureEditorExample/ExampleMapView.swift
```diff
@@ -99,6 +88,15 @@ struct ExampleMapView: View {
     
     /// Sets up the map.
     private func setUp() async throws {
```
philium: Consider naming this `makeMap()` and have it return the created map. Then the caller could do the assignment:

```swift
.task {
    do {
        map = try await makeMap()
    } catch {
        print("Failed to make map:", error)
    }
}
```

FeatureEditorExample/FeatureEditorExample/ExampleMapView.swift
CalebRas: I added a `ProgressView` but used a `Result` instead of an optional `map` since if the latter option was to fail, the `ProgressView` would continue to display, making it seem like it was still loading.

FeatureEditorExample/FeatureEditorExample/ExampleMapView.swift
```diff
@@ -107,6 +116,8 @@ struct ExampleMapView: View {
             password: "I68VGU^nMurF"
         )
         ArcGISEnvironment.authenticationManager.arcGISCredentialStore.add(credential)
-        try await map.retryLoad()
+        try await map.load()
```
philium, line 119: Does this need to be `retryLoad()` in order for retrying to work?

philium (approved):

FeatureEditorExample/FeatureEditorExample/ExampleMapView.swift
CalebRas: I actually don't think so because we are creating a new map each time `makeMap()` runs, so the load status will always be `notLoaded`.

philium: Oh! Of course. Never mind.

chriswebb09 (approved):

chriswebb09:
LGTM

PR #1546, opened 2026-08-27
`FeatureFormView` - Fix test case 14.1

dfeinzimer:
Test run 5092

Fixes FFV test case 14.1 which was working reliably up until Aug 18 when we started seeing another keyboard key being pressed at a certain step, causing the remainder of the test to fail. This repairs the test by explicitly closing the keyboard after the full element title has been keyed in.

I have no viable explanation as to what triggered this sudden change in behavior.

See discussion in swift-testing from Aug 24 for more context.


pgruenler (approved):

dhochhalterruiz (approved):

dfeinzimer:
Thanks for the quick reviews!

PR #1547, opened 2026-08-31
Improve Google VPS localization guidance

rolson:
## Summary

Improve the Google-backed custom VPS example so it does not treat ARCore's basic tracking state as sufficient geospatial localization.

- retain `CLLocationManager` for the complete asynchronous authorization flow
- require Precise Location before configuring ARCore Geospatial
- check VPS availability at the device's current coordinate
- require horizontal and yaw accuracy thresholds before revealing ArcGIS scene content
- use wider thresholds after localization to prevent rapid localized/unlocalized flicker
- surface permission, support, configuration, availability, and localization states in the example UI
- preserve the existing streetscape geometry configuration and toggle

### Location authorization is asynchronous

The existing example creates a local `CLLocationManager`, requests permission, and immediately checks its authorization status. The permission result arrives asynchronously, and the temporary manager is not retained to receive delegate callbacks. On a first run, `start()` therefore returns before ARCore is configured and has no path that resumes setup after the user grants access.

Retaining the manager and handling `locationManagerDidChangeAuthorization(_:)` allows setup to continue after the authorization prompt and also responds to later permission changes.

### Google VPS requires precise location

ARCore Geospatial requires location permission with full accuracy. Checking only `.authorizedWhenInUse` can allow setup to proceed with reduced accuracy, where localization cannot meet the accuracy expected by the example. The updated flow accepts both authorized states and explicitly reports when Precise Location is disabled.

### ARCore tracking is not the same as accurate geospatial localization

`GARTrackingState.tracking` indicates that Earth tracking is active, but it does not guarantee that the camera's global position and orientation are accurate enough to align ArcGIS content. Revealing the scene immediately can place content at a visibly incorrect position or heading.

Following Google's Geospatial sample, initial localization now requires:

- horizontal accuracy below 10 meters
- orientation yaw accuracy below 15 degrees

Once localized, the example retains that state until accuracy exceeds 20 meters or 25 degrees. This hysteresis avoids flickering when measurements hover around the initial threshold.

### VPS availability and failures need visible feedback

VPS coverage varies by location, and ARCore configuration can also fail because of permissions, unsupported devices, invalid authorization, or service errors. Previously these conditions either returned silently or left the camera running without explaining why scene content never appeared.

The example now checks `checkVPSAvailability(coordinate:)` and presents concise guidance or blocking errors over the camera feed. This makes the sample diagnosable and demonstrates the status handling an integrating application needs.

## Scope

This keeps the provider implementation directly in `CustomVPSProviderExample` so the integration remains easy to study. Streetscape geometry remains enabled and its existing **Show Buildings** toggle is unchanged.

## Validation

- Built the `CustomVPSProviderExample` scheme for a generic iOS Simulator destination.
- Ran the example target's existing SwiftLint build phase without violations.

## Commit breakdown

1. [`19981b65d`](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/commit/19981b65d) **Harden Google VPS provider setup** — Retains `CLLocationManager` for asynchronous authorization, requires Precise Location, checks VPS availability, and reports unsupported or failed provider setup.
2. [`d5194a7c1`](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/commit/d5194a7c1) **Require accurate Google VPS localization** — Requires horizontal and yaw accuracy before revealing scene content and adds wider retention thresholds to prevent localization flicker.
3. [`087ad9cb7`](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/commit/087ad9cb7) **Display Google VPS localization status** — Presents actionable tracking guidance and blocking failures over the camera feed.

## Source for accuracy thresholds

The 10 m / 15° initial-localization thresholds and 20 m / 25° loss-of-localization thresholds come directly from [Google’s ARCore Geospatial example](https://github.com/google-ar/arcore-ios-sdk/blob/5aa221f8e9a516643fcc7b080c565f22b20ab57c/Examples/GeospatialExample/GeospatialExample/GeospatialManager.swift#L43-L53). The link is pinned to the ARCore 1.54.0 revision used by this example so the cited values cannot drift with `main`.

### DIFF
--- a/CustomVPSProviderExample/CustomVPSProviderExample/CustomVPSProviderExampleView.swift
+++ b/CustomVPSProviderExample/CustomVPSProviderExample/CustomVPSProviderExampleView.swift
@@ -57,6 +57,17 @@ struct CustomVPSProviderExampleView: View {
                         SceneView(scene: scene)
                     }
                     .calibrationViewHidden(true)
+                    .safeAreaInset(edge: .top) {
+                        Label(
+                            provider.statusMessage,
+                            systemImage: provider.hasBlockingError
+                                ? "exclamationmark.triangle"
+                                : "location.viewfinder"
+                        )
+                        .padding(.horizontal)
+                        .padding(.vertical, 8)
+                        .background(.regularMaterial, in: .capsule)
+                    }
                     .toolbar {
                         ToolbarItem(placement: .bottomBar) {
                             Toggle("Show Buildings", isOn: $streetscapeGeometryEnabled)
--- a/CustomVPSProviderExample/CustomVPSProviderExample/CustomWorldTracking.swift
+++ b/CustomVPSProviderExample/CustomVPSProviderExample/CustomWorldTracking.swift
@@ -17,32 +17,49 @@ import ArcGISToolkit
 import ARCore
 import ARCoreGeospatial
 import ARKit
+import CoreLocation
+import Observation
 import os
 import RealityKit
 
 /// A world-tracking provider backed by the Google ARCore SDK.
-final class CustomWorldTracking {
-    @MainActor let arSessionProvider = ARSessionProvider<ARView>()
+@MainActor
+@Observable
+final class CustomWorldTracking: NSObject {
+    let arSessionProvider = ARSessionProvider<ARView>()
     /// The ARCore session that produces geospatial and streetscape updates.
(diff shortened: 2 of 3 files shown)


rolson:
Closing as @des12437 advises this code adds unnecessary complexity

